PR #1, opened 2016-03-03
Refreshed UI based on recent Dash corpo style

tludek:
Hi @HashEngineering,
here you may find all changes introduced to the code to apply new UI, consistent with the recent Dash corpo branding.
I know it is a lot of changes, but to take advantage of fancy features of Google Material Design had to switch all default common classes to its equivalents from android support library.
Do not be afraid, I know what I'm doing ;)
Do not hesitate to let me know in case of any problems with merging.


HashEngineering:
Let me ask a few questions:
1.  Changes were made in build.gradle that were added to handle the native libraries.  Will the native libraries still end up in the apk in the right place?

Is this the better way of doing it?
wallet/src/main/native-libs/mips64/libx11.so → wallet/src/main/jniLibs/mips64/libx11.so

2.   What is the SVG support feature in Android Studio:  from "Replace SvgDrawableTask with standard SVG support feature from Android Studio"

tludek:
Hi Hash,

`1. Changes were made in build.gradle that were added to handle the native libraries. Will the native libraries still end up in the apk in the right place?`
From a certain time (I can't find the exact version number) the android gradle plugin automatically copies all files from **jniLibs** folder to appropriate place in build folder. Therefore I think it is the better way, because there is no need to apply any custom scripts. The only thing I had to do, was moving all libraries to **jniLibs** folder.

`2. What is the SVG support feature in Android Studio: from "Replace SvgDrawableTask with standard SVG support feature from Android Studio"`
As in the previous case - from a certain time the android gradle plugin supports SVG format by defoult and there is no need to use 3rd-party libraries. You may find more details here: http://developer.android.com/tools/help/vector-asset-studio.html

My motivation, to do all those steps was to simplify the build script and make it more standard.
Please let me know, whether my answers are clear enough or should provide more details.

Regards
Tomasz

HashEngineering:
Thank you.  This helps.

HashEngineering:
Since there are merge conflicts, I am doing the using the command line.

There are two compile errors:
This is missing:
R.drawable.balance_prefix_micro - this refers to a file that is not in the drawable folder.  For now I will comment the related code out.

This is also missing:
R.id.wallet_options_disconnect  - This is something I added to the app recently and it wasn't in your fork.  For now I put this back on the main drawer, under Home.

HashEngineering:
Pull Request merge completed.

PR #2, opened 2016-03-06
Added missing resource file

tludek:
Hi @HashEngineering.
here is the missing resource file. Forgot to add it to git previously, sorry.


tludek:
Thanks!

Issue #35, opened 2017-06-15
Add Confirm Dialog on Send


HashEngineering:
It would be nice if there was an option to turn on a send confirmation dialog, so hitting send makes a box pop up that says something like "Do you really want to send x_address y Dash ($z)? Yes or No"

sambarboza:
Is contribution welcome/allowed? If so I would like to work on this issue. Thanks!

HashEngineering:
Yes, we welcome submissions.

HashEngineering:
This has been addressed by the redesigned interface.

PR #54, opened 2017-12-21
Confirm send

Nutomic:
Requires #53.

Should I add a preference for this? Default on or off?


HashEngineering:
I will be reviewing this soon.  It is an answer to #35.

sambarboza:
We'll consider this for a future update.

PR #59, opened 2018-01-12
Don't show additional toolbar in Address screen

sambarboza:
# Description
- An additional Toolbar is shown when an Address is Selected.

# Changes
- Added `windowActionModeOverlay` and `actionModeBackground` in the app's Theme Styles

# Evidence


HashEngineering:
Excellent work.  Tested and Verified.

PR #63, opened 2018-01-16
Hiding Scan QR Button on Transactions Screen when Scrolling

sambarboza:
# Description
- The Scan QR button hides part of the transaction data on the main screen

# Changes
- Created `ScrollFABBehavior` class and set as layout behavior in the Scan QR Button on the main screen layout, this layout behavior will hide the button when scrolling down the list and show it again when scrolling up.
- Created `WalletTransactionsFragment.OnFilterSelected` interface, implementing it in `Wallet Activity` and calling it form `WalletTransactionsFragment` when a filter is selected to restore the visual state of the Scan QR Button.

# Evidence

### DIFF
--- a/wallet/res/anim/fab_hide.xml
+++ b/wallet/res/anim/fab_hide.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android"
+    android:interpolator="@android:interpolator/decelerate_quad"
+    android:fillAfter="true">
+    <translate
+        android:duration="@android:integer/config_longAnimTime"
+        android:fromYDelta="0%"
+        android:toYDelta="100%p" />
+</set>
--- a/wallet/res/anim/fab_show.xml
+++ b/wallet/res/anim/fab_show.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android"
+    android:interpolator="@android:interpolator/decelerate_quad"
+    android:fillAfter="true">
+    <translate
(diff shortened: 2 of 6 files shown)


HashEngineering:
This works as advertised.  However, what if a user scrolls down and then says, "I can't scan a QR code".  Yes, the user can tap Send Coins (or Send Dash in other translations), but the button is gone.

In Bitcoin Wallet, the scan button was a camera icon next to the Send Coins button.

sambarboza:
I think that usually users will learn that the button appears as soon as they scroll up, but I might be wrong (we cannot be sure about it without a usability testing), however, a different thing that we can do is to add a padding bottom to the `RecyclerView` that is equal the `Scan QR Button` + its margins. This way the button will be always be visible.

HashEngineering:
What I don't like about your solution with adding padding is that that area is not used for anything besides the button.  What is your opinion on that?  My expertise is not with design, if your solution follows acceptable android UX guidelines, then I would go with it.  To date no one has complained about the current design with respect to that button.

The floating button was added as part of our previous redesign.  Before that the scan button was included on the tool bar with Send, Request and other buttons if they fit on the screen (either the top or bottom depending on the device).  Before that we didn't have issue with the button placement.

The Send activity had the scan button at the top on the toolbar.  With our redesign there is no room to put it back there, given that the wallet balance is taking up that room.  The good thing about the floating button was that it was floating on both screens (transaction list and send).

sambarboza:
Ok, I agree that the padding area is not a perfect solution, it's more like a workaround and we should avoid it. As we don't have users complaining about the position of the scan button, we can close this PR unmerged and wait for a better design solution.

HashEngineering:
Another solution would be to add the button that Bitcoin Wallet has in its interface.

PR #64, opened 2018-01-17
Hiding Scan button on Send Coins Screen when Soft Keyboard goes up

sambarboza:
# Description
- The `Scan QR Code Button` is hiding the `Send Button` on the `Send Coins Activity`.

# Changes
- Created a subclass of `CoordinatorLayout` named `KeyboardResponsiveCoordinatorLayout` that hides a view when the soft keyboard is opened.
- Created an optional custom attribute called `viewToHideWhenSoftKeyboardIsOpen` that can be passed to a `KeyboardResponsiveCoordinatorLayout`. The view which id is set on this attribute will be hidden if the soft keyboard shows up.

# Evidence

### DIFF
--- a/wallet/res/layout/send_coins_content.xml
+++ b/wallet/res/layout/send_coins_content.xml
@@ -1,4 +1,5 @@
-<de.schildbach.wallet.ui.widget.KeyboardResponsiveCoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<de.schildbach.wallet.ui.widget.KeyboardResponsiveCoordinatorLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     app:viewToHideWhenSoftKeyboardIsOpen="@+id/fab_scan_qr"
--- a/wallet/res/layout/send_coins_content.xml
+++ b/wallet/res/layout/send_coins_content.xml
@@ -1,6 +1,7 @@
-<android.support.design.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<de.schildbach.wallet.ui.widget.KeyboardResponsiveCoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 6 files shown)


HashEngineering:
What happens to the Scan button if a user:

- goes to this screen and then presses the scan button, then types an amount?  In this case the user scanned the address and doesn't necessarily need to scan it again, but what if he scans the wrong QR code and needs to scan another one?
- goes to this screen then types an amount and then needs to scan a QR code?  Will the scan button be there to scan a code?
- from the main wallet activity scans a QR code (which opens the send screen)?

There may be other combinations where the scan button disappears when it is needed.

Previously, on Bitcoin Wallet the scan button was at the top right of the toolbar.

HashEngineering:
This works as advertised.  However, what if a user scrolls down and then says, "I can't scan a QR code".  Yes, the user can tap Send Coins (or Send Dash in other translations), but the button is gone.

In Bitcoin Wallet, the scan button was a camera icon next to the Send Coins button.

HashEngineering:
When I go to the Send Coins page, the focus is set to the address text edit field and the scan button disappears.  Now I can't scan a QR code, unless I go back to the previous screen.

sambarboza:
The button will always be visible whenever the soft keyboard is hidden, as the amount input is focused automatically it doesn't show initially, we can prevent this from happening so the button will be visible when the user enters in the screen for the first time.

HashEngineering:
I like this idea of the scan button being visible when the user first goes to the Send Dash screen.  They can can an address, then enter a number.  after which the scan button will disappear.  They can press done and the scan button shows up again (as if it was covered by the soft keyboard).

Also, there is a "Scan" menu item on the options menu, which does not do anything.  The Scan menu item shows up as an icon on Bitcoin Wallet tool bar.  Since it is in our menu, we should make sure it works or remove it.  In SendCoinsFragment, there are about 8 lines commented out regarding this.  Some are for not showing the icon and the rest are to not respond to the menu.

sambarboza:
Cool, so I can work on getting it to be visible when user enter the screen and the "Scan" options menu item to work. Thanks for your feedback, I will update this PR with those changes.

sambarboza:
@HashEngineering, just updated the PR with the previous discussed solution.

HashEngineering:
Tested and approved.

HashEngineering:
This was manually added by squashing and rebasing as https://github.com/HashEngineering/dash-wallet/commit/319ee769e16e5ee8aea7dc4f2ace0eecfa570c67.

PR #66, opened 2018-01-23
Inter App Communication

tomasz-ludek:
Added the Inter App Communication based on iOS specification: https://dashpay.atlassian.net/wiki/spaces/DOC/pages/136118304/iOS+Wallet+URL+Scheme
(https://dashpay.atlassian.net/browse/DMD-46)

In order to simplify testing and show sample usage, the relevant section has been added to sample-integration project:

device-2018-01-23-191649

### DIFF
--- a/sample-integration-android/res/layout/sample_activity.xml
+++ b/sample-integration-android/res/layout/sample_activity.xml
@@ -8,12 +8,14 @@
     <LinearLayout
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
-        android:layout_marginBottom="24dp"
+        android:paddingBottom="24dp"
         android:layout_marginTop="24dp"
+        android:paddingLeft="16dp"
+        android:paddingRight="16dp"
         android:orientation="vertical" >
 
         <RadioGroup
-            android:layout_width="256dp"
+            android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="center"
             android:orientation="horizontal" >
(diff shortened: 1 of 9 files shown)


HashEngineering:
I started testing this and am finding these results when the information is returned to the sample app:
`protected void onActivityResult(final int requestCode=2, final int resultCode=0, final Intent data=null)`

resultCode != Activity.RESULT_OK, which is necessary to process the result from DashWallet

I debugged the Dash Wallet Testnet app, and it does get the request for the Master Public Key and it does return a result "tpub..."

This is what Dash Wallet creates to send back:
`Dash in-app payment sample://?callback=masterPublicKey&masterPublicKeyBIP32=tpubD927a2E4q26GrrsWB84ygWjvHQAMFiuLRBrXip5QoXRW3K68fQ2EDkocrF57bGHV4BG9ZcWtEJ8FD3V7ifGRUiQPTdiTUnVX9BKcTxDevQk&masterPublicKeyBIP44=null&source=Dash%20Wallet%20%5Btestnet%5D`

Am I doing something wrong.

tomasz-ludek:
Interesting, here is how it should work:

device-2018-01-24-133125

I will take a look at it.

tomasz-ludek:
Ok, it seems I know what is going on. You probably tested on API level 19 (or lower). It turns out, since WalletActivity is configured in the Manifest as singleTask, it can't handle startActivityForResult(...) correctly on Android below Lollipop (API level 21). More details here: https://stackoverflow.com/questions/8960072/onactivityresult-with-launchmode-singletask

I fixed this bug by moving the code responsible for handling Wallet Uris into dedicated Activity (WalletUriHandlerActivity). This solution is even better, since it works significantly faster (WalletUriHandlerActivity has no UI), and nicely separates the related code from rest of Wallet logic.

Sorry for inconvenience, I should tested it on lower Androids as well.

HashEngineering:
Yes, indeed, I was using an Android 4.4.4 device.  I will test these updates soon.

HashEngineering:
Now everything is working with your latest updates.

PR #73, opened 2018-01-31
Fix: Overlap on Main Screen (smaller devices)

sambarboza:
# Description
This PR Fixes the issue: https://github.com/HashEngineering/dash-wallet/issues/67

On smaller devices, the toolbar at the top of the screen that says DASH and has the balance has a few bugs. In some cases, depending on the precision of the amount of DASH, the DASH text will overlap with the D and the balance.

# Changes 
- Using different dimensions for text, margin and image size in toolbar according to device's DPI.

# Evidence


HashEngineering:
Tested on two devices and it worked.

PR #78, opened 2018-02-06
Improve security of handling `dashwallet://` uris

tomasz-ludek:
1. Added prompt in Dash Wallet when another app requests the master public key or an address using `dashwallet://...` uri.
2. Used wallet.freshReceiveAddress() instead of wallet.currentReceiveAddress() for address requests.
#77

### DIFF
--- a/wallet/AndroidManifest.xml
+++ b/wallet/AndroidManifest.xml
@@ -188,7 +188,7 @@
         <activity
             android:name="de.schildbach.wallet.ui.WalletUriHandlerActivity"
             android:label="@string/app_name"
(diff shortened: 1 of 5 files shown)


HashEngineering:
Can you place the new strings that were added into `wallet/res/values/strings.xml` into `wallet/res/values/strings-extra.xml`?

The reason we do this is to simplify merging from Bitcoin Wallet and it helps insure we don't delete strings by mistake.

tomasz-ludek:
Sure thing!
Done.

HashEngineering:
Review process has begun.

HashEngineering:
Review Complete.  Thank you for this addition.

PR #87, opened 2018-02-26
Removing delay when showing QR button in SendCoins screen

### LINKED ISSUE #85: Crash with KeyboardResponsiveCoordinatorLayout [After Sending Coins]
```
Samsung Galaxy S5 (kltevzw), 2048MB RAM, Android 6.0
LGE LG K7 (m1), 1024MB RAM, Android 5.1


java.lang.NullPointerException: 
  at de.schildbach.wallet.ui.widget.KeyboardResponsiveCoordinatorLayout$1$1.run (KeyboardResponsiveCoordinatorLayout.java:81)
  at android.os.Handler.handleCallback (Handler.java:739)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:158)
  at android.app.ActivityThread.main (ActivityThread.java:7224)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1230)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1120)
```

From an email Crash Report - from logs, appears to happen after sending coins:
```

=== device info ===

Device Model: SM-G900V
Android Version: 6.0.1
Android security patch level: 2016-06-01
ABIs: armeabi-v7a, armeabi

=== stack trace ===

java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.setVisibility(int)' on a null object reference
at de.schildbach.wallet.ui.widget.KeyboardResponsiveCoordinatorLayout$1$1.run(KeyboardResponsiveCoordinatorLayout.java:81)
at android.os.Handler.handleCallback(Handler.java:739)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:158)
at android.app.ActivityThread.main(ActivityThread.java:7224)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
```

sambarboza:
# Description - Fix for #85. 
- The app is crashing with a NPE when calling `viewToHide.setVisibility(View.VISIBLE)`, there is a 100ms delay when showing the QR button and this is the most probable cause of the issue. 
- Removing the delay causes an almost not noticeable flick of the button when it's being shown in the same time that the keyboard is being shown, however, to fix the crash is more important now while we can search for an alternative (non-trivial) in the future to have it not to flick without the possibility of a crash.

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/widget/KeyboardResponsiveCoordinatorLayout.java
+++ b/wallet/src/de/schildbach/wallet/ui/widget/KeyboardResponsiveCoordinatorLayout.java
@@ -67,20 +67,11 @@ public class KeyboardResponsiveCoordinatorLayout extends CoordinatorLayout {
         public void onGlobalLayout() {
             int difference = calculateDifferenceBetweenHeightAndUsableArea();
 
-            if (difference != 0) {
-                if (viewToHide != null) {
+            if (viewToHide != null) {
+                if (difference != 0) {
                     viewToHide.setVisibility(View.GONE);
-                }
-            } else {
-                if (viewToHide != null) {
-                    /*The delay is to avoid showing the scan button while the keyboard hide
-                    animation is running, which causes a visual glitch.*/
-                    viewToHide.getHandler().postDelayed(new Runnable() {
-                        @Override
-                        public void run() {
-                            viewToHide.setVisibility(View.VISIBLE);
-                        }
-                    }, 100);
+                } else {
+                    viewToHide.setVisibility(View.VISIBLE);
                 }
             }
         }


HashEngineering:
The scan floating button does flicker as it appears when the keyboard leaves the screen.  Nevertheless, let's take this fix and consider the nontrivial fix for the flicker as having a rather low priority.

PR #90, opened 2018-03-08
Lock Icon Feature

sambarboza:
# Description
- This feature mimics the feature in iOS. When user setup a spending PIN, their balance will only be shown after clicking on the lock icon and typing the spending PIN.

# Changes
- Showing Dialog to Setup spending pin on the first run.
- Added Lock icon to Transactions (Home), Send, Receive and Exchange screens.

# Evidence

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
+++ b/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
@@ -1032,6 +1032,10 @@ public final class SendCoinsFragment extends Fragment {
         attemptsRemainingTextView.setVisibility(View.GONE);
         final String pin = privateKeyPasswordView.getText().toString().trim();
 
+        if (pinRetryController.isLocked()) {
+            return;
+        }
+
         if (wallet.isEncrypted()) {
             new DeriveKeyTask(backgroundHandler, application.scryptIterationsTarget()) {
                 @Override
@@ -1044,7 +1048,7 @@ public final class SendCoinsFragment extends Fragment {
 
             setState(State.DECRYPTING);
         } else {
-            signAndSendPayment(null);
+            signAndSendPayment(null, pin);
         }
     }
 
--- a/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
+++ b/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
@@ -74,6 +74,7 @@ import de.schildbach.wallet.data.ExchangeRatesLoader;
 import de.schildbach.wallet.data.ExchangeRatesProvider;
 import de.schildbach.wallet.data.PaymentIntent;
 import de.schildbach.wallet.data.PaymentIntent.Standard;
+import de.schildbach.wallet.data.WalletLock;
 import de.schildbach.wallet.integration.android.BitcoinIntegration;
 import de.schildbach.wallet.offline.DirectPaymentTask;
 import de.schildbach.wallet.service.BlockchainState;
@@ -89,6 +90,7 @@ import de.schildbach.wallet.ui.InputParser.StringInputParser;
 import de.schildbach.wallet.ui.ProgressDialogFragment;
 import de.schildbach.wallet.ui.ScanActivity;
 import de.schildbach.wallet.ui.TransactionsAdapter;
+import de.schildbach.wallet.ui.UnlockWalletDialogFragment;
 import de.schildbach.wallet.util.Bluetooth;
 import de.schildbach.wallet.util.MonetarySpannable;
 import de.schildbach.wallet.util.Nfc;
@@ -96,6 +98,7 @@ import de.schildbach.wallet.util.WalletUtils;
 import de.schildbach.wallet_test.R;
 
 import android.app.Activity;
+import android.app.Dialog;
 import android.app.LoaderManager;
 import android.bluetooth.BluetoothAdapter;
 import android.content.AsyncTaskLoader;
@@ -1217,11 +1220,23 @@ public final class SendCoinsFragment extends Fragment {
     }
 
     private void handleEmpty() {
-        final Coin available = wallet.getBalance(BalanceType.AVAILABLE);
-        amountCalculatorLink.setBtcAmount(available);
+        final WalletLock walletLock = WalletLock.getInstance();
+        if (walletLock.isWalletLocked(wallet)) {
+            UnlockWalletDialogFragment.show(getFragmentManager(), new DialogInterface.OnDismissListener() {
+                @Override
+                public void onDismiss(DialogInterface dialog) {
+                    if (!walletLock.isWalletLocked(wallet)) {
+                        handleEmpty();
+                    }
+                }
+            });
+        } else {
+            final Coin available = wallet.getBalance(BalanceType.AVAILABLE);
+            amountCalculatorLink.setBtcAmount(available);
 
-        updateView();
-        handler.post(dryrunRunnable);
+            updateView();
+            handler.post(dryrunRunnable);
+        }
     }
 
     private Runnable dryrunRunnable = new Runnable() {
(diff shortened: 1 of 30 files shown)


HashEngineering:
I will review this in the next day or so.

This includes some spanish translations of new strings. We probably have to push those to Transifex, so they are part of that system.

On some of the files, I noticed that import statements are in different orders.  Is this due to software such as Android Studio reordering them?  This happens with me as well when comparing to Bitcoin Wallet and it is not something I purposefully changed.

HashEngineering:
After installing the app over a previous version, the Set Spending PIN dialog was displayed.  I pressed cancel 3 times and the dialog went away.  Then I added a Spending PIN, then the lock feature was activated.  Pressing the lock button allowed for entering the PIN, and then the transaction history and balance were displayed.

Closing and restarting the app activates the Lock.

This works as stated on Android 4.4.4, though the "Unlock to see your balance" dialog clips the text and the button (on my particular device, screen shots will be posted).  There is an UNLOCK and an OK Button (OK closes the dialog).  

Other notes (these were in the app already):
After setting the PIN, the change PIN dialog is displayed, then it disappears.
After removing the PIN, the Set Spending PIN dialog is displayed, then it disappears.

More information will be posted later.

sambarboza:
> This includes some spanish translations of new strings. We probably have to push those to Transifex, so they are part of that system.

I'll upload it soon.

> On some of the files, I noticed that import statements are in different orders. Is this due to software such as Android Studio reordering them? 

Yes, these changes are made by Android Studio, as Android Studio is the default IDE, if we use the command "optimize imports" when working in class files, we will always have the same order. Usually I use this command to remove unnecessary imports that were added in Development time. But, if you want I can try to revert the ordering and add the needed imports manually.

> After installing the app over a previous version, the Set Spending PIN dialog was displayed. I pressed cancel 3 times and the dialog went away.

I'll check this, dialog should appear once and only if a PIN was not setup.

> the "Unlock to see your balance" dialog clips the text and the button

Will check this.

> There is an UNLOCK and an OK Button (OK closes the dialog).

Can we remove tho Ok button?

> After setting the PIN, the change PIN dialog is displayed, then it disappears.
After removing the PIN, the Set Spending PIN dialog is displayed, then it disappears.

Will check this.

Thanks for your comments!

HashEngineering:
Excellent.  For code formatting, I wouldn't change anything.  Mainly I was curious.  I use Android Studio as well, but the Bitcoin Wallet developer uses Eclipse.  If we pull updates in the future, we can deal with merge conflicts at that time.

> Can we remove the Ok Button?

I say yes if there is a way to cancel the dialog.

HashEngineering:
The device is a Samsung Galaxy S4, Android 4.4.4.  This screenshot demonstrates the clipping observed on this device.

screenshot_2018-03-12-19-00-20

Removal of the OK button works great.  The dialog disappears when tapping elsewhere on the screen.

HashEngineering:
What about adding a max amount of tries for a time period, and an exponentially increasing waiting period between incorrect tries - regarding entering a PIN number.

HashEngineering:
We will need to display the Set Spending PIN dialog after:
1. restoring from backup file, if the backup file doesn't have a PIN code.
2. restoring from recovery phrase 

Not sure where the code will go for those two things, yet.  Both of these operations will get rid of the PIN code.

Additionally, as part of the BIP44 update, when the wallet is loaded and BIP44 support is added, we will need to add the code that will have the user reset the PIN code.  This is because the only the BIP32 part of the wallet will have a PIN set (if one is set).  To have the entire wallet encrypted, then we will need to have the user enter the old pin and a new pin.  This won't be part of this pull request.


I haven't yet repeated the "cancel 3 times" bug.

HashEngineering:
> After setting the PIN, the change PIN dialog is displayed, then it disappears.
> After removing the PIN, the Set Spending PIN dialog is displayed, then it disappears.

I think I determined the cause of this.  
`EncryptKeysDialogFragment.updateView` is called after the encryption state changes.

If setting the PIN, wallet.isEncrypted() is false and hides the old password field.   At the end of the encryption process, wallet.isEncrypted() is true and shows the old password field.

HashEngineering:
Excellent work.

sambarboza:
> "Unlock to see your balance" dialog clips the text and the button

This commit should fix this issue:
https://github.com/HashEngineering/dash-wallet/pull/90/commits/3afe081468c01ffa2b24e80fe4b379943b9602bf

HashEngineering:
Very good explanation.

HashEngineering:
https://github.com/QuantumExplorer/dashwallet/blob/master/DashWallet/BRWalletManager.m#L936

This code describes the timeout function in the iOS app.
After 2 failed attempts, there is no delay to enter again.
After 3 failed attempts, it will requires the user to wait 1 minute.
After 4 failed attempts, then 6 minutes, etc.

The exponential formula is to make the user wait 6^(n-3) minutes, where n is the number of failed attempts and is greater than 3.  

The app preferences would store a value for the number of failed attempts and the time of the first failed attempt.  These are cleared if the PIN is entered correctly.  There is also a reset button, which allows the user to enter a new recovery phrase.

HashEngineering:
The clipping issue is resolved.

HashEngineering:
After entering a recovery phrase or restoring from a recovery phrase, there will be no PIN:

This method (restoreWallet) is called under those situations:
https://github.com/HashEngineering/dash-wallet/blob/master/wallet/src/de/schildbach/wallet/ui/WalletActivity.java#L926

wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
```diff
@@ -1021,24 +1029,26 @@ else if (everythingPlausible())
...
                         application.backupWallet();
-                    signAndSendPayment(encryptionKey);
+                    signAndSendPayment(encryptionKey, pin);
                 }
-            }.deriveKey(wallet, privateKeyPasswordView.getText().toString().trim());
+            }.deriveKey(wallet, pin);
 
             setState(State.DECRYPTING);
         } else {
             signAndSendPayment(null);
```
HashEngineering: this needs to be changed based on the new method parameter `String pin` in `signAndSendPayment`.

HashEngineering:
After entering the PIN wrong 1 time and then restarting the app.  I entered the PIN wrong again and it said "6 attempts remaining".  Entered wrong again, still "6 attempts remaining".  Did this 2 more times.

Restarted app.  Entered PIN wrong, "6 attempts remaining".  Entered the PIN incorrectly many more times (more than 6), still stays "6 attempts remaining".  Then i tapped elsewhere to close the PIN dialog and then the unlock button.  Then the app crashed.
```

 java.lang.IllegalStateException: Fragment UnlockWalletDialogFragment{42d0adf0} not attached to Activity
	at android.app.Fragment.getResources(Fragment.java:767) ~[na:0.0]
	at android.app.Fragment.getText(Fragment.java:779) ~[na:0.0]
	at de.schildbach.wallet.ui.UnlockWalletDialogFragment$4.onBadPassword(UnlockWalletDialogFragment.java:167) ~[na:0.0]
	at de.schildbach.wallet.ui.CheckWalletPasswordTask$1$2.run(CheckWalletPasswordTask.java:51) ~[na:0.0]
	at android.os.Handler.handleCallback(Handler.java:733) ~[na:0.0]
	at android.os.Handler.dispatchMessage(Handler.java:95) ~[na:0.0]
  	at android.os.Looper.loop(Looper.java:146) ~[na:0.0]
	at android.app.ActivityThread.main(ActivityThread.java:5635) ~[na:0.0]
	at java.lang.reflect.Method.invokeNative(Native Method) ~[na:0.0]
	at java.lang.reflect.Method.invoke(Method.java:515) ~[na:0.0]
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1291) ~[na:0.0]
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1107) ~[na:0.0]
   	at dalvik.system.NativeStart.main(Native Method) ~[na:0.0]
```
This crash was not repeated.  But seems like I can enter the wrong PIN many times without any delays.

HashEngineering:
This code was merged into PR #100.

PR #100, opened 2018-04-06
Bip44 + Lock Features

HashEngineering:
Lock Feature was rebased on this branch
BIP44 was rebased on top

Last commit to fix the issue of adding BIP44 support to an encypted wallet.  Proof of Concept more than finished product!

PR #103, opened 2018-04-12
Preventing Upgrade Dialog to be dismissed on device rotation

sambarboza:


HashEngineering:
a0cac55 was cherry-picked onto #100.

Issue #104, opened 2018-04-27
Crash on Setting PIN (v5.18)


HashEngineering:
Today, 6:53 PM on app version 51800
LGE V20 (elsa), 4096MB RAM, Android 7.0
Report 1 of 1
```
java.lang.IllegalStateException: 
  at android.app.Fragment.getResources (Fragment.java:815)
  at de.schildbach.wallet.ui.EncryptKeysDialogFragment.updateView (EncryptKeysDialogFragment.java:358)
  at de.schildbach.wallet.ui.EncryptKeysDialogFragment.access$200 (EncryptKeysDialogFragment.java:65)
  at de.schildbach.wallet.ui.EncryptKeysDialogFragment$2.onShow (EncryptKeysDialogFragment.java:202)
  at android.app.Dialog$ListenersHandler.handleMessage (Dialog.java:1390)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6247)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:872)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:762)
```

HashEngineering:
This also happened on:
Apr 12, 8:57 AM on app version 51713 (5.17.6-lock-bip44) -- this was the Lock+BIP44 update that we were testing.
Motorola Moto G (5th Gen) (cedric), 2048MB RAM, Android 7.0

HashEngineering:
I was not able to reproduce this crash on my Android 7 device.

HashEngineering:
Two more reports, Android 6 and 8.

HashEngineering:
Possible Reason, just guessing is:  Fragment X not attached to Activity

Many solutions include using isAdded() before calling updateView.

HashEngineering:
Question:  Why would the Fragment not be attached when updateView() from onShowListener.onShow()?

sambarboza:
I couldn't reproduce this but submitted a PR (#109) that should prevent this from happening.

> Question: Why would the Fragment not be attached when updateView() from onShowListener.onShow()?

I think it's possible that in some cases the Activity will be killed/dismissed (by the system or by the user) between the call to show the dialog and the dialog being actually shown.

[...]

PR #109, opened 2018-05-09
Checking if DialogFragment is attached to activity.

sambarboza:
- This should address #104 by preventing updating the view when the dialog is not attached to the activity anymore.


HashEngineering:
Excellent! looks great.

PR #111, opened 2018-05-16
View Recovery Phrase Enhancements

HashEngineering:
* Show the recovery phrase when upgrading to BIP44/Set PIN (Lock)
* Add a backup status and time for the recovery phrase to the configuration and Crash Report
* Add a checkbox to the View Recovery Phrase Dialog for writing down the phrase

This checkbox is purposefully missing the first time the recovery phrase dialog is displayed during the upgrade process.  Later, the app will ask the user to backup (top message on the transactions list).  At that point, if the users clicks "backup" and shows the recovery phrase -- the user can check the box and will not be reminded again.  This is probably no the best way to do it, but it can help trigger our minds to determine a better way to remind the user twice.  For users that have upgraded, the reminder on the main screen will be the way to get them to write it down.

* Set a reminder to backup the recovery phrase on the main screen (TransactionsAdapter)
* Arm backup Reminders when restoring from a key backup file (old format)

### DIFF
--- a/wallet/res/layout/backup_wallet_to_seed_dialog.xml
+++ b/wallet/res/layout/backup_wallet_to_seed_dialog.xml
@@ -138,6 +138,12 @@
                 android:textColor="@color/fg_less_significant"
                 android:textSize="@dimen/font_size_small" />
 
+            <CheckBox
+                android:id="@+id/backup_wallet_seed_private_key_written_down"
+                android:layout_width="match_parent"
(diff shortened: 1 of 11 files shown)


HashEngineering:
@sambarboza Please have a look at this.

HashEngineering:
We will have the recovery phrase method as the default backup method.  I will work on that, rebase and commit it.

PR #114, opened 2018-05-23
Preventing Crash when Dialogs are recreated

sambarboza:
# Description
- This PR prevents crashes when Wallet is being Upgraded and the onboarding dialogs get recreated by the system (device rotation, app coming from bg with low memory, etc).


HashEngineering:
Approved.

PR #123, opened 2018-07-24
Detecting local currency

sambarboza:
# Description
- Detect user local currency based on information from phone network and suggests to change default local currency for users that are upgrading and update it automatically for new installs.
- Before this change, users in different countries would have it local currency set to either USD automatically.
- None of this data leaves the phone.

# Evidence


jasonhudgins:
Curious why you need to pull data from the network, couldn't you just use the users currently selected Locale?

HashEngineering:
Before testing, the default currency was changed from USD to AEN, then the apk was installed over the previous version using the testnet flavor.

I tested this on WiFi, which does not return a country code, so no actions are taken.

I am having troubles with my phone and USB to the computer, so no testing has been done on my phone.

HashEngineering:
On a device (tablet) with no mobile data support, this PR does nothing as there is no country code retrieved.

HashEngineering:
Is there a way for us to easily set the currency based on the wifi connected network or the device IP address for devices that have no SIM card?  Or do we run in to this problem?:

> The user selected locale could be different from the desired local currency for different reasons. Besides the different default locales set by the manufacture, users could change the locale to have it in a another language for instance.

sambarboza:
> Is there a way for us to easily set the currency based on the wifi connected network or the device IP address for devices that have no SIM card?
To have this information while using wifi we would have to use a 3rd party service. For that reason, this will work only for users with an active phone network (even without data connection).

PR #124, opened 2018-08-02
Show receiving value in fiat

sambarboza:
# Description
- Shows the receiving value in fiat in transaction list and notification.

# Evidence

### DIFF
--- a/wallet/src/de/schildbach/wallet/data/ExchangeRatesProvider.java
+++ b/wallet/src/de/schildbach/wallet/data/ExchangeRatesProvider.java
@@ -209,7 +209,10 @@ public class ExchangeRatesProvider extends ContentProvider {
                 cursor.getLong(cursor.getColumnIndexOrThrow(ExchangeRatesProvider.KEY_RATE_FIAT)));
         final String source = cursor.getString(cursor.getColumnIndexOrThrow(ExchangeRatesProvider.KEY_SOURCE));
 
-        return new ExchangeRate(new org.bitcoinj.utils.ExchangeRate(rateCoin, rateFiat), source);
+        ExchangeRate exchangeRate = new ExchangeRate(new org.bitcoinj.utils.ExchangeRate(rateCoin,
+                rateFiat), source);
+        WalletApplication.getInstance().setExchangeRate(exchangeRate);
+        return exchangeRate;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


wallet/src/de/schildbach/wallet/data/ExchangeRatesProvider.java
```diff
@@ -209,7 +209,10 @@ public static ExchangeRate getExchangeRate(final Cursor cursor) {
                 cursor.getLong(cursor.getColumnIndexOrThrow(ExchangeRatesProvider.KEY_RATE_FIAT)));
         final String source = cursor.getString(cursor.getColumnIndexOrThrow(ExchangeRatesProvider.KEY_SOURCE));
 
-        return new ExchangeRate(new org.bitcoinj.utils.ExchangeRate(rateCoin, rateFiat), source);
+        ExchangeRate exchangeRate = new ExchangeRate(new org.bitcoinj.utils.ExchangeRate(rateCoin,
+                rateFiat), source);
+        WalletApplication.getInstance().setExchangeRate(exchangeRate);
```
tomasz-ludek: @sambarboza we shouldn't initiate the _WalletApplication.exchangeRate_ field over here, it doesn't smell good... ;)
I will take a closer look at it and try to propose something more reliable.

HashEngineering:
1. The rest of the app uses `ExchangeRatesProvider` to get the exchange rate.  

In most cases it uses `LoaderManager.LoaderCallbacks<Cursor>` and in the `onLoadFinished` method it will access the rate with `ExchangeRatesProvider.getExchangeRate(data);` and do something with it like `updateView`.

What I don't like about this one is that the loaders are with activities that are on for a period of time and the user can modify some of the contents (like amounts).  We are needing the exchange rate in a service.

2. Simpler answer, but might contain old data:   `Configuration.getCachedExchangeRate`

3.  WalletBalanceWidgetProvider.java, line 115.  This might work.

We look forward to your ideas @tomasz-ludek .

tomasz-ludek:
Sorry for the delay guys!
In my opinion, to keep the code consistent, the best approach would be to apply the Loader-based way in transaction list, and _Configuration.getCachedExchangeRate_ way (mentioned by @HashEngineering) for notifications.
What do you think?

HashEngineering:
Wow, we are late on replying.  We should use `Configuration.getCachedExchangeRate` when the transaction is received.  After that the transaction will always have the transaction value.

The notification message displayed after the tx is received and it will have a fiat value.

The Transaction List will eventually display the tx and it will have a fiat value.  I don't think there is away for the app to receive a tx that skips the part above where the ExchangeRate is assigned with `Configuration.getCachedExchangeRate`, therefore there is no need for the Transaction List to have a loader for getting exchange rates.

sambarboza:
The PR was update with the following changes:
- Using exchanged rate stored in `Config` class instead of initializing a field in the application class.
- Not setting the exchange rate in the transaction when the blockchain is being reset/rescan or the tx is more than 30 minutes away from last blockchain head time to avoid setting a fiat price that may have changed drastically from the moment it was sent.

HashEngineering:
When this is merged it will be squashed to 1 commit.

PR #127, opened 2018-08-12
Transaction State Messages

HashEngineering:
There are two objectives:

Notify the user if a transaction was sent to only 1 peer (due to being connected to only 1 peer). Previously the app stated that "the payment has not been transmitted".

Simplify the various transaction statuses to include simplier InstantSend and allowed 0-conf (this is not yet included)

### DIFF
--- a/wallet/build.gradle
+++ b/wallet/build.gradle
@@ -23,7 +23,7 @@ dependencies {
     compile 'com.android.support:support-core-utils:25.1.1'
     compile 'com.android.support:recyclerview-v7:25.1.1'
     compile 'com.android.support:cardview-v7:25.1.1'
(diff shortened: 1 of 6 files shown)


HashEngineering:
This will fail travis tests, because it does it was not rebased to the current master.

HashEngineering:
This has passed QA.

HashEngineering:
This was rebased and the first commit was worded to correct an error `dashj` replaced `dash`

HashEngineering:
This was rebased but still failed the CI tests due to a missing dashj 0.14.4.6 from jcenter().  I can get that published in the next day or two.  Lets make sure this passes the CI tests before merging.

HashEngineering:
Tests now pass.

PR #129, opened 2018-08-31
Obtain the VES exchange rate from coinhills.com to replace VEF

HashEngineering:
This will use the api from Coinhills.com to obtain the VES exchange rate, then it will remove the VEF exchange rate obtained from BitcoinAverage.com.

Question(s):  
1.  Will it be a problem to remove an exchange rate that people may be using in the app?  Perhaps the app returns to the default of USD?
ves


sambarboza:
I've tested this branch and it works.
> Will it be a problem to remove an exchange rate that people may be using in the app? Perhaps the app returns to the default of USD?

Not a problem. If you have VEF as default currency and updates to this version it will show the price in USD.

HashEngineering:
Thanks.  That means we are good to merge this.

PR #131, opened 2018-09-07
Update localizations

HashEngineering:
All localisations were updated, but separated into two commits (complete / incomplete).

pt-BR was "Renamed" to pt.


HashEngineering:
Excellent.

PR #132, opened 2018-09-14
Use Masternode List as Seed Peers if DNS seeders return no peers

HashEngineering:
The List of seed peers for MainNet and TestNet was added to dashj 0.14.4.7.

If the DNS seeders don't return any peers, dashj throws a PeerDiscoveryException.  The code added catches this exception and attempts to connect to the Seed Peers (about 300 masternodes).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ buildscript {
 
 allprojects {
     repositories {
+        mavenLocal()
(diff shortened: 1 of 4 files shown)


HashEngineering:
The CI will fail because DashJ 0.14.4.7 has not been published to jcenter.

PR #133, opened 2018-09-17
Get VES from LocalBitcoins.com instead of coinhills.com

HashEngineering:
Use LocalBitcoins.com


HashEngineering:
This was included in #135

PR #134, opened 2018-09-17
Remove remaining RBF functionality

HashEngineering:
There was only 1 place in the code that relates to RBF.

RBF = replace by fee, a feature of bitcoin that is not part of Dash.


HashEngineering:
This was included in #135 .

PR #135, opened 2018-09-18
v5.18.4 Changes (fixes for connectivity)

HashEngineering:
This was originally to:
Revert API28 changes and finalize version 5.18.3

However, now this includes other commits for 5.18.4.


sambarboza:
Tested on Android 7 & 8 devices. 
It works for me.

PR #136, opened 2018-09-24
Rebranding

sambarboza:
# Rebranding 

## Changes
- Applied new Dash logo and branding.
- Support Library and Gradle version updated to support custom font.


HashEngineering:
**RequestCoinsFragment**  Android  8:  Emulator 
- Keyboard covers up part of the QR code, which does not happen in current version.  Perhaps this is because of the "syncing" message box at the top of the activity?
- The text "Or tap an NFC enabled device" is missing.  Perhaps we don't need it.  I never really noticed it was there.
- The checkbox for accepting payment by bluetooth is missing.  Perhaps we don't really need this.  Perhaps nobody uses it.  Or we can have it as an option in Settings?

screenshot_1537925514

HashEngineering:
Settings.

Pressing back on a settings screen always goes back home (the main screen).  The current version goes back to the previous menu (the Settings, Diagnostics, About page).  This may not be that important.  Users may not normally be looking at every settings menu, except for the first time.

HashEngineering:
Android 9:  The keyboard showing up after entering the PIN occurred once and each time the app was minimized and restored.  After the app was closed and restarted, then the keyboard was missing.

sambarboza:
https://github.com/HashEngineering/dash-wallet/pull/136/commits/571b205a5dc89fdc0cd12285d040dcc2b611f537

This commit address the keyboard issue in Android 9 and decreases the margin of the QR in the `RequestCoinsFragment`.

The QR will inevitably be covered by the keyboard in really small screens, which also happens in the current version, however, the screen is scrollable.

HashEngineering:
Checking now...

HashEngineering:
Reviewed all changes.

PR #137, opened 2018-09-26
Update block explorer lists

HashEngineering:
* Set insight as the default for mainnet and testnet
* Remove obsolete block explorers
* Put testnet list in testnet res/values.xml


HashEngineering:
Excellent.  This can be merged after the rebranding.

HashEngineering:
BitinfoCharts removed because it does not browse the correct block.  it does not take a block hash as a parameter to view a block.

PR #138, opened 2018-09-29
Add Wallet Lock Timer

HashEngineering:
This adds a 3 minute "timer" based on the secure time when a PIN is entered correctly.

This means that for 3 minutes the app will remained unlocked.  After 3 minutes the app will lock if the user switches to a new app or the device locks itself.  Not sure about the exact timing as I don't know how often the secure time value is updated in the app.  

The 3 minute timer is activated if:
1.  The app is unlocked on any activity with the unlock function
2.  The user sends coins (which requires a PIN) which now unlocks the app.

Other functions that require a PIN do not unlock the app, such as Backup and Viewing the recovery phrase.

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
+++ b/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
@@ -1091,7 +1091,7 @@ public final class SendCoinsFragment extends Fragment {
             @Override
             protected void onSuccess(final Transaction transaction) {
                 if (pin != null) {
-                    pinRetryController.successfulAttempt();
+                    pinRetryController.successfulAttempt(config);
                 }
                 sentTransaction = transaction;
 
--- a/wallet/src/de/schildbach/wallet/Configuration.java
+++ b/wallet/src/de/schildbach/wallet/Configuration.java
@@ -74,6 +74,7 @@ public class Configuration {
     private static final String PREFS_KEY_LAST_BACKUP_SEED = "last_backup_seed";
     public static final String PREFS_KEY_INSTANTX_ENABLED = "labs_instantx_enabled";
     public static final String PREFS_KEY_LITE_MODE = "labs_lite_mode";
+    public final static String PREFS_LAST_UNLOCK_TIME = "last_unlock_time";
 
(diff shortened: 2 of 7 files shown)


wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
```diff
@@ -1091,7 +1091,7 @@ else if(Constants.ENABLE_ZERO_FEES && feeCategory == FeeCategory.ECONOMIC || fee
             @Override
             protected void onSuccess(final Transaction transaction) {
                 if (pin != null) {
-                    pinRetryController.successfulAttempt();
+                    pinRetryController.successfulAttempt(config);
```
sambarboza: I'm not sure if the user will expect his wallet to be "unlocked" after using the PIN to send a transaction.
I personally would not.

sambarboza:
Added a comment about unlocking the wallet when sending a transaction:
https://github.com/HashEngineering/dash-wallet/pull/138/files#r221829556

In general, very clean implementation. Great work.

wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
HashEngineering: I see your point.  No other app does it this way.  Perhaps it could be a setting where the user must choose that behavior.  I haven't heard of any user make comments about this part either way (for or against).  Another story can handle further customization of this feature at a later time.

HashEngineering: Nevertheless, I thought it would be useful for the user to see their transaction list without unlocking after sending.

HashEngineering:
Lets get @tomasz-ludek's feedback and I will remove the Send Unlock aspect mentioned above.

wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
sambarboza: Didn't think about it. But I agree. I think it's safe to say that users expect to see the tx in the list after sending. 
If @tomasz-ludek agree I think we can let this as it is implemented.

PR #139, opened 2018-10-03
Use dashj 0.14.7

HashEngineering:
This will fail the Travis CI test due to a missing dashj-core-0.14.7.

This will also have merge conflicts with the Rebranding PR#136.

### DIFF
--- a/wallet/build.gradle
+++ b/wallet/build.gradle
@@ -27,7 +27,7 @@ dependencies {
     implementation "com.android.support:support-core-utils:$androidSupportVersion"
     implementation "com.android.support:recyclerview-v7:$androidSupportVersion"
     implementation "com.android.support:cardview-v7:$androidSupportVersion"
-    implementation 'org.dashj:dashj-core:0.14.4.7'
+    implementation 'org.dashj:dashj-core:0.14.7'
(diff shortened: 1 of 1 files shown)


HashEngineering:
This will be updated for v6.1.1 changes.

HashEngineering:
Code has been updated to work with v6.0.0 and above.  Travis CI will still fail because JCenter doesn't have dashj 0.14.7 yet.

sambarboza:
Tested:
- Recovering wallet from backup file
- Sending/Receiving
- InstantSend 
Everything worked fine.

One issue that I found is that the network monitor frequently showed an unusual list of connected peers, usually when coming back to the app after some time in background with the device locked, some of the peers was repeated 2 or 3 times in the list, please see attached screenshot.

screenshot_20181105-173211_dash wallet

HashEngineering:
This is a problem that I have seen many times over the years with similar types of wallets.  It may be a result in a bug with the UI, where it doesn't clear out previous entries in the list before reloading the list.

HashEngineering:
This is not an easy bug to duplicate.  Let's create an action item to fix it later.

PR #142, opened 2018-10-09
 Fix for archiving during backup to file process

HashEngineering:
The Bitcoin Wallet devs solved this a while ago.


HashEngineering:
This must be tested on Android 8 to begin with and ensure that it works on other versions.  QA will test android 6.0 for now.  if we verify that it works for android 8, then we can merge it.

PR #144, opened 2018-10-18
Update Screenshots with new Design

sambarboza:


HashEngineering:
I will squash this PR since the second commit fixes a mistake in the first.

We need to add these output.json files to `.gitignore`.

PR #145, opened 2018-10-27
Fix wallet wipe

HashEngineering:
Purpose:
Prevent Brute forcing by adding the code to reset the wallet after the PIN is entered incorrectly 8 times and also to allow for logging to show that the 

In WalletApplication, I removed the delete all files methods and replaced it with:
1.  Create a new wallet (nonencrypted)
2.  Replace current wallet (user forgot their PIN that encrypts it) with new wallet (nonencrypted)

This method will preserve all settings and logs (for diagnosis of problems and so we will know that the user wiped the wallet).

The other code that closes the app remains the same.  The user must restart the app and it will be as if it was a fresh install.  The user will need to enter the mandatory PIN.

The second part of this is to add the code that will wipe the wallet after the PIN is entered incorrectly 8 times.

**Testing:**
Install the app with this PR.
Enter the PIN wrong 4 times to see the dialog that allows reset.
Go to the Settings on the android and set the time ahead to skip waiting between entering incorrect PIN's.
On the 8th attempt, ensure that the app doesn't allow canceling the Reset Dialog.  Ensure that there is not another way to reenter the PIN to keep trying.  The app should require a reset.

### DIFF
--- a/wallet/res/values/strings-extra.xml
+++ b/wallet/res/values/strings-extra.xml
@@ -40,7 +40,7 @@
 	<string name="wallet_lock_unlock_dialog_title">Unlock your wallet</string>
 	<string name="wallet_lock_unlock_dialog_message">Your wallet is protected by a spending PIN. Enter your PIN and press Unlock to continue.</string>
 	<string name="wallet_lock_reset_wallet_title">Reset wallet</string>
-	<string name="wallet_lock_reset_wallet_message">Are you sure? All stored data will be deleted and your wallet will only be recoverable by a backup file or recovery seed.</string>
+	<string name="wallet_lock_reset_wallet_message">Are you sure? Your wallet will be deleted and all your Dash will be lost.\n\nThe only way to recover your Dash is with a backup file or recovery phrase.</string>
 
 	<string name="wallet_lock_wallet_disabled">Wallet disabled</string>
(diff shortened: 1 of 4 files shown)


HashEngineering:
New Reset Dialog:

screenshot_1540650814

HashEngineering:
Question:  When the user enters the wrong PIN 8 times, should they get a dialog like the above?  The previous version had a toast message instead.

If Yes, is the above text good enough.  They are not given a choice in that case.

HashEngineering:
I see what you mean.  The "Reset Wallet" dialog is triggered by entering the wrong PIN.  The app doesn't check to see if max tries has been attempted before allowing the user to enter the PIN.

HashEngineering:
This still has a bug:  restarting the app manually will prevent some settings from being saved and the user can try to enter the PIN again.

HashEngineering:
The bug was a result of this:
https://stackoverflow.com/questions/17469583/setstring-in-android-sharedpreferences-does-not-save-on-force-close
https://issuetracker.google.com/issues/36943216

https://github.com/HashEngineering/dash-wallet/pull/145/commits/1a63b56268d8cda6faede311159c5527ac229ae2 includes the fix for this bug.  The HashSet is copied before saving.

PR #148, opened 2018-11-02
Transifex: Add common and uphold-integration strings

HashEngineering:
`tx push -s` was tested and works.

Lets wait until `pull` is used before merging.


AllenChong:
pull

PR #149, opened 2018-11-13
Lower fees on normal transactions (don't require minimum fee of 0.00001)

HashEngineering:
ECONOMIC and NORMAL fees are set to the minimum allowed by Dash Core

InstantSend fees must be higher than the minimum fee

Disable constant for Zero Fees, which are no longer allowed.

---------------
This has only been tested by looking at the Send Coins screen, not with sending.


HashEngineering:
My testing is finished and results are successful on a variety of transaction sizes below 1 KB.

sambarboza:
Normal transactions (non-IS) worked fine for me and fees are lower than before:
- 1 input, 191 bytes, fee: 0.00000193
http://insight.dash.org/insight/tx/56fb4f6f6dcaa9e41aac4d68c7bdcf8b044d34f36fac53f5aa14da3bdf6e1215
- 2 inputs, 339 bytes, fee: 0.00000342
http://insight.dash.org/insight/tx/6cff315856c08bdd162b68028ae78690572476515791efaf6b42fe378f47496d
- 5 inputs, 782 bytes, fee: 0.00000789
http://insight.dash.org/insight/tx/19a0c842e576e44618c158770fdc089832771281398b5d6d43a03286564764c4

Fees for InstantSend transactions seems to be 10x higher:
- IS Tx using Android Wallet:
http://insight.dash.org/insight/tx/cbf7731761c7722d6978e8fefcb76c2369fb6ee04a3b84f6eade0d734901936d
- IS Tx using Core Wallet:
https://insight.dashevo.org/insight/tx/3f2fef3326c661d9dfbcbbb94e321ff0cc6a234ecaa3f625bde8d63519a3dbd7

HashEngineering:
Looks your feedback reveals that with `Empty Wallet`, InstantSend is using the old fees of 0.001.  This is a bug in DashJ, not the PR.  

I recall hearing of one report of this issue, where InstantSend fees were higher than they should have been.

The Line that has to be fixed is here:
https://github.com/HashEngineering/dashj/blob/master/core/src/main/java/org/bitcoinj/wallet/Wallet.java#L4150

fix is this:              
```java
fee = TransactionLockRequest.MIN_FEE.div(params.isDIP0001ActiveAtTip() ? 10 : 1).multiply(tx.getInputs().size());
```

PR #150, opened 2018-11-15
Added foreground service permission in Manifest

sambarboza:
- This is a requirement since Android 9 for apps that execute foreground services.


HashEngineering:
This same fix was upstream.
https://github.com/bitcoin-wallet/bitcoin-wallet/commit/8d60481c4e1cfb0f16b89da6b9838e79d695f593#diff-490eb0445919646316db395eb69f2d25

PR #154, opened 2018-11-28
Unlock with fingerprint

sambarboza:
# Description
Users on devices with fingerprint sensor and Android Fingerprint API implemented by the manufacturer* will be able to unlock the wallet using fingerprint as alternative to the PIN.

- For new installations, after setting a new PIN a new Dialog is shown, on this dialog the user has 3 options:

     - Touch the fingerprint sensor: the PIN will be encrypted by the fingerprint data and stored in Android KeyStore. After enabling fingerprint the user can unlock the wallet touching the device's fingerprint sensor, on "Send Screen", touching the fingerprint is an alternative to entering the PIN.
     -  Tap on "Remind later" dialog button: the user will be reminded to enable fingerprint the next time the wallet is unlocked with PIN.
     - Tap on "Don't remind me" dialog button: the user will not be reminded about setting fingerprint unless he taps the "Enable fingerprint" option in the "Safety" menu.

- Unlocking the wallet with fingerprint is available from the home screen or by tapping the "Lock" icon in the toolbar.
- After enabling fingerprint, the user will be asked to touch the fingerprint sensor when changing PIN.

- The user can user the fingerprint on the "view recovery phrase" dialog.
-  For users that are updating from a previous version, a dialog to enable fingerprint will be shown after the first time the wallet is unlocked.

*Some devices may have fingerprint sensor hardware but won't implement Android Fingerprint API, this feature will not be available on such devices. Known devices with fingerprint sensor but lack of Android Fingerprint API implementation:

Xiaomi Redmi 3

# Changes
- Added `EnableFingerprintDialog` class.
- `EnableFingerprintDialog` is shown after setting PIN on a new installation or after the wallet is unlocked for the first time after updating to a version with this code.
- Added `FingerprintHelper' class.
- Piggyback: https://github.com/HashEngineering/dash-wallet/commit/d9e2fcff8ea751f8eb246cf531d8f3e727b3aa2f - Replaced deprecated Fragment and Loader calls with calls to support library methods. This was sent in a separate commit to facilitate code review.


HashEngineering:
I was surprised to hear about the depreciation of  `Fragment` and `Loader`

Bitcoin Wallet migrated to `androidx` from the support libraries, 19 days ago:
https://github.com/bitcoin-wallet/bitcoin-wallet/commit/1886af9895b5a5931d9af6417efc3f57cd7dd92b

But in March 2018, they migrated to the support libraries:
https://github.com/bitcoin-wallet/bitcoin-wallet/commit/12f8f1633895d08aba46e893315245cd06dec769#diff-d4ae7dc33e88f5c498d1e9b68a80aeeb

I approve of this change and will review/test the rest of the code later.

sambarboza:
https://github.com/HashEngineering/dash-wallet/pull/157 replaces this PR removing https://github.com/HashEngineering/dash-wallet/commit/d9e2fcff8ea751f8eb246cf531d8f3e727b3aa2f which will be sent in a separate PR.

PR #160, opened 2019-01-07
Using custom tabs/external browser for Uphold authentication.

sambarboza:
# Description
- Uphold Integration login was failing in some scenarios which required the user to open a link to continue the authentication process. The integration was using a WebView for login and opening the link on the WebView wasn't possible.
- To fix this, [Chrome Custom Tabs](https://developer.chrome.com/multidevice/android/customtabs) was implemented to handle Uphold Login / Buy Dash screens, an external browser will be used as fallback when a browser which doesn't support Chrome Custom Tabs is available.


HashEngineering:
I began the testing, but since had to reset my Authy account since it is not working on any of my phones.  Therefore I can't finish the sign in and authorization process for 1-2 days to finish testing.

sambarboza:
@HashEngineering conflict fixed.

PR #168, opened 2019-01-27
Remove RaiseFeeDialog and associated commands and strings

HashEngineering:
Translations were not removed.  They will be removed on future updates from Transifex.

DMD-828


HashEngineering:
I found more references to raising the fees in the TransactionAdapter, I will submit a commit to remove them.

HashEngineering:
Since this has been approved, I will not submit a commit for removing those things from TransactionsAdapter.

PR #177, opened 2019-02-04
Handle dust outputs when ensureMinRequiredFee=false

HashEngineering:
I found a bug with `SendRequest.ensureMinRequiredFee = false` -- it can result in dust for outputs which are rejected by the network.  Therefore, we need to check for dusty outputs after we complete the transaction `wallet.completeTx` with `TransactionOutput.isDust` and if dust is found, then recreate the transaction with `SendRequest.ensureMinRequiredFee = true`

DMD-882

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
+++ b/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
@@ -1382,7 +1382,7 @@ public final class SendCoinsFragment extends Fragment {
 
                 wallet.completeTx(sendRequest);
                 if(checkDust(sendRequest)) {
-                    sendRequest = createSendRequest(finalPaymentIntent, RequestType.INSTANT_SEND_AUTO_LOCK, false, true);
+                    sendRequest = createSendRequest(finalPaymentIntent, RequestType.REGULAR_PAYMENT, false, true);
                     wallet.completeTx(sendRequest);
                 }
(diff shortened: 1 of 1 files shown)


wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
```diff
@@ -1372,9 +1378,13 @@ private void executeDryrun() {
 
             try {
                 // check regular payment
-                final SendRequest sendRequest = createSendRequest(finalPaymentIntent, RequestType.REGULAR_PAYMENT, false);
+                SendRequest sendRequest = createSendRequest(finalPaymentIntent, RequestType.REGULAR_PAYMENT, false, false);
 
                 wallet.completeTx(sendRequest);
+                if(checkDust(sendRequest)) {
+                    sendRequest = createSendRequest(finalPaymentIntent, RequestType.INSTANT_SEND_AUTO_LOCK, false, true);
```
tomasz-ludek: Should't be `RequestType.REGULAR_PAYMENT` instead of `RequestType.INSTANT_SEND_AUTO_LOCK` here?

HashEngineering: Yes, copy/paste error.

HashEngineering: This was repaired in 15832ce

PR #182, opened 2019-02-26
Restore backup file crash

tomasz-ludek:
Fix for [DMD-908 Android: Fix crash from PIN entry after restore backup file](https://dashpay.atlassian.net/browse/DMD-908).
It was caused by the lack of fingerprint support in EncryptNewKeyChainDialogFragment. This dialog is being shown when restoring the wallet from an old backup file of not "upgraded" wallets that were secured by the PIN but didn't use BIP44.


HashEngineering:
This was merged manually as d86c034

PR #183, opened 2019-02-26
Preventing from NPE when the fingerprint error occurred

tomasz-ludek:
I'm unable to reproduce this issue but was able to fix it based on static code analysis.
In rare cases calling the {{fingerprintHelper.savePassword}} method can cause the FingerprintHelper.Callback() is fired immediately, therefore the initialization of _fingerprintView_ in } {{EnableFingerprintDialog.onCreateDialog}} had to be performed before calling {{fingerprintHelper.savePassword}}.
Also, I added some additional logs in order to easily identify the cause of {{fingerprintHelper.savePassword}} to fail in the future.


HashEngineering:
Excellent solution.

tomasz-ludek:
Done

PR #184, opened 2019-03-05
Use dashj 0.15.1

HashEngineering:
Does not connect to older nodes (pre 0.13.0)


HashEngineering:
This will fail travis CI since dashj 0.15.1 is not published.

HashEngineering:
This was manually merged 9e1866ff

PR #185, opened 2019-03-07
DMD-901, 907 - Fix Missing Fiat Values and Duplicated Fiat Values

HashEngineering:
Address two issues:
1.  missing fiat values:  now we will show fiat values for all incoming transactions (pending) and transactions that are less then 3 hours old that are in blocks.  we will put a message in the transaction if the fiat value cannot be obtained (a received transaction doesn't follow 1)
2.  duplicate values - we will erase the previous value in the view and then in the case of a null exchangeRate, assign a value of "----".

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
+++ b/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
@@ -157,7 +157,7 @@ public class BlockchainServiceImpl extends android.app.Service implements Blockc
     private static final int MAX_HISTORY_SIZE = Math.max(IDLE_TRANSACTION_TIMEOUT_MIN, IDLE_BLOCK_TIMEOUT_MIN);
     private static final long APPWIDGET_THROTTLE_MS = DateUtils.SECOND_IN_MILLIS;
     private static final long BLOCKCHAIN_STATE_BROADCAST_THROTTLE_MS = DateUtils.SECOND_IN_MILLIS;
-    private static final long TX_EXCHANGE_RATE_TIME_THRESHOLD_MS = TimeUnit.MINUTES.toMillis(30);
+    private static final long TX_EXCHANGE_RATE_TIME_THRESHOLD_MS = TimeUnit.MINUTES.toMillis(180);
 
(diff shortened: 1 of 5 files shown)


wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -181,11 +181,12 @@ public void onCoinsReceived(final Wallet wallet, final Transaction tx, final Coi
             long blockChainHeadTime = blockChain.getChainHead().getHeader().getTime().getTime();
             boolean insideTxExchangeRateTimeThreshold = (now - blockChainHeadTime) < TX_EXCHANGE_RATE_TIME_THRESHOLD_MS;
 
-            if (tx.getExchangeRate() == null && !replaying && insideTxExchangeRateTimeThreshold) {
+            if (tx.getExchangeRate() == null && ((!replaying || insideTxExchangeRateTimeThreshold) || tx.getConfidence().getConfidenceType() == ConfidenceType.PENDING)) {
```
HashEngineering: Pending transactions should always get assigned an exchange rate.

wallet/src/de/schildbach/wallet/ui/TransactionsAdapter.java
```diff
@@ -514,17 +515,26 @@ private void bind(final Transaction tx) {
...
             final ExchangeRate exchangeRate = tx.getExchangeRate();
+            fiatView.setAmount(null);  //clear the exchange rate first
             if (exchangeRate != null) {
                 fiatView.setFormat(Constants.LOCAL_FORMAT.code(0,
                         PREFIX_ALMOST_EQUAL_TO + exchangeRate.fiat.getCurrencyCode() + " "));
                 Coin absCoin = Coin.valueOf(Math.abs(txCache.value.value));
                 fiatView.setAmount(exchangeRate.coinToFiat(absCoin));
+            } else {
+                fiatView.setText(PREFIX_ALMOST_EQUAL_TO + WalletApplication.getInstance().getConfiguration().getExchangeCurrencyCode() + " ----");
             }
```
HashEngineering: Above:  set the amount to null to clear previous values -- this may not be necessary because we will set a value regardless of the value of exchangeRate (null or not null).

PR #188, opened 2019-03-20
DMD-812 Simplify Settings Menu

HashEngineering:
This will remove the following from the Settings / Settings page: 

1. Enable InstantSend - This was added to be able to turn off the ability to InstantSend when the feature was first added 3 years ago. It is no longer an experimental feature, we can remove the setting.
2. Enable Lite Mode - This was added along with InstantSend to allow more InstantSend verification. If the setting was off, the app would download the masternode list and use that list to verify instantsend lock votes. The verification was not reliable due to various limitations addressed by Dash Core 0.13. At this point, there is no lite mode for the app with the coming Deterministic Masternode lists, therefore we can remove this setting.
3. Show disclaimer will be moved out of the Labs section, but be kept.
4. Look up wallet names. This feature allows one to look up a wallet name from the DNS system. I don't know what that means. Most likely no software knows how to do this, I don't know how this app does it. I doubt anyone in the Dash Community uses it. We should remove this setting. This will also remove a dependency library that supplies the functionality.

Additionally the BIP70 functionality will be re-enabled. It was mistakenly disabled during a QR code bug in 2017.


HashEngineering:
These are the old settings:
old-settings

These are the new settings:
new-settings

HashEngineering:
Merge conflict is fixed and I removed the extra white space changes.

PR #190, opened 2019-03-21
[DMD-952] Backup to File Crash after requesting permission (Android 6)

tomasz-ludek:
Fixed the crash on Android 6 when BackupWalletDialogFragment.show() method is called before WalletActivity.onResume() causing IllegalStateException or RuntimeException to be thrown.

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/WalletActivity.java
+++ b/wallet/src/de/schildbach/wallet/ui/WalletActivity.java
@@ -148,6 +148,8 @@ public final class WalletActivity extends AbstractBindServiceActivity
 
     private CanAutoLockGuard canAutoLockGuard;
 
+    private boolean showBackupWalletDialog = false;
+
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -300,6 +302,14 @@ public final class WalletActivity extends AbstractBindServiceActivity
 
         checkLowStorageAlert();
         detectUserCountry();
+        showBackupWalletDialogIfNeeded();
+    }
+
+    private void showBackupWalletDialogIfNeeded() {
+        if (showBackupWalletDialog) {
+            BackupWalletDialogFragment.show(getSupportFragmentManager());
+            showBackupWalletDialog = false;
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -486,7 +496,7 @@ public void onDismiss(DialogInterface dialog) {
 
         if (ContextCompat.checkSelfPermission(this,
                 Manifest.permission.WRITE_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED)
-            BackupWalletDialogFragment.show(getSupportFragmentManager());
```
HashEngineering: Since this line is removed: `BackupWalletDialogFragment.show(getSupportFragmentManager());` the backup dialog is not shown until onResume() is called.

tomasz-ludek: Good point @HashEngineering, I've just fixed this issue.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -342,7 +352,7 @@ public void onRequestPermissionsResult(final int requestCode, final String[] per
                                            final int[] grantResults) {
         if (requestCode == REQUEST_CODE_BACKUP_WALLET) {
             if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED)
-                handleBackupWallet();
+                showBackupWalletDialog = true;
```
HashEngineering: handleBackupWallet() requires that the wallet is unlocked before proceeding.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -300,6 +302,14 @@ public void run() {
 
         checkLowStorageAlert();
         detectUserCountry();
+        showBackupWalletDialogIfNeeded();
+    }
+
+    private void showBackupWalletDialogIfNeeded() {
+        if (showBackupWalletDialog) {
+            BackupWalletDialogFragment.show(getSupportFragmentManager());
```
HashEngineering: by calling BackupWalletDialogFragment.show instead of handleBackupWallet(), we skip the step of checking that the wallet is unlocked (if locked, the user enters the PIN).

Can you put handleBackupWallet() here instead?

PR #194, opened 2019-03-29
[DMD-755] Fix the app shortcuts

tomasz-ludek:
Fixed the issue with the app from shortcuts - in never versions of Gradle the shortcut.xml have no access to ${applicationId} variable so it has to be hardcoded for each flavor. Also fixed the shortcuts icons.


HashEngineering:
Excellent.

PR #196, opened 2019-04-05
DMD-1003 Autolock Announcment Dialog Crash - hotfix 1

tomasz-ludek:
Added missing `super.onActivityResult(requestCode, resultCode, intent)` call to avoid `IllegalStateException `when showing the dialogs after `onSaveInstanceState` and other potential issues with WalletActivity.


HashEngineering:
This change is in Bitcoin Wallet as well.

https://github.com/bitcoin-wallet/bitcoin-wallet/blame/master/wallet/src/de/schildbach/wallet/ui/WalletActivity.java#L379

PR #208, opened 2019-06-03
DMD-1049 Scan Crash (from launcher icon)

tomasz-ludek:
Moved the initialization of WalletLock to WalletApplication in order to avoid crashes when using it in activities not derived from AbstractWalletActivity.


HashEngineering:
This PR we need to keep, but one change needs to be made in WalletLock() where the configuration is assigned from the WalletApplication singleton.

tomasz-ludek:
Done

HashEngineering:
This was merged manually here:  https://github.com/HashEngineering/dash-wallet/commit/b40e275663eb969930b244faccb765848c7ce2f9

PR #209, opened 2019-06-03
DMD-1048 Set PIN Crash Fix after Restoring a Recovery Phrase

tomasz-ludek:
Checking whether getArguments() isn't null before requesting for CANCELABLE_ARG argument in order to avoid app crashes when showing the dialog differently than using show(boolean, FragmentManager) static method.


HashEngineering:
@tomasz-ludek This has conflicts now.  Was this a PR to close without merging?

PR #212, opened 2019-06-13
DMD-1030 - Spork20 - LLMQ based InstantSend Spork

HashEngineering:
Spork 20, when activated will treat all transactions as instantSend Autolock

This is not the ideal way to do this.  Perhaps later, we can refactor the code after the spork is activated and we know that we won't go back to the 0.13 versions of instantSend.


HashEngineering:
This was merged manually up to this commit:  https://github.com/HashEngineering/dash-wallet/commit/4f342d5b197c02641f5f86af580da10f2244cdc4

PR #219, opened 2019-07-08
DMD-1103 Wallet Lock Timer | Restart

HashEngineering:
This code will lock the wallet by resetting the last unlock time to 0 when the device is rebooted or the app is upgraded.


HashEngineering:
This was mistakenly merged into the wrong branch.   It was manually cherry-picked into the master branch.

PR #221, opened 2019-07-09
DMD-797 Uphold logout

HashEngineering:
This handles the case of logging out.

It also handles the case of the user changing their password.

This also addresses some other items to make the screen look more like that of the iOS app.


common/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     api 'com.squareup.moshi:moshi:1.3.1'
     api 'com.squareup.retrofit2:retrofit:2.3.0'
     api 'com.squareup.retrofit2:converter-moshi:2.3.0'
+    api 'com.squareup.retrofit2:converter-scalars:2.3.0'
```
HashEngineering: We need this to handle the response from the revoke command, which returns text and not json.

uphold-integration/src/main/java/org/dash/wallet/integration/uphold/data/UpholdClient.java
```diff
@@ -76,13 +77,32 @@
...
+            okhttp3.Response response = chain.proceed(request);
+
+            long t2 = System.nanoTime();
+            log.info(String.format("Received response for %s in %.1fms%n%s\n%s",
+                    response.request().url(), (t2 - t1) / 1e6d, response.headers(), response.networkResponse()));
+
+            return response;
+        }
+    };
+
```
HashEngineering: This is here for debug purposes.

uphold-integration/src/main/java/org/dash/wallet/integration/uphold/data/UpholdClient.java
```diff
@@ -120,14 +141,37 @@ public void onResponse(Call<UpholdAccessToken> call, Response<UpholdAccessToken>
                     storeAccessToken();
                     getCards(callback, null);
                 } else {
-                    log.error("Error to obtain Uphold access token", response.message());
+                    log.error("Error to obtain Uphold access token " + response.message());
                     callback.onError(new Exception(response.message()), false);
```
HashEngineering: For the logs, we need to use %s in the strings or + to get the extra information included.  Many times the message is blank.

uphold-integration/src/main/java/org/dash/wallet/integration/uphold/data/UpholdClient.java
```diff
@@ -162,8 +206,8 @@ public void onResponse(Call<List<UpholdCard>> call, Response<List<UpholdCard>> r
                         }
                     }
                 } else {
-                    log.error("Error to obtain cards", response.message());
-                    callback.onError(new Exception(response.message()), false);
+                    log.error("Error obtaining cards " + response.message());
+                    callback.onError(new UpholdException("Error obtaining cards", response.message(), response.code()), false);
```
HashEngineering: The UpholdException class was created to send the response codes to the callback methods.

uphold-integration/src/main/java/org/dash/wallet/integration/uphold/data/UpholdClient.java
```diff
@@ -236,7 +280,16 @@ public void onSuccess(String data) {
...
             public void onError(Exception e, boolean otpRequired) {
-
+                log.error("Error loading Dash balance: " + e.getMessage());
+                if(e instanceof UpholdException) {
+                    UpholdException ue = (UpholdException)e;
+                    if(ue.getCode() >= 400) {
+                        //we don't have the correct access token, let's logout
+                        accessToken = null;
+                        storeAccessToken();
+                    }
```
HashEngineering: Error 401 is Unauthorized which is the result when the user changes the password in uphold and the app cannot link to the uphold account with its access token.

Here we are allowing any error greater than 400 to be used as a reason to reset the access token, which will require the user to relink their account.  We may want to limit this to 401 only.

HashEngineering: I also found that the getDashBalance is the first method called when the user accesses the Uphold screen after linking to their account.  If this fails with 401, then we can unlink it in the app, requiring the user to relink in the case of a changed password.

uphold-integration/src/main/java/org/dash/wallet/integration/uphold/ui/UpholdAccountActivity.java
```diff
@@ -178,4 +194,81 @@ public void onTransfer() {
...
+
+                @Override
+                public void onError(Exception e, boolean otpRequired) {
+                    showErrorAlert();
+                }
+            });
+        } else {
+            showAutoLogoutAlert();
+            openUpholdToLogout();
+        }
```
HashEngineering: These two lines will return the app to the UpholdSplash screen and show the uphold page so the user can log out manually.  We may need to put more instructions in the logout alert.

uphold-integration/src/main/res/layout/uphold_account_screen.xml
```diff
@@ -93,6 +95,16 @@
...
+        <Button
+            android:id="@+id/uphold_logout"
+            style="@style/MontserratButton"
+            android:layout_alignParentBottom="true"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:text="@string/uphold_logout"
+            android:maxLines="2"
+            android:layout_margin="16dp" />
+
```
HashEngineering: The logout button is on the bottom, everything else on top.

uphold-integration/src/main/res/values/strings-uphold.xml
```diff
@@ -23,5 +23,6 @@
     <string name="uphold_otp_dialog_title">Enter your 2FA code below:</string>
     <string name="uphold_transfer">Transfer</string>
     <string name="uphold_see_on_uphold">See on Uphold</string>
+    <string name="uphold_logout">Logout of Uphold</string>
```
HashEngineering: This text and other text that is not in the strings file will need to be approved for consistency.

HashEngineering:
Comments are placed in the code.

uphold-integration/src/main/java/org/dash/wallet/integration/uphold/data/UpholdClient.java
HashEngineering: unless that last parameter of the log methods is an Exception.

uphold-integration/src/main/res/layout/uphold_account_screen.xml
HashEngineering: The logout button was later moved to the top of the screen.

PR #224, opened 2019-07-17
DMD-1065  On boarding | New Install

tomasz-ludek:
This pull request contains the new Onboarding flow where user can choose whether during the very first launch the wallet should be:

1. Created from scratch (Create a New Wallet)
2. Restored from seed (Restore Wallet From Passphrase)
3. Restored from file (Restore Wallet From File)

In order to speed up the app launching process the initialization process from WalletApplication.onCreate() method was split into two basic parts.


wallet/res/drawable/splash_background.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android"
+    android:opacity="opaque">
+
+    <item
+        android:drawable="@drawable/splash_activity_background"
+        android:gravity="fill_vertical" />
+
+    <item
+        android:drawable="@drawable/splash_activity_slogan"
```
HashEngineering: This file appears to be missing.

HashEngineering:
Building in Android Studio requires that 
Settings / Experimental / Only sync the active variant is OFF.

https://stackoverflow.com/questions/54201216/android-studio-3-3-new-gradle-sync-is-not-supported-due-to-containing-kotlin-mo

wallet/res/drawable/splash_background.xml
HashEngineering: splash_activity_slogan

HashEngineering: https://github.com/HashEngineering/dash-wallet/pull/223/commits/ee30ba3f51a28471d113ea74d8b1b943e7953707

This commit from the #223 PR has the files

tomasz-ludek:
> Building in Android Studio requires that
> Settings / Experimental / Only sync the active variant is OFF.
> 
> https://stackoverflow.com/questions/54201216/android-studio-3-3-new-gradle-sync-is-not-supported-due-to-containing-kotlin-mo

In Android Studio 3.4.2 it works with no issues even if this option is enabled.

HashEngineering:
I am using Android Studio 3.4.1

tomasz-ludek:
Oh, interesting...

tomasz-ludek:
Regarding other issues:

> After installing and creating a new wallet, if the back button is pressed to exit the app, then the app returns to the onboarding screen and the user can create a new wallet. This activity should finish before going to the WalletActivity.

Fixed

> When installing over a previous version, the app goes to the onboarding screen instead of the WalletActivity.

Fixed

> When restoring from a backup file or recovery phrase from the onboarding screen, the app sometimes syncs the blockchain from the wrong point and does not pickup the transactions or coins.

Investigating

HashEngineering:
excellent, we are getting closer.

tomasz-ludek:
This changeset will be merged as a part of #223

PR #225, opened 2019-07-24
BlockchainServiceImpl:  use static setBootStrapStream method to avoid NPE

HashEngineering:


HashEngineering:
This was manually added to #226

PR #227, opened 2019-07-30
DMD-1079 Home Page | Unlocked

tomasz-ludek:
The new design of home screen:

image

The Quick Actions pane isn't a part of this story but is ready for review as well, the related ticket can be found here: https://dashpay.atlassian.net/browse/DMD-1081

The footer is out of the scope and will be implemented within https://dashpay.atlassian.net/browse/DMD-1080

@sambarboza ,
in this changeset, I added the `ui-redesign` module which is intended to contain all the new styles, fonts, colors, common custom controls, etc. it will simplify the reusability. If you agree I would like use this approach in all redesign-related work..


tomasz-ludek:
Replaced by #230

PR #239, opened 2019-09-18
Home screen sync progress

denlb:


wallet/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     }
     implementation 'com.github.tony19:logback-android-core:1.1.1-6'
     implementation 'com.google.code.findbugs:jsr305:3.0.2'
+    implementation 'org.greenrobot:eventbus:3.1.1'
```
HashEngineering: what does eventbus do?

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -361,6 +365,24 @@ public void run() {
                 broadcastBlockchainState();
             }
         };
+
+        @Override
+        protected void progress(double pct, int blocksLeft, Date date) {
+            super.progress(pct, blocksLeft, date);
+            final SyncProgressEvent event = new SyncProgressEvent(pct);
+            log.info(event.toString());
+            EventBus.getDefault().postSticky(new SyncProgressEvent(pct));
```
HashEngineering: `event` is created, but not used.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -958,6 +980,19 @@ private void broadcastBlockchainState() {
         }
     }
 
+    private int percentageSync() {
+        int chainHeadHeight = blockChain.getChainHead().getHeight();
+        int mostCommonChainHeight;
+        if (peerGroup.getMostCommonChainHeight() > 0) {
```
HashEngineering: This will generate a NullPointerException if peerGroup is null.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
denlb: Used on the next line 
```
log.info(event.toString());
```

wallet/build.gradle
denlb: It is used to send progress event to activities
http://greenrobot.org/eventbus/documentation/configuration/sticky-events/

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
denlb: How to make it null? I've tried to turn off my internet connection and clear blockchain data but it's still not null

HashEngineering: In my case, I found it to be **null** when the testnet app was taking a long time to connect and when I turned off the network/mobile/data on the emulator.

HashEngineering:
Instead of checking if `peerGroup` is null, which can also mean the BlockchainService hasn't finished starting up, we should use `BlockchainState.impediments` to determine if the network is not available or the device is offline.  A peerGroup will be created if the network is not available or if there are other reasons by no peers can be contacted.  Only after trying to connect and failing then the BlockchainState is updated with `impediments == Impediment.NETWORK`.

Additionally, when testing this code, if turning off the network connectivity on an emulator running Android 8.1, I do see that peerGroup is null, but the red box doesn't appear to indicate that the network is offline.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
HashEngineering: Oh right.

HashEngineering:
@denlb This now has merge conflicts due to the Pay/Receive PR being merged.  Can you fix them?

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -1069,6 +1097,56 @@ public void onNewKeyChainEncrypted() {
...
     }
 
+    @Override
+    public void onStart() {
+        super.onStart();
+        findViewById(R.id.restart_sync_icon).setOnClickListener(new View.OnClickListener() {
+            public void onClick(final View v) {
+                findViewById(R.id.sync_error_pane).setVisibility(View.GONE);
+                findViewById(R.id.sync_progress_pane).setVisibility(View.VISIBLE);
+            }
```
HashEngineering: The restart button appears to only change the status visually.  Does it try to reconnect?  

I am not sure how to get the block chain service to restart.

PR #240, opened 2019-09-19
Redesign: New Successful Transaction Screen

sambarboza:
# New Successful Transaction Screen


HashEngineering:
This question applies more to the story that covers Sending payments than this PR.

Do the transaction details popup occur before or after the app confirms that the transaction was sent?  "Confirmation" means that other peers have announced the TX or the TX is ISLocked.  Currently the app will close the SendCoinsActivity when such confirmation is received.

When I review this PR, I will test for the case that there is only 1 peer connected and 0 peers connected.  If there are no connections, then the app currently will allow the creation of a transaction, but it's status will be "The payment has not been broadcast".  With the redesign, we will have to see how it handles various error producing scenarios.

@tomasz-ludek @sambarboza

jhodges10:
Loving the upcoming re-design as well @sambarboza

HashEngineering:
I would recommend showing the transaction detail window here:

https://github.com/HashEngineering/dash-wallet/blob/redesign/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java#L356

It is at this point that the app knows that the transaction was sent.

sambarboza:
> I would recommend showing the transaction detail window here:
> 
> https://github.com/HashEngineering/dash-wallet/blob/redesign/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java#L356
> 
> It is at this point that the app knows that the transaction was sent.

Thanks for the comment, @HashEngineering 
It was addressed at https://github.com/HashEngineering/dash-wallet/pull/240/commits/67ebba6532ec44c96ce4728bdd788bec8ad73ed9.

HashEngineering:
This will have several merge conflicts and compiler errors to fix once #243 is merged.

HashEngineering:
This code is part of #243.  This PR will not be merged since the relevant code is already merged.

HashEngineering:
This was merged as part of #254.

PR #242, opened 2019-09-20
Redesign:  Upgrade to DashJ 0.17

HashEngineering:
This will be a work in progress.

**Known issues:**
1.  scrypt hashing may be slow -> PIN verification and spending could slow down considerably.  However, the commit for this change in bitcoinj said:  "Amazingly, on a Pixel 2 this implementation is a bit quicker than Will Glozer/Colin Percival's native code."

This PR will be rebased as more stories are added to the `redesign` branch

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 ext {
-    dashjVersion = "0.16.2"
+    dashjVersion = "0.17-SNAPSHOT"
 }
 
 buildscript {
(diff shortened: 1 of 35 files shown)


HashEngineering:
This PR won't be merged until after reviews are completed and also QA has successful results.

build.gradle
```diff
@@ -1,5 +1,5 @@
 ext {
-    dashjVersion = "0.16.2"
+    dashjVersion = "0.17-SNAPSHOT"
```
denlb: Why not just `0.17`?

HashEngineering: 0.17 is not completed.  It will be completed when the redesign is ready to be released.  Additionally, SNAPSHOT allows us to host these snapshot builds on a maven type repo that hosts snapshots.  This will allow the rest of the developers to not worry about building dashj locally on their machine in order to build the app when 0.17 is required.

PR #244, opened 2019-10-01
SetPinActivity:  Initialize seed after decrypting a backup file

HashEngineering:
This resolves the issue of seeing a blank recovery phrase after
restoring from a backup file after installing for the first time.


HashEngineering:
What was happening on during a restore from backup file operation during the onboarding process is that the recovery phrase wasn't saved to pass on to the verify recovery phrase activities.

This line was checking the state of encryption of the wallet.  In the case of a new wallet, it was not yet encrypted.  In the case of a restore operation from a backup file, the wallet was encrypted and therefore the `seed` was still null.

https://github.com/HashEngineering/dash-wallet/blob/ec8ee80fc4d013a24c26814a742657501d992074/wallet/src/de/schildbach/wallet/ui/SetPinActivity.kt#L97

PR #245, opened 2019-10-02
DMD-1203 Redesign:  Fix Syncing, update DNS Seeds and Testnet block explorers

HashEngineering:
Rebase this PR to keep the commits separated

DMD-1203

### DIFF
--- a/wallet/src/de/schildbach/wallet/Constants.java
+++ b/wallet/src/de/schildbach/wallet/Constants.java
@@ -66,7 +66,7 @@ public final class Constants {
         switch (BuildConfig.FLAVOR) {
             case "prod":
             case "beta": {
-                DNS_SEED = new String[]{"dnsseed.dash.org"};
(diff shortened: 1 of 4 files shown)


wallet/src/de/schildbach/wallet/Constants.java
```diff
@@ -66,7 +66,7 @@
         switch (BuildConfig.FLAVOR) {
             case "prod":
             case "beta": {
-                DNS_SEED = new String[]{"dnsseed.dash.org"};
+                DNS_SEED = new String[]{"dnsseed.dash.org", "dnsseed.dashdot.io"};
```
denlb: Maybe also masternode.io?
https://github.com/dashpay/dash/blob/master/src/chainparams.cpp#L298

HashEngineering: Good question.  masternode.io hasn't been working for the past year from what I could tell.

PR #249, opened 2019-10-15
Replace obsolete firebase-core with firebase-analytics and firebase-inappmessaging

HashEngineering:
firebase documentation:
https://firebase.google.com/docs/android/setup#available-libraries

This should resolve the GMS test failures regarding unofficial API usage.


HashEngineering:
This will not be merged, because a separate PR was created for the redesign branch.  #256

PR #251, opened 2019-10-17
Redesign:  Do not show transaction notifications when restoring from seed or file

HashEngineering:
Expected Behavior:
- Don’t show notifications if rescanning the blockchain (current behavior)
- Don’t show notifications if restoring from a recovery phrase or backup file (until 1 day ago)
- Do show notifications on new transactions received while the app is running (that are not in a block yet)
- Do show notifications on new transactions received since the app was last used.


wallet/src/de/schildbach/wallet/ui/preference/PreferenceActivity.java
```diff
@@ -68,7 +68,7 @@ private void initToolbar()
     @Override
     public void onBuildHeaders(final List<Header> target) {
         loadHeadersFromResource(R.xml.preference_headers, target);
-        hidePersonalSettingsIfLocked(target);
+        //hidePersonalSettingsIfLocked(target);
```
denlb: It should be commented?

HashEngineering: This should not be commented out in the PR.

PR #254, opened 2019-10-25
DMD-1318 Enter Pin

tomasz-ludek:


HashEngineering:
@tomasz-ludek This needs to be merged with `redesign` after we get the Transaction Details PR merged.

common/src/main/java/org/dash/wallet/common/util/MonetarySpannable.java
```diff
@@ -50,7 +50,7 @@ private static CharSequence format(final MonetaryFormat format, final boolean si
         if (monetary == null)
             return "";
 
-        checkArgument(monetary.signum() >= 0 || signed);
+        //checkArgument(monetary.signum() >= 0 || signed);
```
HashEngineering: Why is this commented out?

wallet/AndroidManifest.xml
```diff
@@ -233,6 +233,15 @@
...
 
+        <activity android:name="de.schildbach.wallet.ui.ReceiveActivity"
+            android:screenOrientation="portrait"
+            android:theme="@style/My.Theme.ChildActivity" />
+
+        <activity
+            android:name="de.schildbach.wallet.ui.TransactionResultActivity"
+            android:screenOrientation="portrait"
+            android:theme="@style/My.Theme.Fullscreen" />
+
```
HashEngineering: Seems like something isn't working correctly with GitHub's diff.  This code was added in the Pay/Receive PR and it is already in the `redesign` branch.

HashEngineering: Nevermind on this.

common/src/main/java/org/dash/wallet/common/util/MonetarySpannable.java
tomasz-ludek: Hmm... It was commented out by @sambarboza , I don't know why.

HashEngineering: The checkArgument is checking for positive numbers, but we are now using this method to format negative numbers.

PR #247, opened 2019-10-09
Updated ScanActivity to the latest version from bitcoin-wallet

tomasz-ludek:

PR #255, opened 2019-10-28
Redesign:  Update ScanActivity to the latest version from bitcoin-wallet

HashEngineering:
This is a redo of #247 because that was merged into the wrong branch.

There is something wrong with this PR because it shows 87 files changed, but should be less than 10.


HashEngineering:
This was done manually here:  https://github.com/HashEngineering/dash-wallet/commit/f1e561b38750378caf638c18c321b7664ff46f55

PR #257, opened 2019-11-06
Resolve issue of 100x dash -> fiat values

HashEngineering:
This resolves the issue of ever increasing fiat values, when tapping on dash and fiat fields on the send or request screens in locales that switch , and . compared to US.

### DIFF
--- a/wallet/src/de/schildbach/wallet/Constants.java
+++ b/wallet/src/de/schildbach/wallet/Constants.java
@@ -212,6 +212,8 @@ public final class Constants {
 
     public static final MonetaryFormat LOCAL_FORMAT = new MonetaryFormat()
             .noCode().withLocale(Locale.getDefault()).minDecimals(2).optionalDecimals().withGroupingSeparator();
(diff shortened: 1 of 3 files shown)


HashEngineering:
We may instead go with a complete revert of the PR that added these features for the 6.6.4 version, while another try to get this feature working correctly will be attempted for the redesign.

wallet/src/de/schildbach/wallet/Constants.java
```diff
@@ -212,6 +212,8 @@
 
     public static final MonetaryFormat LOCAL_FORMAT = new MonetaryFormat()
             .noCode().withLocale(Locale.getDefault()).minDecimals(2).optionalDecimals().withGroupingSeparator();
+    public static final MonetaryFormat LOCAL_FORMAT_NO_SEP = new MonetaryFormat()
```
denlb: What does `SEP` stand for? Not obvious

HashEngineering: NO_SEP = No separators.

HashEngineering:
We are not using this PR since the Fiat Separators PR was reverted completely in 6.6.4.

PR #259, opened 2019-11-07
Remove obsolete code from DashJ 0.17

HashEngineering:
Obsolete code pertaining to dashj-0.17 was removed concerning:
masternodes 0.12-0.13
InstantSend 0.12-0.13

This will not pass the Travis Tests because the dashj in question has not been pushed to the snapshot repo.

Code changes in dashj are here:
https://github.com/dashevo/dashj/pull/61


HashEngineering:
This passes because it is compatible with the current 0.17-snapshot.

PR #260, opened 2019-11-08
Reset wallet

denlb:
Safety -> Reset wallet

clears the wallet data and terminates the app


wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -543,6 +544,46 @@ public void onDismiss(DialogInterface dialog) {
...
+                    }
+                }
+            });
+            return;
+        }
+
+        log.info("Clearing all app data and exiting.");
+        try {
+            Runtime.getRuntime().exec("pm clear "+ getApplicationContext().getPackageName());
+        } catch (IOException e) {
```
HashEngineering: We already have a method `WalletApplication.eraseAndCreateNewWallet()` that handles removing the wallet.  This method does not remove any other files (such as the logs, which we want to keep in the event there is an issue with the wallet reset function).  Additionally, it will create a new empty wallet.

Another place to look that handles this is when the PIN is entered incorrectly to many times in is in `PinRetryController.showResetWalletDialog`

HashEngineering: In our first implementation of this reset functionality, all the logs were deleted as well as all other files.  Then when we got user reports of issues, there were no logs detailing what previously happened.

denlb: Tried `WalletApplication.eraseAndCreateNewWallet()` - it doesn't start the process from scratch (OnboardingActivity). I believe as a user i want all the data to be removed, like app was reinstalled, don't you think so?

denlb: For the user it feels like app has crashed when i do `WalletApplication.eraseAndCreateNewWallet()`

HashEngineering: With the redesign, the flow and other code will need to be updated or changed to make it go to the OnboardingActivity.  `WalletApplication.eraseAndCreateNewWallet()` only handles the wallet file and wallet object (deleting and creating a new file).  It doesn't do any UI.  The point of my first comment is to use `WalletApplication.eraseAndCreateNewWallet()` instead of 
`Runtime.getRuntime().exec("pm clear "+ getApplicationContext().getPackageName());` because the `pm clear` will delete other files that should remain.

HashEngineering:
@denlb Will this PR be replaced with #268?

denlb:
@HashEngineering i'll update this PR once 268 is merged. it's a separate task

denlb:
Will be implemented in #268

PR #264, opened 2019-11-13
Hide balance

denlb:
Hide balance from home screen


wallet/res/xml/preference_settings.xml
```diff
@@ -14,6 +14,11 @@
         android:summary="@string/preferences_own_name_summary"
         android:title="@string/preferences_own_name_title" />
 
+	<CheckBoxPreference
+		android:defaultValue="false"
+		android:key="hide_balance"
+		android:title="@string/preferences_hide_balance" />
+
```
HashEngineering: This is obsolete, but the current designs don't say where this setting should be in the UI.

denlb: @HashEngineering is the new design PR merged already?

HashEngineering: Yes, it is.  The Menu PR has been merged.

you can look at this to get an idea of how to add menu items:  #258

common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -45,6 +45,7 @@
 
     public static final String PREFS_KEY_BTC_PRECISION = "btc_precision";
     public static final String PREFS_KEY_OWN_NAME = "own_name";
+    public static final String PREFS_KEY_HIDE_BALANCE_ON_LAUNCH = "hide_balance_on_launch";
```
HashEngineering: from the story "If the setting for Hide Balance = ON then the balance should always be hidden when returning to the home screen (regardless of the last state)"

wallet/src/de/schildbach/wallet/ui/HeaderBalanceFragment.java
```diff
@@ -101,6 +106,7 @@ public void onAttach(final Activity activity) {
         this.activity = (AbstractBindServiceActivity) activity;
         this.application = (WalletApplication) activity.getApplication();
         this.config = application.getConfiguration();
+        hideBalance = config.getHideBalanceOnLaunch();
```
HashEngineering: Given this requirement:

If the setting for Hide Balance = ON then the balance should always be hidden when returning to the home screen (regardless of the last state)

hideBalance should be set also in `onResume()`

wallet/src/de/schildbach/wallet/ui/HeaderBalanceFragment.java
```diff
@@ -184,12 +192,28 @@ public void onEvent(SyncProgressEvent event) {
...
+        hideShowBalanceHint.setText(R.string.home_balance_hide_hint);
+        showBalanceButton.setVisibility(View.GONE);
+
         if (!isAdded()) {
             return;
         }
 
-        View balances = viewBalance.findViewById(R.id.balances);
-        TextView walletBalanceSyncMessage = viewBalance.findViewById(R.id.wallet_balance_sync_message);
         if (!isSynced) {
```
HashEngineering: If the app is not synced, should we also hide "tap to hide balance"?

wallet/src/de/schildbach/wallet/ui/SecurityActivity.kt
```diff
@@ -72,8 +83,46 @@ class SecurityActivity : BaseMenuActivity(), AbstractPINDialogFragment.WalletPro
...
+    fun resetWallet(view: View) {
+        val dialogBuilder = AlertDialog.Builder(this)
+        dialogBuilder.setTitle(R.string.wallet_lock_reset_wallet_title)
+        dialogBuilder.setMessage(R.string.wallet_lock_reset_wallet_message)
+        dialogBuilder.setNegativeButton(R.string.wallet_lock_reset_wallet_title) { _, _ -> executeResetWallet() }
+        dialogBuilder.setPositiveButton(android.R.string.no, null)
+        dialogBuilder.setCancelable(false)
+        val dialog: Dialog = dialogBuilder.create()
+        dialog.setCanceledOnTouchOutside(false)
+        dialog.show()
```
HashEngineering: This line of code is sufficient for resetting the wallet:
        
`PinRetryController.showResetWalletDialog(view.context, false)` -- This may have a few bugs but they are being addressed in #267 -- by using this line, we don't have to worry about how it works.

What this does is erases the wallet file then restarts the app, so the user will go through the onboarding process.

wallet/src/de/schildbach/wallet/ui/HeaderBalanceFragment.java
denlb: We shouldn't because user might want to hide balance before it shows up

wallet/src/de/schildbach/wallet/ui/HeaderBalanceFragment.java
denlb: It shouldn't because state is stored in hideBalance

denlb:
@HashEngineering 

> When the home screen is started for the first time or resumed, the balance should be hidden if the hide setting is ON.
fixed https://github.com/dashevo/dash-wallet/pull/264/files#diff-564e0f2b6314534c1d89b480007b4196R175
> Should we hide "tap to hide balance" when the app is not synced (the balance is not shown in that case anyways)
No, because in this case balance will be visible for a second before user will have a chance to hide it
> Use this PinRetryController.showResetWalletDialog(view.context, false) to reset the wallet.
done https://github.com/dashevo/dash-wallet/pull/264/files#diff-7de70db1c2c14b9b4f5b273bab274f82R83

HashEngineering:
Changes made look good.  This can be merged after #267.

PR #266, opened 2019-11-18
NMA-14 Home Page | Locked

tomasz-ludek:
NMA-14 Home Page | Locked
NMA-4 fingeprint doesn't work if remove the old one and add new fingerprint to device


wallet/build.gradle
```diff
@@ -100,7 +100,7 @@ android {
         disable 'MissingTranslation'
     }
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 19
```
tomasz-ludek: @HashEngineering what do you think about this change? I decided to switch from old API 16 to API 19 (which is still very old, though) when making system status and navigation bars translucent. In my opinion, we should systematically switch to newer API levels and redesign is a great opportunity for that.

HashEngineering: This will reduce the number of devices that can run the app.  We will need to see how many devices/users we would be excluding and run it by the Product Owner.  Bitcoin Wallet with their version 5.45 app had made their min SDK version = 19 (KitKat) and their version 7 (current) might be 21 or Android 5.0.  Users bring up issues concerning 4.4 and they are told to use a previous version of Bitcoin Wallet.  That works well enough given that Bitcoin doesn't have breaking protocol upgrades.  Previous versions of Bitcoin Wallet will work for the most part, while past versions of our app have lower likely hood of working.   

This particular change is probably OK, though, given that there probably are not many 4.1 to 4.3 devices around.  4.4 does have its own "cult" following though.

16 - Jelly Bean | 4.1
19 - Kit Kit | 4.4

HashEngineering: I was able to build this with 16.  I am not sure what features are made available by changing the min sdk level.  I thought that was only related to what devices that the app could run on.

HashEngineering: @tomasz-ludek I would like your comments on what I wrote above.  This might be a change that should be done separately with product owner's input.  This type of change could also be done repeatedly in the future and it would be a good idea to determine a process for this kind of change now.

wallet/src/de/schildbach/wallet/ui/ForgotPinActivity.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.android.synthetic.main.activity_forgot_pin.*
+
+class ForgotPinActivity : AppCompatActivity() {
+
+    companion object {
+
+        @JvmStatic
+        fun createIntent(context: Context): Intent {
+            return Intent(context, ForgotPinActivity::class.java)
+        }
```
HashEngineering: I don't see a place where this code is called to start this activity.

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -671,28 +674,25 @@ public void lockWalletIfNeeded() {
...
+        log.info("Removing all the data and restarting the app.");
 
-        log.info("creating new wallet after wallet wipe");
+        resetBlockchain();
+        wallet.shutdownAutosaveAndWait();
+        stopBlockchainService();
 
-        File walletBackupFile = getFileStreamPath(Constants.Files.WALLET_KEY_BACKUP_PROTOBUF);
-        if(walletBackupFile.exists())
-            walletBackupFile.delete();
```
HashEngineering: Why is the code to remove the backup file removed?

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -70,18 +35,54 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
+
```
HashEngineering: Why does android studio reorder the imports?

wallet/src/de/schildbach/wallet/WalletApplication.java
HashEngineering: Though the backup file would be recreated and therefore replace the old one during the onboarding process.  But we should check on that.

tomasz-ludek: I removed this code, since noticed it is duplicated by what is happening in `cleanupFiles()` method:
https://github.com/dashevo/dash-wallet/pull/266/files/1de855c9542b4460f5aa344238a3adc099ae9aa1#diff-cbdcd0a31a45f4288739e96a6627f2faR526

wallet/src/de/schildbach/wallet/ui/ForgotPinActivity.kt
tomasz-ludek: As we discussed in NMA-14 I left this activity for future use, it is not used anywhere currently.

wallet/src/de/schildbach/wallet/WalletApplication.java
tomasz-ludek: I used to use `Ctrl+Alt+L` shortcut very often when writing the code. This shortcut automatically reformats the code based on rules defined in Android Studio, it can be also configured to reorganize imports so that they were consistent in the whole project.

Unfortunately, the Bitcoin Wallet was using different formatting rules originally and when Android Studio reorganizes imports (by removing unused entries and so one) it also reorders entries.

Before the redesign, I was avoiding this to happen by formatting only the edited lines (Android Studio can be configured that way by `Ctrl+Alt+Shift+L` shortcut) but in the meantime, I understood it causes a mess in the files, but on the other hand, keeps the PRs easier to review (by the lack unneeded reformats)... 

However, since redesign requires so many changes and adds so many new files, it doesn't make sense to be consistent with the original Bitcoin Wallet formatting rules anymore and it is a good occasion to take the full advantage of auto-formatting ;)

There are still few files left that were not yet reformatted and `WalletApplication.java` was one of them. I hope you don't mind I didn't discuss this with you before...

wallet/build.gradle
tomasz-ludek: That is true, the code can be built with 16 and Gradle doesn't report any issues, but it doesn't mean the app will not crash on devices running Jelly Bean though...
I will test the app on Jelly Bean device and let's see what will happen.

tomasz-ludek: Crap, on Jelly Bean the app crashes when trying to initialize BLS:
`
Exception Ljava/lang/RuntimeException; thrown while initializing Lorg/dashj/bls/BLS;
11-21 11:59:06.851 2574-2574/hashengineering.darkcoin.wallet_test D/AndroidRuntime: Shutting down VM
`
It shows the onboarding screen but blow-up after clicking `Create new wallet`.

wallet/src/de/schildbach/wallet/WalletApplication.java
HashEngineering: That is fine.

wallet/build.gradle
HashEngineering: That is a known issue.  We have had that problem with the current app since 6.6.2.

Therefore we should have 19 as our minimum on the redesign.

tomasz-ludek:
> The ForgotPinActivity is not used from what I can tell.
> 
> The number of attempt remaining stops going down at 6 on the screen, though internally the app keeps counting down.

As we discussed in NMA-14 I left ForgotPinActivity for future use, it is not used anywhere currently.

The number of attempts remaining will be fixed within NMA-12.

wallet/build.gradle
HashEngineering: This is a result of something in the BLS library and threads.

HashEngineering:
This is approved on the basis that #267 will fix the remaining issues.

PR #267, opened 2019-11-21
NMA-12 PIN | Invalid Pin Entry

tomasz-ludek:
This feature was added in three places:
- Lock screen (when entering PIN to unlock the wallet)
[video](https://drive.google.com/file/d/14DfmvKCnhXgIHC8g-RTmD5HroQDJcOqW)
- Enter PIN dialog (for instance when authenticating during the quick pay flow ie. `Scan to pay` from Lock screen) [video](https://drive.google.com/file/d/1sH9Oy6wX5uCAWBuoy4nuhdjHZJ2LMyK6)
- Change PIN screen (when entering old PIN)


wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -132,6 +132,11 @@ public void onCreate() {
                 //noinspection ResultOfMethodCallIgnored
                 walletFile.delete();
                 config.clear(false);
```
HashEngineering: There is an issue with this `config.clear(false)` being here:

If a user does wipe the wallet, they will go to the onboarding process and create or recover a wallet.  When the app is restarted, this section of code will delete the wallet file.  The wallet file was recreated during the onboarding process, but ensureWipe is still true.  So here, the wallet file is deleted and ensureWipe = false.  

later in the app loading process, the wallet file will not be found and then the backup file is used to restore the wallet and this requires a blockchain reset.  
```
2019-12-03 13:48:28.485 14725-14725/hashengineering.dash.wallet.beta E/WalletApplication: [main] problem loading wallet
    java.io.FileNotFoundException: /data/user/0/hashengineering.dash.wallet.beta/files/wallet-protobuf (No such file or directory)
```

`config.clear(false);` should be placed at the end of the onboarding process, to clear this state.

 After the wipe() method is called, the wallet file is deleted.  Then the app is restarted.  When onCreate() is executed, the walletFile does not exist, so the `config.clear(false);` is not called to clear the state of ensureWipe.

Perhaps this block of code doesn't need to be here at all since the delete file is in wipe() and config.clear() can go to the end of onboarding.  The blockchain file is deleted in wipe().

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -676,18 +681,26 @@ public void lockWalletIfNeeded() {
...
-        PinRetryController.clearPrefs();
+        PinRetryController.getInstance().clearPinFailPrefs();
+
+        File walletBackupFile = getFileStreamPath(Constants.Files.WALLET_KEY_BACKUP_PROTOBUF);
+        if(walletBackupFile.exists())
+            walletBackupFile.delete();
+
+        File blockChainFile = new File(getDir("blockstore", Context.MODE_PRIVATE), Constants.Files.BLOCKCHAIN_FILENAME);
+        if (blockChainFile.exists())
+            blockChainFile.delete();
```
HashEngineering: This may be unnecessary since the wallet file is deleted.

Deleting the wallet file will trigger a blockchain reset.

HashEngineering: We don't need to worry about the DMN List, since that is handled by the BlockchainService

wallet/src/de/schildbach/wallet/ui/OnboardingActivity.kt
```diff
@@ -155,4 +157,19 @@ class OnboardingActivity : RestoreFromFileActivity() {
...
+        //Inverting dialog answers to prevent accidental wallet reset
+        dialogBuilder.setNegativeButton(R.string.wallet_lock_reset_wallet_title) { _, _ ->
+            WalletApplication.getInstance().wipe(context)
+        }
+        dialogBuilder.setPositiveButton(android.R.string.no, null)
+        dialogBuilder.setCancelable(false)
+        val dialog = dialogBuilder.create()
+        dialog.setCanceledOnTouchOutside(false)
+        dialog.show()
+    }
```
HashEngineering: We do need a showResetWalletDialog method of some sort for the Security / Reset Wallet menu item #264 (it will go in the hide balance PR @denlb) - `PinRetryController.showResetWalletDialog(this, false)`

Can you make this `static` or otherwise accessible to an outside class?  or suggest a different way to reuse this code.

tomasz-ludek: Fixed by converting ResetWallet dialog from OnboardingActivity into a separate DialogFragment, it can be now used everywhere simple by:
ResetWalletDialog.newInstance().show(getSupportFragmentManager(), "reset_wallet_dialog");

wallet/src/de/schildbach/wallet/WalletApplication.java
tomasz-ludek: Hmm... interesting observation. The idea was that the app is being immediately restarted after wipe (by calling `ProcessPhoenix.triggerRebirth(context);`) and mentioned code from the `WalletApplication.onCreate` method, that ensures removing wallet file, is called before the user is able to create/restore the wallet using onboarding flow.
I will double-check this.

wallet/src/de/schildbach/wallet/WalletApplication.java
tomasz-ludek: Ok, got it. I was wondering if this is needed or not.
Removing.

wallet/src/de/schildbach/wallet/WalletApplication.java
HashEngineering: I understand your reasoning, but the **wipe()** method has already deleted the wallet file.  I like the idea of keeping the delete wallet function in the **wipe()** method and not here.  It is a bit scary to see a **walletFile.delete()** near the code that is executed each time the app is opened up.

tomasz-ludek: Ok, I will re-validate this code.
It was added since in the past I had the issues with deleting the wallet file before restarting the wallet. The `walletFile.delete()` from `wipe()` returned `false` and file wasn't deleted. I thought it was caused by the use of this file in `BlockchainServiceImpl` but this set of instructions:
```
resetBlockchain();
wallet.shutdownAutosaveAndWait();
stopBlockchainService();
```
should avoid such a failure.
Anyways, I think the `wipe()` function became so susceptible that it should be developed and tested within a dedicated task...

HashEngineering: Ok, we will do a separate PR on wipe().

wallet/src/de/schildbach/wallet/WalletApplication.java
HashEngineering: Turns out this is needed.  We can fix this as part of the new wipe PR Story.

PR #271, opened 2019-11-26
Redesign - Import private key

sambarboza:


wallet/src/de/schildbach/wallet/ui/send/SweepWalletFragment.java
```diff
@@ -130,12 +127,8 @@
 	private TransactionsAdapter sweepTransactionAdapter;
```
HashEngineering: This "adapter" shows up on my device during the send progress.

HashEngineering:
This will be merged on the basis of having two approving reviews.

PR #273, opened 2019-12-04
PIN Optimizations [NMA-219]

HashEngineering:
**This PR will do two things:**
1.  Reduce the number of scrypt hashes on 32 bit devices.  On a 32 bit emulator this reduces PIN verification time by < 30%.  This change affects wallet files created by "Create Wallet" and "Restore from Recovery Phrase".  If the user restores from a backup file, the number of scrypt hashes remains the same. 
~~2.  Remove the native scrypt libraries because dashj uses bouncycastle (which includes a scrypt implementation that runs fairly fast on android devices).  The native script library has been unused since dashj was upgraded to 0.17.~~  This change was reverted.
3.  Revert the change in DashJ 0.17 to use the Bouncy Castle Scrypt hasher vs the native Scrypt hasher from com.lambdaworks.crypto.  On slower phones, the lambdaworks scrypt was faster.

**Background:**
When the user enters the PIN, that PIN is taken a salt value is appended.  That data is then hashed 64K times (or half that on low memory / 32 bit devives) and that is taken to derive a key to encrypt/decrypt the keys in the wallet file.

Generally, scrypt is about 1000 times slower than SHA256D (the common hash used by Dash for most purposes except for Blocks).

We should keep an eye on the speed of verifying the PIN as there are a few things that can be done if it is seen as been too slow for the user.

**Security Concerns:**
Reducing the number of scrypt hashes reduces the time it would take to brute force attack the wallet file.  The attacker must still get the wallet file either as a backup file or from a rooted phone.


HashEngineering:
I wonder if we should lower the scrypt hash number lower for slower devices.  On Krip phones, the PIN process was still taking several seconds.

HashEngineering:
This PR is based on this:
https://github.com/bitcoin-wallet/bitcoin-wallet/commit/3fb571292b552a07eae05f74053e3d28611518c4

HashEngineering:
More optimizations need to be done, including a few changes in DashJ before this PR is ready to merge.

PR #280, opened 2019-12-16
NMA-24 Transaction Details: Add Status

HashEngineering:
This will add the transaction status to the Transaction Details screen.  The status can be:
1.  Error Sending (red)
2.  Special Transaction Type (black)
3.  Processing, Confirming (orange) for Sent transactions (not updated if the status changes while the details screen is open)

The status at the bottom of the dialog does not show "Sent", "Received" or "Internal" since that information is at the top or middle of the dialog already.


HashEngineering:
This PR will conflict with #271.  This must be merged after that one.

PR #284, opened 2019-12-18
NMA-192  PIN | Lock Timer

tomasz-ludek:
Acceptance Criteria:

- OFF = The user is always returned to the lock screen if they minimize the app / switch apps / turn screen off-on and return to any screen in the app

- ON = The user is NOT returned to the lock screen if they minimize the app / switch apps / turn screen off-on and return to any screen in the app within a specified time limit

Also refactored the code by removing unused stuff related to WalletLock and CanAutoLockGuard.

(As you may notice, the base of this PR is `advanced_security` branch, thanks to this the AdvancedSecurityActivity tan be used to configure Auto Logout feature)

Basically all the changes related to the core of this story were implemented within this single commit: https://github.com/dashevo/dash-wallet/pull/284/commits/81579bbd9d32a5b748c5e314991f29c20ac0c243)


tomasz-ludek:
> This removes the PIN requirement for making a backup. What is the reasoning, @tomasz-ludek

@HashEngineering I added back this functionality and replaced the old UnlockWalletDialogFragment with the new CheckPinDialog when asking for pin.

PR #285, opened 2019-12-19
NMA-27: Redesign: rescan does nothing

HashEngineering:
NMA-27 nothing happens when click Rescan wallet

The rescan function appears to do nothing since the screen remained on the Settings screen instead of returning to the home screen.

This PR will address that by returning the app to the Home Screen.  Question:  Is there a better way to do this?

Additionally, the same thing will happen for the Import Private Key function on the tools screen.

Lastly, this will fix a crash on testnet when the blockchain is reset.


HashEngineering:
#290 is perferred over this PR.

PR #286, opened 2019-12-19
Fingerprint Toggle in Security screen

sambarboza:
#### Changes:
- Showing/Hiding the "Fingerprint Authentication" row based on device's availability of a Fingerprint sensor and compatible APIs.
- Setting state of the Fingerprint Authentication switch based on whether user enrolled with fingerprint.
- Enabling/Disabling Fingerprint Authentication according to user interaction with the switch.


wallet/res/values/strings.xml
```diff
@@ -372,7 +372,7 @@
     <string name="activity_security_backup_wallet">Backup Wallet to File</string>
     <string name="activity_security_view_recovery_phrase">View Recovery Phrase</string>
     <string name="activity_security_advanced">Advanced Security</string>
-    <string name="activity_security_biometric_auth">Biometric Authentication</string>
+    <string name="activity_security_biometric_auth">Fingerprint Authentication</string>
```
denlb: It can be also face auth, isn't it? So I guess biometric fits better here
The rest looks good

HashEngineering: This app currently only supports Fingerprint authentication.

wallet/src/de/schildbach/wallet/ui/UnlockWalletDialogFragment.java
```diff
@@ -32,20 +32,24 @@
 
     private static final String FRAGMENT_TAG = UnlockWalletDialogFragment.class.getName();
 
-    public static void show(final FragmentManager fm) {
-        new UnlockWalletDialogFragment().show(fm, FRAGMENT_TAG);
+    public static UnlockWalletDialogFragment show(final FragmentManager fm) {
+        UnlockWalletDialogFragment unlockWalletDialogFragment = new UnlockWalletDialogFragment();
+        unlockWalletDialogFragment.show(fm, FRAGMENT_TAG);
+        return unlockWalletDialogFragment;
```
HashEngineering: Based on the work done in the Pay/Receive PR's, the `CheckPinDialog `will be replacing `UnlockWalletDialogFragment`

Also, depending on how secure we want this setting to be, it may be not be necessary to prompt the user for their PIN.

tomasz-ludek: @sambarboza in the following commit you can find an example on how to easily do this: 
https://github.com/dashevo/dash-wallet/pull/284/commits/4fdaf6c74a4f4f09541859d6218fe7cae2212c1c#diff-7de70db1c2c14b9b4f5b273bab274f82R68

tomasz-ludek:
> On the Lock Screen, after the PIN is entered the app asks the user to set up fingerprint authentication, even if the Fingerprint Authentication setting is OFF.

I fixed this in #284

wallet/src/de/schildbach/wallet/ui/UnlockWalletDialogFragment.java
sambarboza: Done on https://github.com/dashevo/dash-wallet/pull/286/commits/e6c46b0be2f9fe3ed8a3f091a79391f181cdb6b5.
@HashEngineering , @tomasz-ludek

PR #287, opened 2019-12-26
NMA-24 (ii) Fix status disagreements between tx history and details 

HashEngineering:
This will display Sending instead of Processing on sent transactions.  This may disagree with iOS.


HashEngineering:
This will be merged on the basis that it passed QA and was a simple fix.

PR #289, opened 2019-12-30
Redesign recovery phrase error fix [NMA-284 #2]

HashEngineering:


HashEngineering:
This will be merged on the basis the the original PR #283 and that this is a simple fix

PR #222, opened 2019-07-10
Okhttp logging interceptor

tomasz-ludek:

PR #293, opened 2020-01-02
update okhttp

denlb:
In order to fix rates fallback strategy


tomasz-ludek:
> With this PR, then the app has references to 3 versions of okhttp.
> 
> One version is 2.7.5 which perhaps still results in the same issues of a NullPointerException when processing LocalBitcoins.com

Good point @HashEngineering ,
I think we should switch everything to the same version... Probably that is the reason why my old PR related to the same topic was much more complex (https://github.com/dashevo/dash-wallet/pull/222)

denlb:
Fixed the issue differently here https://github.com/dashevo/dash-wallet/pull/297

PR #294, opened 2020-01-03
Change spending PIN fixes

tomasz-ludek:
NMA-286 Wrong PIN! 7 attempts remaining” appearing on the Confirm PIN
NMA-288 locks the user out of the app when fails to do Confirm PIN
NMA-288 locks the user out of the app when fails to do Confirm PIN


tomasz-ludek:
Replaced by #295

PR #297, opened 2020-01-09
Fix Localbitcoins fallback strategy

denlb:
Missing _slash_ caused HTTP 301
Also USDC caption added because BitPay added USDC rate to the response
No need to update okhttp so far

### DIFF
--- a/wallet/src/de/schildbach/wallet/rates/ExchangeRate.java
+++ b/wallet/src/de/schildbach/wallet/rates/ExchangeRate.java
@@ -68,6 +68,9 @@ public class ExchangeRate {
         if (currencyCode.equalsIgnoreCase("VES")) {
             return "Venezuelan Bolívar";
         }
+        if (currencyCode.equalsIgnoreCase("USDC")) {
+            return "USD Coin (USDC)";
+        }
 
         if (currencyName == null) {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
--- a/wallet/src/de/schildbach/wallet/rates/ExchangeRate.java
+++ b/wallet/src/de/schildbach/wallet/rates/ExchangeRate.java
@@ -68,9 +68,6 @@ public class ExchangeRate {
         if (currencyCode.equalsIgnoreCase("VES")) {
             return "Venezuelan Bolívar";
         }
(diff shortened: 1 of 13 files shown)


denlb:
@HashEngineering  please give it a try with the last commit (updated okhttp3). I assume it might be the source of *different* problem which can be reproducible only on specific Android devices/os versions

HashEngineering:
@denlb 
I agree that this could be related to android version

I tried this updated PR on Android 8.1 and 9.0 and both result in no exchange rates and the same exception.  

However, if the call to LocalBitcoins is removed, then all rates are obtained and this ends up in the log:
`I/ExchangeRatesRepository: [pool-8-thread-1] exchange rates updated successfully with de.schildbach.wallet.rates.DashRatesSecondFallback@569f7b7`.  Of course, the VES price is incorrect.

I don't understand why its only the LocalBitcoins reponse that gives this error / exception, while all other calls work fine.

denlb:
@HashEngineering please try again with the last commit, I've fixed another serializer problem. Should work now:

image

HashEngineering:
This appears to be working.

The rates are retrieved.

This PR will also superceed #222.

wallet/src/de/schildbach/wallet/rates/ExchangeRate.java
```diff
@@ -68,6 +68,9 @@ public String getCurrencyName() {
         if (currencyCode.equalsIgnoreCase("VES")) {
             return "Venezuelan Bolívar";
         }
+        if (currencyCode.equalsIgnoreCase("USDC")) {
+            return "USD Coin (USDC)";
+        }
```
HashEngineering: This will be handled by #302 .  Can you remove this?  

Also it appears that there are many other currencies without names in the list.  

Other currencies from bitpay:
XRP, PAX, MRU, 

Some currencies appear to crash the app.  I will try to handle that in #302 also.

denlb: Done
Crash issue resolved in https://github.com/dashevo/dash-wallet/pull/298/commits/afc6bd3719960f8810489950a47e0c3b0026e5ec#diff-d59db2bdbe7841e4cb222c288bc4e156R76

PR #298, opened 2020-01-09
Redesign: Set Default Currency [NMA-243]

HashEngineering:
Several Issues are being address:
1. Special case of BYR (old currency for Belarus-BY that is still in the locale information of older devices before 2017) did not exist in the price sources and this resulted in a crash after sending a transaction.  This crash was a result of the currency not being found in the price source list, which could also happen if the price sources were not available for a new user who just installed the app.  Code was added to treat BYR as BYN to successfully get an exchange rate.
2. Add a list of currency names for non ISO 4217 currencies that are returned by the price sources
3. Add a list of currency codes that should be treated as other codes for the purposes of getting the name.  This applies to non ISO 4218 currency codes that will have the same name as the one that is in ISO 4217.  Example:  CNH vs CNY.
4. Lastly this will set the default currency based on the locale of the device if there is no Sim Card instead of using USD.  The risk with this change is that a locale may specify a currency that is not in the price sources.  In this case the user would have to select a different currency.  This code was added to resolve the first problem above, but it didn't.  This can be removed if we don't want this behavior.
5.  Logging is added to the detectUserCountry() code to help diagnose errors that may occur with users.

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/data/CurrencyInfo.java
+++ b/common/src/main/java/org/dash/wallet/common/data/CurrencyInfo.java
@@ -60,7 +60,8 @@ public class CurrencyInfo {
 
     static {
         obsoleteCurrencyMap = new HashMap<>();
-        obsoleteCurrencyMap.put("BYR", "BYN");
+        obsoleteCurrencyMap.put("BYR", "BYN"); // Belarus Ruble, changed in 2016
+        obsoleteCurrencyMap.put("MRO", "MRU"); // Mauritania Ouguiya, changed in 2018
 
         otherCurrencyMap = new HashMap<>();
         otherCurrencyMap.put("VES", "Venezuelan Bolívar");
@@ -68,7 +69,11 @@ public class CurrencyInfo {
         otherCurrencyMap.put("JEP", "Jersey Pound");
         otherCurrencyMap.put("IMP", "Isle of Man Pound");
         otherCurrencyMap.put("USDC", "USD Coin");
-
+        otherCurrencyMap.put("LTC", "Litecoin");
+        otherCurrencyMap.put("ETH", "Etherium");
+        otherCurrencyMap.put("BTC", "Bitcoin");
+        otherCurrencyMap.put("PAX", "Paxos Standard");
+        otherCurrencyMap.put("GUSD", "Gemini Dollar");
 
         useOtherNameMap = new HashMap<>();
         useOtherNameMap.put("CNH", "CNY");
--- a/common/src/main/java/org/dash/wallet/common/data/CurrencyInfo.java
+++ b/common/src/main/java/org/dash/wallet/common/data/CurrencyInfo.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2019 Dash Core Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/dash/wallet/common/data/CurrencyInfo.java
```diff
@@ -0,0 +1,125 @@
...
+        These currencies are listed in the price data, but are not
+        ISO 4217 currency codes.  This will map those codes to the
+        currency names.
+     */
+    private static HashMap<String, String> otherCurrencyMap;
+
+    /*
+        These currencies are listed in the price data and have the
+        same name as a different currency.  This differs from
+        obsoleteCurrencyMap because both of these codes exist in the
```
denlb: Should we put **USDC** here as well?

denlb: It's in the [bitpay response](https://bitpay.com/rates)

HashEngineering: Yes, good idea.

HashEngineering: this was added

HashEngineering: There are other currencies to add text for that come from BtcPay, so I will add those also.

common/src/main/java/org/dash/wallet/common/data/CurrencyInfo.java
```diff
@@ -60,15 +60,20 @@
...
         otherCurrencyMap.put("GGP", "Guernsey Pound");
         otherCurrencyMap.put("JEP", "Jersey Pound");
         otherCurrencyMap.put("IMP", "Isle of Man Pound");
         otherCurrencyMap.put("USDC", "USD Coin");
-
+        otherCurrencyMap.put("LTC", "Litecoin");
+        otherCurrencyMap.put("ETH", "Etherium");
+        otherCurrencyMap.put("BTC", "Bitcoin");
+        otherCurrencyMap.put("PAX", "Paxos Standard");
+        otherCurrencyMap.put("GUSD", "Gemini Dollar");
```
denlb: Will not be used I suppose? https://github.com/dashevo/dash-wallet/pull/298/commits/afc6bd3719960f8810489950a47e0c3b0026e5ec#diff-d59db2bdbe7841e4cb222c288bc4e156R76

HashEngineering: There is one currency, USDT that will be 4 letters long.  This check will prevent the exception that occurs for currency codes that are not 3 letters.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -1080,24 +1082,70 @@ private void detectUserCountry() {
...
+                    setDefaultCurrency();
             }
         } catch (Exception e) {
             //fail safe
+            log.info("NMA-243:  Exception thrown obtaining Locale information: ", e);
+            if (config.getExchangeCurrencyCode() == null)
+                setDefaultCurrency();
+        }
+    }
+
```
sambarboza: Please remove empty spaces.

HashEngineering: This will be done before merging.

HashEngineering: This has been fixed.

PR #300, opened 2020-01-10
Redesign: Fix View Recovery Phrase Flow and other security related flows [NMA-299]

HashEngineering:
The primary focus of this PR is to use the correct design for the flow of viewing the recovery phrase.  Previously the flow was the same as the verify recovery phrase flow.

ViewRecoveryPhrase

Additionally, the dialog was changed on the Secure Now function:

SecureNow

**Implementation:**
The CheckPinDialog was extended with a new subclass called DecryptSeedWithPinDialog while trying to maximize code reuse.  DecryptSeedWithPinDialog can use a PIN or a fingerprint to decrypt the seed that is given to the ViewSeedActivity.

I chose to derive a new class from CheckPinActivity instead of using CheckPinActivity to verify the PIN and then have a secondary set of task and screens with progress bars to perform the seed decryption with the key.

The old dialogs for decrypting the seeds are left in the app for the Wallet Upgrade processes (when a user upgrades the app from before v5.18.

This is a draft for now as there are a few questions on allowing fingerprint authentication on this feature and other security menu related items.

Additionally, this PR fixes a few things on the security menu:
1.  Back Up Wallet to File -> the first time this is run, it asks for permissions.  Then the user must click on the button again
2.  Advanced Security is put behind a CheckPinDialog

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
+++ b/wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
@@ -1,0 +1,88 @@
+package de.schildbach.wallet.ui
+
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProviders
+import de.schildbach.wallet.livedata.Status
+import de.schildbach.wallet.ui.preference.PinRetryController
+import org.bitcoinj.wallet.DeterministicSeed
+
+/**
+ * @author:  Eric Britten
+ *
+ * DecryptSeedWithPinDialog uses DecryptSeedSharedModel which is derived
+ * from CheckPinShared model but does not call the onCorrectPinCallback
+ * event
+ */
+
+class DecryptSeedWithPinDialog : CheckPinDialog() {
+
+    companion object {
+
+        private val FRAGMENT_TAG = DecryptSeedWithPinDialog::class.java.simpleName
+
+        private const val ARG_REQUEST_CODE = "arg_request_code"
+        private const val ARG_PIN_ONLY = "arg_pin_only"
+
+        @JvmStatic
+        fun show(activity: AppCompatActivity, requestCode: Int = 0, pinOnly: Boolean = false) {
+            val checkPinDialog = DecryptSeedWithPinDialog()
+            if (PinRetryController.getInstance().isLocked) {
+                checkPinDialog.showLockedAlert(activity)
+            } else {
+                val args = Bundle()
+                args.putInt(ARG_REQUEST_CODE, requestCode)
+                args.putBoolean(ARG_PIN_ONLY, pinOnly)
+                checkPinDialog.arguments = args
+                checkPinDialog.show(activity.supportFragmentManager, FRAGMENT_TAG)
+            }
+        }
+
+        @JvmStatic
+        fun show(activity: AppCompatActivity, requestCode: Int = 0) {
+            show(activity, requestCode, false)
+        }
+
+    }
+
+    override fun initViewModel() {
+        viewModel = ViewModelProviders.of(this).get(DecryptSeedViewModel::class.java)
+        (viewModel as DecryptSeedViewModel).decryptSeedLiveData.observe(viewLifecycleOwner, Observer {
+            when (it.status) {
+                Status.ERROR -> {
+                    pinRetryController.failedAttempt(it.data!!.second!!)
+                    if (pinRetryController.isLocked) {
+                        showLockedAlert(context!!)
+                        dismiss()
+                        return@Observer
+                    }
+                    setState(State.INVALID_PIN)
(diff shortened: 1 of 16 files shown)


wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
```diff
@@ -0,0 +1,114 @@
...
+                if (!canceled) {
+                    fingerprint_view.showError(exceededMaxAttempts)
+                }
+            }
+
+            override fun onHelp(helpCode: Int, helpString: String) {
+                fingerprint_view.showError(false)
+            }
+        })
+    }
```
tomasz-ludek: @HashEngineering ,
in order to condense the code even more and avoid duplicates, instead of overriding the whole `startFingerprintListener()` method, you can "override" just the `onSuccess(savedPass: String)` by extracting its content into a separate method.
To be more clear. In the `CheckPinDialog` add the method like `onSuccessFingerprint(savedPass: String)` and call it instead of `dismiss(savedPass)`. The base version of this method (in CheckPinDialog) will then call `dismiss(savedPass)` whereas the overridden version (in `DecryptSeedWithPinDialog`) `(viewModel as DecryptSeedViewModel).checkPin(savedPass)`

wallet/src/de/schildbach/wallet/ui/ViewSeedActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+            intent.extras!!.getStringArray(EXTRA_SEED)!!
+        } else {
+            throw IllegalStateException("This activity needs to receive a String[] Intent Extra " +
+                    "containing the recovery seed.")
+        }
+
+        val sb = StringBuilder(12)
+        seed.forEach {
+            sb.append("$it  ")
+        }
```
tomasz-ludek: This can be replaced by simple `seed.joinToString("  ")`

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -529,7 +531,31 @@ public void handleRestoreWallet() {
...
+                }
+            }
+        });
+        DecryptSeedWithPinDialog.show(this, AUTH_REQUEST_CODE_VIEW_RECOVERYPHRASE);
+    }
+
+    private void startVerifySeedActivity(DeterministicSeed seed) {
+        List<String> mnemonicCode = seed.getMnemonicCode();
+        String [] seedArray = mnemonicCode.toArray(new String[0]);
+        Intent intent = VerifySeedActivity.Companion.createIntent(this, seedArray);
```
tomasz-ludek: You can add the `@JvmStatic` annotation to `createIntent(context: Context, seed: Array<String>)` method definition in `VerifySeedActivity` class in order to be able to simply call `VerifySeedActivity.createIntent(this, seedArray)` without `Companion`.

wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
HashEngineering: Excellent idea.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
HashEngineering: Thank you for this idea.

wallet/src/de/schildbach/wallet/ui/ViewSeedActivity.kt
HashEngineering: Excellent.

wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
HashEngineering: I think we can do this in a few more places also.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
HashEngineering: https://github.com/dashevo/dash-wallet/pull/300/commits/82fc9d131051506372c3c9f69ac6bb890066e896

wallet/src/de/schildbach/wallet/ui/ViewSeedActivity.kt
HashEngineering: https://github.com/dashevo/dash-wallet/pull/300/commits/e556361cfcd7154f9350d4f80be30a668393726b

wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
HashEngineering: https://github.com/dashevo/dash-wallet/pull/300/commits/be532926bb667e8d4b425ba9f5b1375c7f808232

PR #301, opened 2020-01-11
Remove Insufficient Funds message when pay from lock screen

denlb:
Fixed behaviour when scanning to pay from the lock screen (i.e. before authentication) - the insufficient funds error is displayed right away.


denlb:
Implemented in #303

PR #303, opened 2020-01-15
Pay screen flow improvement

denlb:
# What's improved:
-  Pin entry before confirmation popover (if pin was not entered before or if _payment confirmation_  is ON).
- Returns back to login screen after successful payment if started from login screen in security reasons.
- Insufficient funds message will not be shown if PIN was not entered.


denlb:
@HashEngineering we cannot show _Confirmation_ before _Authentication_ - it includes the fee, but the fee cannot be calculated if there're some exceptions, for example insufficient funds (it's dashj limitation)

HashEngineering:
The implementation is fine.  We will need another PR later when it is decided what the flow should be.

I will look at the travis build failures before merging this.

HashEngineering:
#309 was merged which contains these changes.

PR #305, opened 2020-01-16
Avoid displaying DecryptSeedWithPinDialog on startup

tomasz-ludek:
Tiny fix, avoiding displaying DecryptSeedWithPinDialog when showing Home screen if the wallet wasn't secured previously.


HashEngineering:
This needs to be modified to show the dialog once after 24 hours from the onboarding process.

tomasz-ludek:
In order to avoid displaying the DecryptSeedWithPinDialog to show Backup Reminder (which should be done every 24h if the user didn't back up the seed) I replaced the use of `CheckPinViewModel.checkPinLiveData` with `DecryptSeedViewModel.decryptSeedLiveData` when checking the pin on Lock Screen.
@HashEngineering , I'm curious whether the `decryptSeedLiveData` will be significantly slower than `checkPinLiveData`? If so, I can try to call it only when Backup Reminder is supposed to be displayed.

HashEngineering:
@tomasz-ludek For speed, I don't know the answer yet.

The backup reminder should be displayed 1 time 24 hours after the onboarding process.  Showing it every 24 hours, which is what it might do now, is not the correct behavior.

tomasz-ludek:
You right @HashEngineering,
but I didn't change that part of the logic, just understood it incorrectly. It will be displayed only once.

tomasz-ludek:
It turned out some changes were needed against the original implementation. The dedicated JIRA ticket will be created to track this change.

HashEngineering:
This looks like it is working.

I will do a test to see what the speeds of the two methods are with CheckPin and DecryptSeed

HashEngineering:
Here is some interesting data on these two methods.  Three tests were run on each on an x86 emulator with Android 8.1.

CheckPin:  760 ms
DecryptSeed:  18 ms

DecryptSeed is significantly faster.  However, on this emulator, the app still has the status indicator bar moving for a few seconds after the logs report the time and before the Home Screen appears.

Is this because of running the app in debug mode?

Either way, you may have come across another optimization that will benefit all users that choose higher security, though this story was originally to fix a nagging issue of a backup reminder.

On an Android 9 device (Galaxy S9), I noticed a significant improvement in the processing time after entering the PIN.  There could be other optimizations that we could do as well, but let us see what happens with the Autologout = OFF story.

HashEngineering:
What CheckPin does is this:
1. derive key from PIN (includes scrypt hashing the PIN)
2. decrypt the first private key in the wallet
3. calculates the public key of that private key (this might be slow)
4. compares the public key to the first public key of the wallet

DecryptSeed does this:
1. derive key from PIN (includes scrypt hashing the PIN)
2. decrypt the seed of the wallet

PR #313, opened 2020-01-27
New security model based on Android Keystore

tomasz-ludek:
Introduced a new security model based on Android Keystore.
Separates the PIN from wallet file encryption password, which is now a complex random string generated by the app when creating the wallet.
More details can be found in the NMA-324 story description.

Includes required by NMA-324 Show No Lock Screen if Auto Logout = OFF.


wallet/src/de/schildbach/wallet/ui/VerifySeedActivity.kt
```diff
@@ -32,30 +35,59 @@ class VerifySeedActivity : AppCompatActivity(), VerifySeedActions {
...
+
+    private fun initViewModel() {
+        decryptSeedViewModel = ViewModelProviders.of(this).get(DecryptSeedViewModel::class.java)
+        decryptSeedViewModel.decryptSeedLiveData.observe(this, Observer {
+            when (it.status) {
+                Status.ERROR -> {
+                    finish()
+                }
+                Status.LOADING -> {
+
```
sambarboza: Do we need this empty block?

tomasz-ludek: Removed :)

wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.java
```diff
@@ -158,7 +161,7 @@ public void onClick(final DialogInterface dialog, final int which) {
     public void onAttach(final Context context) {
         super.onAttach(context);
 
-        this.activity = (AbstractBindServiceActivity) context;
+        this.activity = (SendCoinsActivity) context;
```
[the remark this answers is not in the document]
tomasz-ludek: Good point, however, in this case, I don't want to add additional complexity to already complex changeset. Let's leave it as it is for now.

PR #314, opened 2020-01-28
NMA-317 wallet crashes when try to specify too big amount to Receive

tomasz-ludek:
Fixed by limiting the max number of decimal digits to 6 (number of decimals was already limited).


tomasz-ludek:
Good find @HashEngineering , thanks!
I fixed it by catching the exception as you suggested. I analyzed other solutions, but this is the most robust.

PR #318, opened 2020-01-31
NMA-340 Disabled biometric limit when fingerprint disabled

dnllb:


common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -61,7 +61,7 @@
     public static final String PREFS_KEY_AUTO_LOGOUT_ENABLED = "auto_logout_enabled";
     public static final String PREFS_KEY_AUTO_LOGOUT_MINUTES = "auto_logout_minutes";
     private static final String PREFS_KEY_SPENDING_CONFIRMATION_ENABLED = "spending_confirmation_enabled";
-    private static final String PREFS_KEY_SPENDING_CONFIRMATION_LIMIT = "spending_confirmation_limit";
+    private static final String PREFS_KEY_BIOMETRIC_LIMIT = "biometric_limit";
```
denlb: This wording makes more sense

HashEngineering: this change is ok.

dnlilib:
@HashEngineering you're right. Preference does not reflect the UI toggle.

tomasz-ludek:
All issues fixed

PR #320, opened 2020-02-03
NMA-354 Resend tx if disconnected when sending

HashEngineering:
All of the relevant changes are in DashJ.

https://github.com/dashevo/dashj/pull/65


HashEngineering:
This fails the Travis test because dashj 0.16.5 is not published.

PR #322, opened 2020-02-05
Redesign - Transaction detail adjustments

sambarboza:
Transaction Detail screens adjustments to be in line with iOS app.


sambarboza:
@tomasz-ludek - conflicts resolved.

tomasz-ludek:
Oh snap, I opened it on a different device and it looks very ugly :|
This dialog should never take up the whole screen: 
Screenshot_20200210-172817

@HashEngineering this issue should be addressed for sure. @sambarboza maybe, we could make the top and bottom margins smaller?

sambarboza:
I agree, @tomasz-ludek / @HashEngineering. I'll fix this, thanks!

PR #326, opened 2020-02-10
Incorrect app name in the prod builds

tomasz-ludek:
Fixed the issue with the app name being "ui-components" instead of Dash Wallet in prod builds


common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -176,8 +176,7 @@ public boolean getTrustedPeerOnly() {
     }
 
     public Uri getBlockExplorer() {
-        return Uri.parse(prefs.getString(PREFS_KEY_BLOCK_EXPLORER,
-                res.getStringArray(R.array.preferences_block_explorer_values)[0]));
+        return Uri.parse(prefs.getString(PREFS_KEY_BLOCK_EXPLORER, null));
```
HashEngineering: was this an intended change for this fix?

HashEngineering: Perhaps because it was because this file was moved:  common/src/main/res/values/values.xml

But, then the getBlockExplorer will return null on new installs and that might make the "View in Explorer" link crash on the Tx Details screens.

HashEngineering:
There are a few spots to change:

```
/home/travis/build/dashevo/dash-wallet/wallet/src/de/schildbach/wallet/ui/BlockListFragment.java:181: error: method getBlockExplorer in class Configuration cannot be applied to given types;
                    startActivity(new Intent(Intent.ACTION_VIEW, Uri.withAppendedPath(config.getBlockExplorer(),
                                                                                            ^
  required: int
  found: no arguments
  reason: actual and formal argument lists differ in length
/home/travis/build/dashevo/dash-wallet/wallet/src/de/schildbach/wallet/ui/WalletAddressesFragment.java:182: error: method getBlockExplorer in class Configuration cannot be applied to given types;
                    startActivity(new Intent(Intent.ACTION_VIEW, Uri.withAppendedPath(config.getBlockExplorer(),
```

tomasz-ludek:
yeah, sorry I was in a hurry and missed it.
Fixed.

PR #340, opened 2020-02-20
NMA-376 App Shortcuts and Widget Crash Fixes

tomasz-ludek:
Fixed crash when using app appwidget before initializing the wallet.
Updating appwidget when receiving and sending funds.


HashEngineering:
Top is initialized, bottom is not regarding the wallet.

Screenshot_1582261083

PR #348, opened 2020-03-05
NMA-409 - Final beta updates - various UI fixes

HashEngineering:
[X] Trailing zeros should be removed from the Dash amount
* The default will be to show two zeros.

[X] The text on the login screen should be center aligned for all translated text

[X] Add line breaks in shortcut buttons if a translated word requires two lines
* use simple break strategy

[X] Add ellipses to the end of translated words in shortcut buttons which do not fit on one line
* this was already done on the Lock Screen

[X] Logout timer should be changed from "1 Day" to "24 hrs" in the advanced settings
* no translation update required, reusing "24 hrs" string

[X] Make the entire row in the payments screen actionable instead of just the button (keep the button for now)
* added a new rounded corner ripple effect that doesn't change the color of the unclicked state

[X] Change the title of the change pin header from "Change Spending Pin" --> "Change Pin"
* translation should not be required.  Already exists on iOS text.

[ ] Include updated translations (this will be a different PR)

**Other items**

1. Set Spending PIN -> Set PIN (Onboarding flow)
2. removed duplicate select_rounded_corner.xml file (used only in one place and didn't have a pre v21 equivalent)

This was mistakenly forked from #339, but that is possibly a good thing since now Transifex now has all the new strings for #339.  This will be rebased once #339 is merged.


HashEngineering:
Since this was forked from a test branch using for reviewing stories, it will be closed and replaced with #349

PR #350, opened 2020-03-06
NMA-364 Support Android 4.4

tomasz-ludek:
Made the app working on KitKat (API level 19)


HashEngineering:
@tomasz-ludek Currently DashJ uses `okhttp 3.12.3` what does that mean for our app if this PR is using `3.12.2`?  Do I need to lower DashJ to `3.12.2`?

tomasz-ludek:
@HashEngineering I'm not sure, but it turned out that 3.12.3 works on KitKat as well.
(I used 3.12.2 because I read on Stack Overflow that this is the last version supporting KitKat)

PR #344, opened 2020-03-02
Avoid app bar resizing when hiding/showing wallet balance

tomasz-ludek:
(this can be merged after https://github.com/dashevo/dash-wallet/pull/339)

### Before:
device-2020-03-02-140211
### After:
device-2020-03-02-140005

PR #354, opened 2020-03-10
Avoid app bar resizing when hiding/showing wallet balance [v2]

HashEngineering:
This will replace #344 since it was based on a PR that was updated and had many merge conflicts.


HashEngineering:
This was reviewed under #344.

PR #361, opened 2020-03-18
NMA-281 - Create identity processing

sambarboza:


wallet/res/layout/create_username.xml
```diff
@@ -63,92 +64,153 @@
...
+                android:id="@+id/processing_identity_dismiss_btn"
+                style="@style/DashButton.White.BlueText"
+                android:layout_width="match_parent"
+                android:layout_height="50dp"
+                android:layout_marginLeft="15dp"
+                android:layout_marginRight="15dp"
+                android:layout_marginBottom="25dp"
+                android:text="@string/processing_let_me_know"
+                android:textSize="18sp"
+                app:layout_constraintBottom_toBottomOf="parent" />
```
tomasz-ludek: @sambarboza ,
styling of this button isn't consistent with the styling of similar buttons on OnboardingActivity (create_new_wallet, recovery_wallet, restore_wallet). Last time I overlooked this, but it also applies to `Register` button. I guess this is caused by the inconsistency in designs but should be unified.
Here is how it looks like:
https://github.com/dashevo/dash-wallet/blob/08e361ef5ecf5da58485757dac79d6938b77ce3d/wallet/res/layout/activity_onboarding.xml#L47-L52

The biggest difference is the font size, but also height should be the same.
I applied the same style in `Confirm & Pay` button from NewAccountConfirmDialog:
https://github.com/dashevo/dash-wallet/blob/eec4f47959928bfa6a73e14710522e8c740149b2/wallet/res/layout/dialog_new_account_confirm.xml#L64-L71

PR #363, opened 2020-03-19
NMA-446 Crash after upgrade - IllegalStateException: Not encrypted

tomasz-ludek:
Fixed 'IllegalStateException: Not encrypted' crash:

```
2020-03-19 14:27:38.419 11201-11321/hashengineering.darkcoin.wallet_test E/AndroidRuntime: FATAL EXCEPTION: backgroundThread
Process: hashengineering.darkcoin.wallet_test, PID: 11201
java.lang.IllegalStateException: Not encrypted
at com.google.common.base.Preconditions.checkState(Preconditions.java:511)
at org.bitcoinj.wallet.KeyChainGroup.checkPassword(KeyChainGroup.java:468)
at org.bitcoinj.wallet.Wallet.checkPassword(Wallet.java:1399)
at de.schildbach.wallet.ui.CheckWalletPasswordTask$1.run(CheckWalletPasswordTask.java:40)
at android.os.Handler.handleCallback(Handler.java:883)
at android.os.Handler.dispatchMessage(Handler.java:100)
at android.os.Looper.loop(Looper.java:214)
at android.os.HandlerThread.run(HandlerThread.java:67)
```

If the old version (before 7.0.0) was installed and the PIN wasn't set by the user
the wallet file isn't encrypted, which causes crash when trying to check the PIN after upgrade.
This PR cause that the user is asked to set up a new PIN in such a case.

### DIFF
--- a/wallet/res/layout/fragment_enter_pin.xml
+++ b/wallet/res/layout/fragment_enter_pin.xml
@@ -29,6 +29,7 @@
                 android:padding="16dp">
 
                 <TextView
+                    android:id="@+id/title"
                     style="@style/MontserratMedium.PageTitle"
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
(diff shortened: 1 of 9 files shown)


tomasz-ludek:
> After installing older version, then restoring a phrase and not setting the PIN, then installing this PR, the PIN code doesn't work on `View Recovery Phase`

Good find!
Fixed.

PR #366, opened 2020-04-03
NMA-453 Fix for Notification / Network Monitor do not update during sync process

HashEngineering:
- Replace sync percentage calculation (use relative progress rather than absolute)

- broadcast `ACTION_BLOCKCHAIN_STATE `for Network Monitor (no other activity listens for this since https://github.com/dashevo/dash-wallet/commit/10f34a67a47de59c7e0d69a33fb8b2858aa3c07f, #339).  The network monitor needs to know when the blockchain tip has changed so it can display that last 50+ transactions.  It does not need the blockchain state itself (at least not until information about chainlocks needs to be displayed).

- Update blockchain state more regularly, possibly.  This may help the sync notification close more often.  The assumption is made that the `progress `and `doneDownload ` methods may not always trigger the blockchain state to be updated and the notification stays on for a longer period of time.  This is only a guess, but returns some behavior to the v6 app, which did work for having the notification close when the app was done syncing.

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
+++ b/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
@@ -170,6 +170,7 @@ public class BlockchainServiceImpl extends LifecycleService implements Blockchai
     public static final String START_AS_FOREGROUND_EXTRA = "start_as_foreground";
 
     private Executor executor = Executors.newSingleThreadExecutor();
+    private int syncPercentage = 0; // 0 to 100%
 
     private final ThrottlingWalletChangeListener walletEventListener = new ThrottlingWalletChangeListener(
             APPWIDGET_THROTTLE_MS) {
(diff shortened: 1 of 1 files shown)


wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -377,19 +378,32 @@ public void run() {
                     if(timeAgo < DateUtils.DAY_IN_MILLIS)
                         config.setRestoringBackup(false);
                 }
+                // this method is always called after progress or doneDownload
+                updateBlockchainState();
```
HashEngineering, line 382: In version 6, this was where the blockchain state was updated and broadcast.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -1026,6 +1040,11 @@ private void broadcastPeerState(final int numPeers) {
     }
 
     private void handleBlockchainStateNotification(BlockchainState blockchainState) {
+        // send this out for the Network Monitor, other activities observe the database
+        final Intent broadcast = new Intent(ACTION_BLOCKCHAIN_STATE);
+        broadcast.setPackage(getPackageName());
+        LocalBroadcastManager.getInstance(this).sendBroadcast(broadcast);
```
HashEngineering, line 1046: There may be a better way to do this in the BlockListFragment, with a database observer.  For now this is the quickest way to trigger the block loader in BlockListFragment with out more significant rewrites.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -1045,19 +1064,7 @@ private void handleBlockchainStateNotification(BlockchainState blockchainState)
...
-        }
-        if (peerGroup.getMostCommonChainHeight() > 0) {
-            mostCommonChainHeight = peerGroup.getMostCommonChainHeight();
-        } else {
-            mostCommonChainHeight = chainHeadHeight;
-        }
-        float percentage = ((float) chainHeadHeight / (float) mostCommonChainHeight) * 100;
-        log.info("mostCommonChainHeight: " + mostCommonChainHeight + "\tchainHeadHeight: " + chainHeadHeight + "\t" + percentage + "%\t" + config.getBestChainHeightEver());
-        return (int) percentage;
+        return syncPercentage;
```
HashEngineering, line 1067: This is a calculation that give the percentage in terms an absolute percentage of how much the entire blockchain is synced (from block 0).  This is some left over code that should have been caught in a previous review 4 months ago when the method was changed to match what `dashj `provides in the `progress `method above.

PR #369, opened 2020-04-08
NMA-307 Home Screen Status | Error State (UI)

tomasz-ludek:
Added error handling on Home screen for processing identity (used latest designs)


wallet/res/layout/dialog_new_account_confirm.xml
```diff
@@ -31,25 +31,86 @@
...
-            android:textColor="@color/dash_black"
-            app:layout_constraintDimensionRatio="3:1"
-            app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toTopOf="parent" />
-
-    </androidx.constraintlayout.widget.ConstraintLayout>
+        android:gravity="center"
+        android:padding="16dp"
+        android:text="You have chose “Johndoe” as your username. Username cannot be changed once its registered."
```
sambarboza, line 112: Should we move this to a resource string and format it with the chosen username?

tomasz-ludek: Oh, it is only for preview, I should probably use `tools:text` property instead of `android:text`.
I'll change this. The value of this field is being set with the real data over here: https://github.com/dashevo/dash-wallet/pull/369/files/eb371b8504c380c3e508fd210471ab1eef17a350

PR #370, opened 2020-04-08
NMA-448 Fix InstantSend/Chainlocks crashes by Updating to dashj 0.17.1

HashEngineering:
This resolves some crashes when processing instantsend and chainlocks.

Another fix to dashj was to remove the infinite loop when the send transaction fails (added for 0.14 nodes, which are not common anymore on the network).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 ext {
-    dashjVersion = "0.17"
+    dashjVersion = "0.17.1"
 }
 
 buildscript {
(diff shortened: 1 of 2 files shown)


HashEngineering:
https://github.com/dashevo/dashj/commits/release-0.17 - see the recent commits here for what else was changed in dashj 0.17.1

HashEngineering:
Tests will fail because dashj 0.17.1 is not published.

HashEngineering:
Tests are now passing.

PR #380, opened 2020-04-21
NMA-444: Connect UI to backend Part I: Check username existance

HashEngineering:
This PR uses a ViewModel, LiveData, Background Task methodology to determine if a username exists on a Dash Evonet and displays a message on the Create Username screen.  The idea behind this PR is to suggest a methodology to use to continue connecting the backend.  Two other connects that need to be made are this:

1. Determine if Platform is available
2. Register a username (4 step process)

To build this (Travis CI will fail on this PR), one must `git clone` these repo's and `gradlew assemble` each one in this order:

- https://github.com/dashevo/android-dpp (dpp:0.11-SNAPSHOT)

- https://github.com/dashevo/dapi-client-android (dapiclient:0.11-SNAPSHOT)

- https://github.com/HashEngineering/android-dashpay (platform-core & dashpay:0.11-SNAPSHOT).  This repo will be moved soon to the `dashevo` account.

**Known Issues:**
1.  InAppMessaging is disabled due to a crash that occurs when the above 3 libraries are included.  Most likely there is a conflict with dapi-client.
2.  As the user types a letter, the Platform will be queried.  But if another letter is typed while the last query is being executed, this should (but it may not be working well) cancel the last query and execute the new query.  There may be a better way to handle this.  Perhaps wait till the user stops typing and then make the query?
3. The first time the Platform is queries to check the existance of a username, it appears to be slower and the Register button can potentially still be pressed before the check for the username is started.

Questions:
1. Can a single live data class be used to make different types of queries (get username documents, get platform network status, etc).  In the future other Platform methods may be required.
2. Can the registration process be combined with the existing live data classes or should it be separated.
3. Coroutines are not currently used but there might be useful if because it seems that they have less overhead than creating a thread.  But perhaps there aren't enough threads created to justify coroutines.
4.  Is there a better way that using a ViewModel, LiveData, Background thread for these Platform Queries.  The Username creation process will be done one time, but other Platform queries and operations may be used much more frequently. 
5. The red checkmark should be a red x, possibly, when a username is found to exist.


HashEngineering:
@tomasz-ludek The last 2 commits add a function to check for the availability of Platform and the node meets some requirements (connections, protocol version) and then will show or hide the Join DashPay button on that basis.  

No other functions will be added to this PR.  Username Registration will be a separate PR.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -320,7 +321,7 @@ private void initViewModel() {
         // the Join DashPay (evolution) button on the shortcuts bar.
         // If that is the only function that the platform required for, then we can
         // conditionally execute this code when a username hasn't been registered.
-        dashPayViewModel = ViewModelProviders.of(this).get(DashPayViewModel.class);
+        dashPayViewModel = new ViewModelProvider(this).get(DashPayViewModel.class);
```
tomasz-ludek: This old way of retrieving `ViewModel`s is deprecated in 'androidx.core:core:1.2.0', I applied the recommended one.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -345,7 +345,7 @@ private void showHideJoinDashPayAction() {
             boolean canAffordIt = walletBalance.isGreaterThan(Constants.DASH_PAY_FEE)
                     || walletBalance.equals(Constants.DASH_PAY_FEE);
             boolean visible = canAffordIt && config.getShowJoinDashPay();
-            joinDashPayAction.setVisibility(View.VISIBLE);//visible ? View.VISIBLE : View.GONE);
+            joinDashPayAction.setVisibility(visible ? View.VISIBLE : View.GONE);
```
tomasz-ludek: Last time I forgot to revert this temporary debugging change.

wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
```diff
@@ -29,13 +29,10 @@ class DashPayViewModel(application: Application) : AndroidViewModel(application)
 
     private val platformRepo = PlatformRepo(application as WalletApplication)
 
-
     private val usernameLiveData = MutableLiveData<String>()
-    private val platformStatusLiveData = MutableLiveData<Boolean>()
 
     // Job instance (https://stackoverflow.com/questions/57723714/how-to-cancel-a-running-livedata-coroutine-block/57726583#57726583)
     private var getUsernameJob = Job()
-    private var isPlatformAvailableJob = Job()
```
tomasz-ludek: This field and all related code is only needed if you want to be able to reload the value of related `LiveData` multiple times and each time cancel previous request is it wasn't yet finished. I'll try to explain it in more details on Slack.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -330,7 +331,6 @@ public void onChanged(Resource<Boolean> status) {
                 showHideJoinDashPayAction();
             }
         });
-        dashPayViewModel.isPlatformAvailable();
```
HashEngineering: I am curious about this change.  How do we trigger this call?

tomasz-ludek: It is called automatically by the SDK when you starting observing on it:
`dashPayViewModel.isPlatformAvailableLiveData().observe(...)`

PR #391, opened 2020-04-26
Add Uphold Disabled Screen

HashEngineering:
This does #390 but changes the language of the new screen to Java.


HashEngineering:
#390 was fixed, so this is not required.

PR #400, opened 2020-05-06
Attempt to resolve RemoteServiceException's

HashEngineering:
Based on some research this PR makes the following changes:

1. Change `startService` to `ContextCompat.startForegroundService`
2. Notification channel id’s are not 0
3. call `startForeground` in `onCreate`

However, on an 8.1 emulator:

1. entered PIN on lock screen
2. there is an ANR
3. Click on More button
4. App Crash, log shows RemoteServiceException
5. What I did make the app crash more often.

### DIFF
--- a/wallet/src/de/schildbach/wallet/Constants.java
+++ b/wallet/src/de/schildbach/wallet/Constants.java
@@ -230,11 +230,11 @@ public final class Constants {
 
     public static final int MEMORY_CLASS_LOWEND = 64;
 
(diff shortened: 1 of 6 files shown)


tomasz-ludek:
Well done Eric, you broke the app! lol :)
Just kidding, it is grate we are finally able to reproduce the original crash.
I will investigate the issue with interest and let you know about the results.

HashEngineering:
This will be replaced with a much more simple #408 and this issue will be addressed later, if needed based on feedback.

PR #406, opened 2020-05-13
NMA-473 Payment Request from another app gets stuck on Transaction Result Screen

tomasz-ludek:
Added SendCoinsInternalActivity being not exported version of SendCoinsActivity extended with ability to be authorised internally.
Closing the app if it was launched by external Intent.

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/TransactionResultActivity.kt
+++ b/wallet/src/de/schildbach/wallet/ui/TransactionResultActivity.kt
@@ -46,19 +46,25 @@ class TransactionResultActivity : AbstractWalletActivity() {
         private const val EXTRA_PAYEE_VERIFIED_BY = "payee_verified_by"
 
         @JvmStatic
-        fun createIntent(context: Context, transaction: Transaction, userAuthorized: Boolean): Intent {
-            return createIntent(context, transaction, userAuthorized, null, null)
+        fun createIntent(context: Context, action: String? = null, transaction: Transaction, userAuthorized: Boolean): Intent {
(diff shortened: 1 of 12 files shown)


wallet/src/de/schildbach/wallet/ui/TransactionResultActivity.kt
```diff
@@ -77,10 +83,16 @@ class TransactionResultActivity : AbstractWalletActivity() {
...
+                when {
+                    intent.action == Intent.ACTION_VIEW -> {
+                        finish()
+                    }
+                    intent.getBooleanExtra(EXTRA_USER_AUTHORIZED_RESULT_EXTRA, false) -> {
+                        startActivity(WalletActivity.createIntent(this))
+                    }
+                    else -> {
+                        startActivity(LockScreenActivity.createIntentAsNewTask(this))
+                    }
```
HashEngineering, line 95: This part doesn't seem to work.

When using the secondary app that makes a payment request, the last screen of the payment process is the Transaction Result screen with a "Close" button.  Clicking the close button appears to execute the "else" clause here rather than the first clause that will **finish()** the activity.

tomasz-ludek: As we discussed on Slack the left button of the "Hack" app works in a hacky way so you should use the right button,

HashEngineering: Yes, this code is working under normal circumstances.

PR #408, opened 2020-05-13
Another attempt to resolve RemoteServiceException's

tomasz-ludek:
Moved calling BlockchainServiceImpl.startForeground(...) into main thread to avoid RemoteServiceExceptions

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
+++ b/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
@@ -41,6 +41,7 @@ import android.os.PowerManager.WakeLock;
 import android.text.format.DateUtils;
 
 import androidx.core.app.NotificationCompat;
+import androidx.lifecycle.LifecycleService;
 import androidx.lifecycle.Observer;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
@@ -58,8 +59,8 @@ import org.bitcoinj.core.Sha256Hash;
 import org.bitcoinj.core.StoredBlock;
 import org.bitcoinj.core.Transaction;
 import org.bitcoinj.core.TransactionConfidence.ConfidenceType;
-import org.bitcoinj.core.listeners.DownloadProgressTracker;
 import org.bitcoinj.core.Utils;
+import org.bitcoinj.core.listeners.DownloadProgressTracker;
 import org.bitcoinj.core.listeners.PeerConnectedEventListener;
 import org.bitcoinj.core.listeners.PeerDataEventListener;
 import org.bitcoinj.core.listeners.PeerDisconnectedEventListener;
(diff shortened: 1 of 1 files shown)


wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -1055,10 +1044,8 @@ private void handleBlockchainStateNotification(BlockchainState blockchainState)
                 nm.cancel(Constants.NOTIFICATION_ID_BLOCKCHAIN_SYNC);
             } else if (blockchainState.getReplaying() || syncing) {
                 //Shows ongoing notification when synchronizing the blockchain
-                Notification notification = createNetworkSyncNotification(blockchainState);
-                if (notification != null) {
-                    nm.notify(Constants.NOTIFICATION_ID_BLOCKCHAIN_SYNC, notification);
-                }
+                Notification notification = createNetworkSyncNotification();
```
HashEngineering, line 1047: We still need to pass the blockchainState here so the notification will be updated with the sync state.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -941,15 +933,12 @@ public void onTrimMemory(final int level) {
...
+    private Notification createNetworkSyncNotification() {
         Intent notificationIntent = OnboardingActivity.createIntent(this);
         PendingIntent pendingIntent = PendingIntent.getActivity(this, 0,
                 notificationIntent, PendingIntent.FLAG_UPDATE_CURRENT);
 
-        String message = BlockchainStateUtils.getSyncStateString(blockchainState, this);
-        if (message == null) {
-            message = getString(R.string.blockchain_state_progress_downloading);
-        }
+        String message = getString(R.string.blockchain_state_progress_downloading);
```
HashEngineering, line 941: I would propose this code, so that the notification can be updated with the current state.  But the first time it is called in startForeground, then `blockchainState = null`

```java
        String message = blockchainState != null ? BlockchainStateUtils.getSyncStateString(blockchainState, this) : null;
        if (message == null) {
            message = getString(R.string.blockchain_state_progress_downloading);
        }
```

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -941,15 +933,12 @@ public void onTrimMemory(final int level) {
         }
     }
 
-    private Notification createNetworkSyncNotification(BlockchainState blockchainState) {
+    private Notification createNetworkSyncNotification() {
```
HashEngineering, line 936: We should keep the blockchainState parameter for the reasons mentioned in the other comments.

tomasz-ludek:
Of course, you're right.
I don't know how could I missed this.
Fixed

PR #409, opened 2020-05-14
Post redesign cleanups

tomasz-ludek:
Removes unused code and resources


tomasz-ludek:
This will most likely be split to separate PRs

PR #412, opened 2020-05-19
NMA-508 Long Pin Issue

tomasz-ludek:
Restarting the app after update in order to force updating process (setup PIN length, etc.)
Explanation of the issue:
https://drive.google.com/file/d/1P67CIOIG9UcXxbzBmQSfM6OhuWYExpoO


wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -152,6 +153,7 @@ public void onCreate() {
         log.info("WalletApplication.onCreate()");
         config = new Configuration(PreferenceManager.getDefaultSharedPreferences(this), getResources());
         autoLogout = new AutoLogout(config);
+        myPackageReplaced = false;
```
sambarboza, line 156: If the var is already initialized at `false` on its declaration, why it needs to be set as `false` here as well?

tomasz-ludek: Removed

PR #416, opened 2020-05-21
NMA-430 | Layouts adjustments in payments and more screens

sambarboza:


HashEngineering:
The developer notes were added to Transifex, but they can also be added/modified in the source files (strings-extra.xml).

PR #417, opened 2020-05-26
NMA-452 Upgrade Crash (Firebase / Protobuf)

tomasz-ludek:
Switched to DashJ 0.17.2 which uses `protobuf-javalite` instead of `protobuf-java`

The Firebase SDK didn't yet switch to `protobuf-javalite` but there are some chances that this PR will fix the issue from NMA-452.

This seems to be the most promising solution https://github.com/gitprelimtek/protobuf-javalite-firebase-wellknowntypes however I can't get it working.


wallet/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation 'androidx.preference:preference:1.1.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.0.0-beta4'
     implementation "org.dashj:dashj-core:$dashjVersion"
-    implementation 'com.google.protobuf:protobuf-java:3.4.0'
+
```
HashEngineering, line 40: I would have thought we would have replaced this dependency with `protobuf-javalite`.  Where will our app get the code for this library?  Is it included in another dependency like dashj (bundled)?

tomasz-ludek:
Replaced by https://github.com/dashevo/dash-wallet/pull/485

PR #431, opened 2020-06-03
NMA-333 | User Search

sambarboza:

User Search feature

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone


wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -248,18 +259,15 @@ class PlatformRepo(val walletApplication: WalletApplication) {
             val profile = blockchainIdentity.watchProfile(10, 5000, BlockchainIdentity.RetryDelayType.SLOW20)
                     ?: throw TimeoutException("the profile was not found to be created in the allotted amount of time")
 
-            if (profile != null) {
```
HashEngineering, line 251: Good call on removing this null check.  This was copied from the recoveryIdentity method, but given that the watchProfile will return the profile document or throw an exception, we don't need to check for null.

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -92,19 +93,26 @@ class PlatformRepo(val walletApplication: WalletApplication) {
...
             // Names.search does support retrieving 100 names at a time if retrieveAll = false
-            var nameDocuments = platform.names.search(text, Names.DEFAULT_PARENT_DOMAIN, true)
+            val nameDocuments = platform.names.search(text, Names.DEFAULT_PARENT_DOMAIN, true)
+
+            val nameDocByUserId = nameDocuments.associateBy({ it.userId }, { it })
+            var userIds = arrayListOf<String>()
+            for (nameDoc in nameDocuments) {
+                userIds.add(nameDoc.userId)
+            }
+            val dashPayDocuments = Profiles(platform).getList(userIds)
```
HashEngineering, line 103: getList will only return the first 100 matches, but there might be a Exception from Platform if `userIds `contains more than 100 items.  We don't have any tests/examples in the DashPay library to know if this is the case.

https://dashplatform.readme.io/docs/reference-query-syntax (see "in")

HashEngineering: An exception will cause the search method to return an error without any results.  With only about 200 usernames that are registered, we wouldn't encounter this exception.  On Mainnet we would if more than 100 match the search text.

wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
```diff
@@ -68,7 +67,7 @@ class DashPayViewModel(application: Application) : AndroidViewModel(application)
         }
     }
 
-    fun searchUsernames(text: String, userId: String) {
+    fun searchUsernames(text: String, userId: String = "") {
```
HashEngineering, line 70: **userId** is the current **userId** of the user of the app.  This value is required to get all contacts, both the contacts that this user has created and that other users have created with this user.

wallet/src/de/schildbach/wallet/ui/SearchUserActivity.kt
```diff
@@ -0,0 +1,210 @@
...
+
+    private fun searchDashPayProfile() {
+        adapter.results = listOf()
+        hideEmptyResult()
+        if (this::searchDashPayProfileRunnable.isInitialized) {
+            handler.removeCallbacks(searchDashPayProfileRunnable)
+        }
+        if (query.length >= USERNAME_MIN_LENGTH) {
+            searchDashPayProfileRunnable = Runnable {
+                dashPayViewModel.searchUsernames(query)
```
HashEngineering, line 188: Since no value is specified for userId, this search will not retrieve any contact information for the current user with respect to the search results.

Of course, there won't any contact results to retrieve until this app can Request contacts.

HashEngineering:
This was replaced by #434

PR #442, opened 2020-06-17
NMA-332 View Pending Requests | Incoming (UI)

HashEngineering:
Modify **ContactsActivity** to show only pending contact requests when the user clicks "view all".

This should be merged after #440 

Screenshot_1592337626


HashEngineering:
This PR will be merged on the basis that it is included in the #444 PR.

PR #460, opened 2020-07-13
DashPay: Notifications fix -  NMA-551

HashEngineering:

The notification count on the home screen was not being updated all the time.

This was due to a race condition or reading the contact requests from the database before the database was completely updated from Platform.  In such a case the notification retrieval failed and no information about new notifications was displayed.

There are still some known issues with updating notifications:
1. After creating a user, the update contact timer does not appear to be activated.
2. After the blockchain service shuts down, the update contact timer is stopped.

This must be merged after #456 and its related PR's.

https://github.com/dashevo/android-dashpay/pull/16

## What was done?

Notification related LiveData uses a listener to update notifications and the notification count after the contacts requests are updated from platform.

These changes also made it possible to easily update the Notification Screen when new contact requests are detected.

## Checklist:

- [x] I have performed a self-review of my own code


wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsLiveData.kt
```diff
@@ -0,0 +1,55 @@
...
+        GlobalScope.launch {
+            val contactRequests = platformRepo.searchContacts(query, UsernameSortOrderBy.LAST_ACTIVITY)
+
+            //TODO: gather other notification types
+            // * invitations
+            // * payments
+            // * other
+
+            postValue(contactRequests)
+        }
```
HashEngineering, line 53: i am not sure if the GlobalScope.launch is the best way to run a suspend method.

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -62,6 +68,24 @@ class PlatformRepo(val walletApplication: WalletApplication) {
...
+        fun removeContactsUpdatedListener(listener: OnContactsUpdated?) {
+            onContactsUpdatedListeners.remove(listener)
+        }
+
+        fun queueContactsUpdatedListeners(updateBackgroundHandler: Handler) {
+            updateBackgroundHandler.post {
+                for (listener in onContactsUpdatedListeners)
+                    listener.onContactsUpdated()
+            }
+        }
```
HashEngineering, line 88: This section of code was made "static" because there could be more than one instance of PlatformRepo running in the app.  DashPayViewModel has one instance.  The BlockchainServiceImpl creates a new instance every time the timer triggers an update of the DashPay state (currently only updates the contact requests and profiles)

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -1072,20 +1072,18 @@ public void run() {
         });
     }
 
-    boolean updatingDashPayState = false;
     private void updateDashPayState() {
-        if (blockchainIdentityData != null && !updatingDashPayState) {
+        if (blockchainIdentityData != null) {
```
HashEngineering, line 1076: I would like to replace dependence on BlockchainServiceImpl for the timer that triggers DashPay related updates from Platform.  When the blockchain service is shutdown, the timer is also stopped and no further updates occur.  What is the best place to start a timer in the app that can be shut down when the app is closed?  This timer would only be active if the user has a username.

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
tomasz-ludek: @HashEngineering I think we should convert PlatformRepo to the Singleton.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
tomasz-ludek: As I mentioned above, I think we should convert the PlatformRepo to a singleton. After that the periodic updates could be triggered by PlatformRepo itself. Also, I recommend the use of Handler instead of Timer, like this:
```
Handler handler = new Handler();

private Runnable runnableCode = new Runnable() {
    @Override
    public void run() {
      // Do something here on the main thread
      Log.d("Handlers", "Called on main thread");
      // Repeat this the same runnable code block again another 2 seconds
      // 'this' is referencing the Runnable object
      handler.postDelayed(this, timeInterval);
    }
};

// Start the initial runnable task by posting through the handler
handler.post(runnableCode);

// Removes pending code execution
handler.removeCallbacks(runnableCode);
```
Here is more info about Handlers vs Timers on Android: https://stackoverflow.com/questions/20330355/timertask-or-handler#:~:text=3%20Answers&text=Handler%20is%20better%20than%20TimerTask,%2C%20here%2C%20and%20here).

wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsLiveData.kt
tomasz-ludek: It is okey :)

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
HashEngineering: Thanks.  I have implemented this except for the `handler.removeCallbacks` -- I wasn't sure where to put that so that it will be executed when the app is closed down.

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -638,9 +666,58 @@ class PlatformRepo(val walletApplication: WalletApplication) {
...
+    }
+
+    // Create the Handler object (on the main thread by default)
+    val timerHandler = Handler()
+    var timerStarted = false
+
+    // Define the code block to be executed
+    private val executeUpdateContacts = object : Runnable {
+        override fun run() {
+            // Do something here on the main thread
```
sambarboza, line 685: It looks like this and the comment below can be removed.

wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
```diff
@@ -109,16 +112,12 @@ class DashPayViewModel(application: Application) : AndroidViewModel(application)
...
+        notificationsLiveData.searchNotifications(text)
     }
 
-    val getNotificationCountLiveData = Transformations.switchMap(notificationCountLiveData) { date: Long ->
-        liveData(Dispatchers.IO) {
-            val count = platformRepo.getNotificationCount(date)
-            if (count >= 0)
-                emit(count)
-        }
+    fun getNotificationCount(date: Long) {
```
sambarboza, line 119: Are we going to use the `date` parameter in the future? If so, can we add a comment? Otherwise we should remove it.

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -638,9 +666,58 @@ class PlatformRepo(val walletApplication: WalletApplication) {
...
+
+    fun addContactsUpdatedListener(listener: OnContactsUpdated) {
+        onContactsUpdatedListeners.add(listener)
+    }
+
+    fun removeContactsUpdatedListener(listener: OnContactsUpdated?) {
+        onContactsUpdatedListeners.remove(listener)
+    }
+
+    fun queueContactsUpdatedListeners() {
```
sambarboza, line 717: It doesn't look like we're queuing calls on this method, so I would rename to `fireContactsUpdatedListeners` or something similar. Also, this function can be private.

HashEngineering: Fixed here:
https://github.com/dashevo/dash-wallet/pull/460/commits/929801ebf4d37deb29e65188d61588f5ddc7c7f1

wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
HashEngineering: date was once used, now the associated liveData retrieves this from the preferences directly

https://github.com/dashevo/dash-wallet/pull/460/commits/929801ebf4d37deb29e65188d61588f5ddc7c7f1

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
HashEngineering: https://github.com/dashevo/dash-wallet/pull/460/commits/929801ebf4d37deb29e65188d61588f5ddc7c7f1

PR #463, opened 2020-07-27
Fix: transaction dialog crash

sambarboza:
## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/WalletTransactionsFragment.java
+++ b/wallet/src/de/schildbach/wallet/ui/WalletTransactionsFragment.java
@@ -257,18 +257,6 @@ public class WalletTransactionsFragment extends Fragment implements LoaderManage
         return walletTransactionsFragmentView;
     }
 
-    @Override
-    public void onResume() {
-        super.onResume();
-        //reloadTransactions();
-    }
-
-    @Override
-    public void onPause() {
-
-        super.onPause();
-    }
-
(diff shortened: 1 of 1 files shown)


wallet/src/de/schildbach/wallet/ui/WalletTransactionsFragment.java
```diff
@@ -257,18 +257,6 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
         return walletTransactionsFragmentView;
     }
 
-    @Override
-    public void onResume() {
-        super.onResume();
-        //reloadTransactions();
-    }
```
HashEngineering, line 264: We don't need to reload the transactions onResume?  Not sure when this was commented out. or is the reload method called else where?

sambarboza: @HashEngineering , we don't need. This is called from the filter. If I recall correctly I added this while making the refactoring then I was it was not needed.

HashEngineering: ok, great

PR #475, opened 2020-08-05
DashPay: Send Payment | Frequent Contacts

HashEngineering:
## Issue being fixed or feature implemented

NMA-371
## Screenshots / Videos

Screenshot_1596668939

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [x] I have performed a self-review of my own code and added comments where necessary
- [x] I have added or updated relevant unit/integration/functional/e2e tests

PR #477, opened 2020-08-10
NMA-378 User Profile | Transaction History

tomasz-ludek:

NMA-378 User Profile | Transaction History

## Screenshots / Videos

device-2020-08-13-213447

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsForUserLiveData.kt
```diff
@@ -13,13 +12,17 @@ class NotificationsForUserLiveData(walletApplication: WalletApplication, platfor
...
+            if (contactRequests.data != null) {
+                contactRequests.data.filter { cr ->
+                    cr.dashPayProfile.userId == userId
                 }.forEach {
-                    results.add(NotificationItem(it))
+                    results.add(NotificationItemContact(it))
+                    if (it.type == UsernameSearchResult.Type.CONTACT_ESTABLISHED) {
+                        val invitationItem = if (it.incoming) it.copy(toContactRequest = null) else it.copy(fromContactRequest = null)
+                        results.add(NotificationItemContact(invitationItem, isInvitationOfEstablished = true))
+                    }
```
tomasz-ludek, line 25: Is it the right approach?
I generate "invitationItem" for each established contacts so that it could be displayed in the history.

HashEngineering: very cool.  Splitting into two items seems like a good way to do it.  But I will need to review this more to get a better idea of how it will work.

wallet/src/de/schildbach/wallet/data/UsernameSearchResult.kt
```diff
@@ -1,30 +1,45 @@
...
+    val type: Type
+        get() = when (requestSent to requestReceived) {
+            true to true -> Type.CONTACT_ESTABLISHED
+            false to true -> Type.REQUEST_RECEIVED
+            true to false -> Type.REQUEST_SENT
+            else -> throw IllegalStateException("toContactRequest and fromContactRequest can't both be null at the same time")
+        }
+
+    val incoming: Boolean
+        get() = toContactRequest!!.timestamp > fromContactRequest!!.timestamp
```
HashEngineering, line 38: Here, incoming appears to mean that we received a request first, then we sent our request later.

One issue here is that a UserSearchResult is also used for pending contact requests, in which cases `toContactRequest` will be null.  The NotificationActivity and ContactsFragment.  At this time, I don't think `incoming` will be used on either screen.

For the DashPayUserActivity screen, currently, or at least the way I wrote it, the history is only shown when there is a contact request going both ways (neither toContactRequest or fromContactRequest is null).

Also, I think the SearchUserActivity screen also uses UsernameSearchResult to hold a sent request where fromContactRequest is null.  While we don't use `incoming` in that activity, maybe it is better to make this getter null pointer exception proof.

tomasz-ludek: I replaced it with a local variable inside `NotificationsForUserLiveData.searchNotifications`

wallet/src/de/schildbach/wallet/ui/WalletFragment.kt
```diff
@@ -86,9 +91,23 @@ class WalletFragment : Fragment() {
...
+        GlobalScope.launch {
+            val contactRequests = PlatformRepo1.getInstance().searchContacts("", UsernameSortOrderBy.DATE_ADDED, true)
+            if(contactRequests.data != null) {
+                contactRequests.data.forEach {
+                    println("contactRequests:\t$it")
+                }
+            }
+        }
+    }
+
```
HashEngineering, line 110: are these changes for debugging?

wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsAdapter.kt
```diff
@@ -115,321 +82,64 @@ class NotificationsAdapter(val context: Context, val wallet: Wallet, val onConta
...
-            NOTIFICATION_CONTACT_ADDED -> holder.bind(results[position].notificationItem!!.usernameSearchResult!!, results[position].isNew)
-            NOTIFICATION_NEW_HEADER -> (holder as HeaderViewHolder).bind(R.string.notifications_new)
-            NOTIFICATION_EARLIER_HEADER -> (holder as HeaderViewHolder).bind(R.string.notifications_earlier)
-            NOTIFICATION_NEW_EMPTY -> (holder as ImageViewHolder).bind(R.drawable.ic_notification_new_empty, R.string.notifications_none_new)
-            NOTIFICATION_PAYMENT -> (holder as TransactionViewHolder).bind(results[position].notificationItem!!.tx!!)
-            else -> throw IllegalArgumentException("Invalid viewType ${results[position].viewType}")
+        return when (getItemViewType(position)) {
+            NOTIFICATION_HEADER -> 1L
+            NOTIFICATION_EMPTY -> 2L
+            else -> getLongValue(getItem(position).notificationItem.getId())
```
HashEngineering, line 88: This is the only place that has a conflict with my PR #475, which is pretty impressive.  there are some imports, so the merge process will very easy.

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -299,9 +299,7 @@ class PlatformRepo private constructor(val walletApplication: WalletApplication)
                 val usernameSearchResult = UsernameSearchResult(profile.value!!.username,
                         profile.value!!, toContact, fromContact)
 
-                // only include contacts that have sent requests to us (we may have accepted them)
-                // do not include contactRequest that we have sent but have not been accepted
-                if (usernameSearchResult.requestReceived)
+                if (usernameSearchResult.requestReceived || (includeSentPending && usernameSearchResult.requestSent))
                     usernameSearchResults.add(usernameSearchResult)
```
HashEngineering, line 303: very good!

wallet/src/de/schildbach/wallet/ui/dashpay/notification/ContactViewHolder.kt
```diff
@@ -0,0 +1,128 @@
...
+                }
+                UsernameSearchResult.Type.REQUEST_SENT -> {
+                    contact_added.setImageResource(R.drawable.ic_add_contact)
+                    R.string.notifications_you_sent
+                }
+            }
+
+            @Suppress("DEPRECATION")
+            val displayNameText = context.getString(displayNameResId, "<b>$name</b>")
+            displayName.text = Html.fromHtml(displayNameText)
```
HashEngineering, line 88: Very good! this gets the formatting to match the designs. I had no idea how make the display name bold.

wallet/src/de/schildbach/wallet/ui/WalletFragment.kt
tomasz-ludek: Yeap, I removed them. Good find! :)

wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsAdapter.kt
tomasz-ludek: New polymorphic data model allowed me to simplify this method even more :)

Also, I used `UUID.randomUUID()` for NOTIFICATION_HEADER `ID` instead of hardcoded number because in `NotificationActivity` we have two items of this type (New/Earlier) whereas `ID` should be unique as we as we use the `setHasStableIds(true)`. Previously, when there was a separate type NOTIFICATION_EARLIER_HEADER the uniqueness was assured by hardcoded values. One of advantage of the new approach is that we can have as much header items as we want without the need of creating new item type for each.

wallet/src/de/schildbach/wallet/ui/WalletFragment.kt
tomasz-ludek: Removed

wallet/src/de/schildbach/wallet/ui/dashpay/notification/ContactViewHolder.kt
tomasz-ludek: I had few ideas, but this one seems to be the best for different languages.

PR #502, opened 2020-09-02
BIP70 | Broadcast Transaction After Server Ack

tomasz-ludek:

NMA-593 BIP70 | Broadcast Transaction After Server
## How Has This Been Tested?

Performed few BIP70 payments and analysed operations order in logs.
Performed few regular transactions in order to make sure that nothing was accidentally broken.

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/send/PaymentProtocolViewModel.kt
+++ b/wallet/src/de/schildbach/wallet/ui/send/PaymentProtocolViewModel.kt
@@ -51,7 +51,7 @@ class PaymentProtocolViewModel(application: Application) : SendCoinsBaseViewMode
     val sendRequestLiveData = MutableLiveData<Resource<SendRequest>>()
 
     val exchangeRateData: LiveData<ExchangeRate>
-    val directPaymentAckLiveData = MutableLiveData<Resource<Pair<Transaction, Boolean>>>()
+    val directPaymentAckLiveData = MutableLiveData<Resource<Pair<SendRequest, Boolean>>>()
 
     val exchangeRate: org.bitcoinj.utils.ExchangeRate?
         get() = exchangeRateData.value?.run {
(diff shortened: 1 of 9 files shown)


wallet/src/de/schildbach/wallet/ui/send/PaymentProtocolViewModel.kt
```diff
@@ -127,14 +126,22 @@ class PaymentProtocolViewModel(application: Application) : SendCoinsBaseViewMode
         signAndSendPayment(finalPaymentIntent!!, baseSendRequest!!.ensureMinRequiredFee)
     }
 
-    fun directPay(transaction: Transaction) {
+    override fun signAndSendPayment(sendRequest: SendRequest, txAlreadyCompleted: Boolean) {
```
HashEngineering, line 129: I am not sure what `txAlreadyCompleted` means.  On line 126 above, `txAlreadyCompleted` is set to `ensureMinRequiredFee` which I thought was usually set to true.

wallet/src/de/schildbach/wallet/ui/send/SendCoinsOfflineTask.java
```diff
@@ -51,14 +51,23 @@ public SendCoinsOfflineTask(final Wallet wallet, final Handler backgroundHandler
...
                 org.bitcoinj.core.Context.propagate(Constants.CONTEXT);
 
                 try {
                     log.info("sending: {}", sendRequest);
-                    final Transaction transaction = wallet.sendCoinsOffline(sendRequest); // can take long
+                    if (txAlreadyCompleted) {
+                        wallet.commitTx(sendRequest.tx);
+                    } else {
+                        wallet.sendCoinsOffline(sendRequest);
+                    }
```
HashEngineering, line 69: Where does the value of txAlreadyComplete come from? The logic here seems to be that if the tx is completed (`wallet.completeTx` was called) then it only needs to be committed, otherwise `wallet.sendCoinsOffline` will complete and commit the transaction.  I suppose this makes sense, so this relates to a comment above about how the value of txAlreadyComplete is set.

wallet/src/de/schildbach/wallet/ui/send/PaymentProtocolViewModel.kt
tomasz-ludek: It is pretty tricky ;)
Before sending the transaction to server it has to be completed. Previously it was completed by the `sendCoinsOffline` method since we were broadcasting the transaction before sending to server. Now as we first want to send it to the server it has to be completed explicitly (and it is in https://github.com/dashevo/dash-wallet/pull/502/files/af444fa05fec1e0f6e56b242a7f5394f0707b94a#diff-6c4b3d3eb5b78806c89f5e5ffda4e1a1R131). On the other hand it cannot be completed again when we will be broadcasting the transaction. That is why I added this flag, but you probably already know that.
The tricky part is related to inheritance of payment related ViewModels. In the mentioned line 126, other version of this method is being called, it gets PaymentIntent as a first parameter, not the SendRequest.
The only place where the `txAlreadyCompleted` flag is true is here:
```
fun commitAndBroadcast(sendRequest: SendRequest) {
        super.signAndSendPayment(sendRequest, true)
}
```
https://github.com/dashevo/dash-wallet/pull/502/files/af444fa05fec1e0f6e56b242a7f5394f0707b94a#diff-6c4b3d3eb5b78806c89f5e5ffda4e1a1R168

PR #512, opened 2020-09-23
NMA-431 | Settings | View DashPay Profile

sambarboza:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering:
I would suggest that we add view functionality to this PR.  First we will need to create a default and empty profile for the local database (not on Platform):

During username creation, CreateIdentityService.kt:314 in `createIdentity`:
```kotlin
        // Step 6: A profile will not be created on Platform, since the user has not yet specified
        //         a display name, public message (bio) or an avatarUrl
        //         However, a default empty profile will be saved to the local database.
        val emptyProfile = DashPayProfile(blockchainIdentity.uniqueIdString, blockchainIdentity.currentUsername!!)
        platformRepo.updateDashPayProfile(emptyProfile)
```

And also during the recovery of a phrase, PlatformRepo.Kt:686 in `recoverDashPayProfile`.  Instead of returning if no profile is found on platform, create an empty one:

```kotlin
// recovery will only get the information and place it in the database
                val profile = blockchainIdentity.getProfile() ?:
                    profiles.createProfileDocument("", "",
                        "", blockchainIdentity.identity!!)
```

Lastly, `PlatformRepo.updateDashPayProfile` will need to be made public.

Then on the `MoreActivity `and the `EditProfileActivity` an observer such as this `AppDatabase.getAppDatabase().dashPayProfileDao().loadDistinct(blockchainIdentity!!.uniqueIdString).observe(...)` can be used to obtain the avatar and displayname and bio for the current username.  As in other screens, when accessing a profile and the avatarUrl is blank, we use the UserAvatarPlaceholderDrawable which is already in this PR.  If required, an example is here:  https://github.com/dashevo/dash-wallet/blob/evonet-develop/wallet/src/de/schildbach/wallet/ui/DashPayUserActivity.kt#L106

HashEngineering (changes requested):
See the comment above and also adjust the character limits for the display name and bio to match what is in the Story.

HashEngineering (changes requested):
This was tested on a few devices, by clicking and scanning QR codes.  Overall it works well:

**Here are some bugs:**
1. When sending a contact request to a user, once the operation is completed "Activity" appears but there is no activity listed.  The only item would be that the contact request was sent on a specific date/time.  The activity_rv has the "Activity" row item, but not a Notification item.  This problem also existed before you work on this story, though the "Activity" Row was also missing.
2. The Accept button on a contact or notification row on other screens (Contacts, Notifications, Add a Contact) does nothing when pressed.
3. One comment was added to the code to align method names between the `DashPayProfileDao*` classes
4. Merge with `evonet-develop` and fix one compile issue related to `loadDistinct` and `loadByUserNameDistinct` - this is from merging #512.

HashEngineering (approved):
All comments have been addressed.

PR #515, opened 2020-09-30
NMA-617 Notifications are not updated automatically(v8.0.3)

tomasz-ludek:
Fixed the issue with incorrect value of LastSeenNotificationTime causing the notifications count to not be displayed.
(there were microseconds saved instead of milliseconds)

NMA-617
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/Configuration.java
+++ b/common/src/main/java/org/dash/wallet/common/Configuration.java
@@ -437,10 +437,17 @@ public class Configuration {
     }
 
     public long getLastSeenNotificationTime() {
-        return prefs.getLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, 0);
+        long savedTime = prefs.getLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, 0);
+        long distantFutureTime = 2524608000L; //2050-01-01
+        if (savedTime > distantFutureTime) {
+            // date incorrectly saved in microseconds, fix it
+            savedTime = savedTime / 1000;
+            setPrefsLastSeenNotificationTime(savedTime);
+        }
+        return savedTime;
     }
 
     public void setPrefsLastSeenNotificationTime(long lastSeenNotificationTime) {
-        prefs.edit().putLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, lastSeenNotificationTime).commit();
+        prefs.edit().putLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, lastSeenNotificationTime).apply();
     }
 }
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsActivity.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsActivity.kt
@@ -145,7 +145,7 @@ class NotificationsActivity : InteractionAwareActivity(), TextWatcher,
                             (notificationsAdapter.getItem(currentPosition).notificationItem as NotificationItemContact).usernameSearchResult.toContactRequest = it.data!!
                             notificationsAdapter.notifyItemChanged(currentPosition)
                             currentPosition = -1
-                            lastSeenNotificationTime = it.data.timestamp.toLong()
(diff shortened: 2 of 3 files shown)


common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -437,10 +437,17 @@ public void setShowJoinDashPay(boolean showJoinDashPay) {
...
 
     public long getLastSeenNotificationTime() {
-        return prefs.getLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, 0);
+        long savedTime = prefs.getLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, 0);
+        long distantFutureTime = 2524608000L; //2050-01-01
+        if (savedTime > distantFutureTime) {
+            // date incorrectly saved in microseconds, fix it
+            savedTime = savedTime / 1000;
+            setPrefsLastSeenNotificationTime(savedTime);
+        }
```
tomasz-ludek, line 446: This code can be removed as soon as we make sure all existing users (i.e. alpha-testers) updated to this version.

HashEngineering: We won't need this because after our recent updates in our code to v0.15 Platform, users in the alpha program can't update without starting from scratch.  The next version that they get will be on fresh new network.

wallet/src/de/schildbach/wallet/ui/dashpay/NotificationsActivity.kt
```diff
@@ -145,7 +145,7 @@ class NotificationsActivity : InteractionAwareActivity(), TextWatcher,
                             (notificationsAdapter.getItem(currentPosition).notificationItem as NotificationItemContact).usernameSearchResult.toContactRequest = it.data!!
                             notificationsAdapter.notifyItemChanged(currentPosition)
                             currentPosition = -1
-                            lastSeenNotificationTime = it.data.timestamp.toLong()
+                            lastSeenNotificationTime = it.data.timestamp / 1000
```
HashEngineering, line 148: we could store lastSeenNotificationTime in milleseconds to stay consistent with all other Platform timestamps.

HashEngineering: https://github.com/dashevo/dash-wallet/blob/evonet-develop/wallet/src/de/schildbach/wallet/data/NotificationItemContact.kt#L15

This line which takes a millisecond time and multiplies by 1000 is the problem, probably.  If this is not done, then we don't need to divide by 1000.

common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -437,10 +437,17 @@ public void setShowJoinDashPay(boolean showJoinDashPay) {
...
+        return savedTime;
     }
 
     public void setPrefsLastSeenNotificationTime(long lastSeenNotificationTime) {
-        prefs.edit().putLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, lastSeenNotificationTime).commit();
+        prefs.edit().putLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, lastSeenNotificationTime).apply();
```
HashEngineering, line 451: this was changed from .apply() to .commit() to ensure that the setting would actually be saved right away.

Otherwise a user can leave the Notifications screen which has New items and then return immediately and the New items are still new, rather than Earlier.

Is there a better way?  Does/can this cause an ANR?

tomasz-ludek: I'm not sure if there is a need for protect against mentioned issue.
The documentation of `apply()` method reads:
```As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value.```
And we actually ignore the return value.

tomasz-ludek:
Replaced by https://github.com/dashevo/dash-wallet/pull/517

PR #513, opened 2020-09-24
Restore Identity from First Public Key

HashEngineering:
## Issue being fixed or feature implemented

NMA-590

Due to the changes made, the identity, username and contacts will be loaded at 30%, instead of the normal 90-100 percent which results in missing transactions and requires a blockchain rescan.

This resolves some causes of missing payments from contacts after restoring a wallet from a recovery phrase.
## Related PR's and Dependencies

https://github.com/dashevo/android-dashpay/pull/24

## Screenshots / Videos

Screenshot_1600883380-smaller
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #517, opened 2020-10-01
Notifications are not updated automatically(v8.0.3)

tomasz-ludek:

NMA-617
Fixed the issue with incorrect value of NotificationItemContact.getDate() causing the notifications count to not be updated.
(this is a leftover after data formats change in platform)

https://github.com/dashevo/dash-wallet/pull/513
(merge directly to `chain-sync-phase-3a` or change base to `evonet-develop`)
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -441,6 +441,6 @@ public long getLastSeenNotificationTime() {
     }
 
     public void setPrefsLastSeenNotificationTime(long lastSeenNotificationTime) {
-        prefs.edit().putLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, lastSeenNotificationTime).commit();
+        prefs.edit().putLong(PREFS_LAST_SEEN_NOTIFICATION_TIME, lastSeenNotificationTime).apply();
```
tomasz-ludek, line 444: Based on the documentation of `apply()` method and lack of issues I suggest to use it over `commit()`

> You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states.

source: https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()

HashEngineering (approved):
These changes work.  Even the change from commit() to apply()

PR #516, opened 2020-09-30
Master into evonet-develop merge

tomasz-ludek:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #519, opened 2020-10-01
NMA-615 | Contacts Page | Unregistered Users

sambarboza:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [X] I have performed a self-review of my own code and added comments where necessary


wallet/src/de/schildbach/wallet/ui/MainActivity.kt
```diff
@@ -115,6 +134,32 @@ class MainActivity : AbstractBindServiceActivity(), ActivityCompat.OnRequestPerm
...
+            if (status?.status === Status.SUCCESS) {
+                if (status.data != null) {
+                    isPlatformAvailable = status.data
+                }
+            } else {
+                isPlatformAvailable = false
+            }
+            walletFragment.setPlatformAvailability(isPlatformAvailable)
+        })
+        AppDatabase.getAppDatabase().blockchainIdentityDataDao().loadBase().observe(this,
```
HashEngineering, line 153: `blockchainIdentityDataDao` was renamed `blockchainIdentityDataDaoAsync`.  This PR must be based on an older commit from the `evonet-develop` branch.

HashEngineering: `blockchainIdentityDataDao()` is now only to be used in `suspend` methods, while `blockchainIdentityDataDaoAsync()` is actually asynchronous.

wallet/src/de/schildbach/wallet/util/WalletExtensions.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import de.schildbach.wallet.Constants
+import org.bitcoinj.core.Coin
+import org.bitcoinj.wallet.Wallet
+
+fun Wallet.canAffordIdentityCreation(): Boolean {
+    val walletBalance: Coin = getBalance(Wallet.BalanceType.ESTIMATED)
+    return (walletBalance.isGreaterThan(Constants.DASH_PAY_FEE)
+            || walletBalance == Constants.DASH_PAY_FEE)
+}
```
HashEngineering, line 11: My question is not related to this PR, but what happens when the user only has 0.01 DASH and they try to register.  Will it succeed or fail?  The reason for the question is that there is a fee that is required to send the transaction.  Though, in the case of sending all the value, then the fee may be taken out of the 0.01.  What if the balance is 0.01 - something less than the fee?  We can have QA figure this out and make a new ticket if there is a problem.

HashEngineering (commented):
I left a few comments.  It seems that this PR is not based on the most recent commit from the `evonet-develop` branch, though, GitHub does not point this out.

This PR will have conflicts with #516 (Merging the Master Branch into evonet-develop to add many of the fixes for the past few releases of Dash Wallet).  We will need to determine the best merge order.

wallet/src/de/schildbach/wallet/ui/MainActivity.kt
sambarboza: Fixed on https://github.com/sambarboza/dash-wallet/commit/964b74e840cd75422b5beaaee8e6c853f207c2c8

sambarboza:
> I left a few comments. It seems that this PR is not based on the most recent commit from the `evonet-develop` branch, though, GitHub does not point this out.
> 
> This PR will have conflicts with #516 (Merging the Master Branch into evonet-develop to add many of the fixes for the past few releases of Dash Wallet). We will need to determine the best merge order.
> 
> One issue found is this:
> 
> 1. From Home Screen with no username, click on the contacts Screen.
> 2. On Contacts Screen, click upgrade.
> 3. Cancel / go back to the home screen, the Contacts button in the footer is still highlighted
> 4. Clicking on the Contacts button on the footer doesn't return to the Contacts Screen

Good find, @HashEngineering! Issue addressed.

tomasz-ludek:
I'm all for refactoring WalletFragment and MainActivity but let's do it right!
I've limited time today so decided to commit proposed improvements instead of just describing them in comments.
Please have a look at https://github.com/dashevo/dash-wallet/pull/519/commits/ef6db8aabc49cef4ebfca621df5ca163ce24e5da
First of all callback based communication between fragment and activity is so old school... Let's switch to shared view model: https://developer.android.com/topic/libraries/architecture/viewmodel#sharing
Also, in DashPayViewModel we should keep only common methods/LiveDatas used by different Activities/Fragments whereas less common stuff should be kept in dedicated ViewModels. That is why `isPlatformAvailable` has been moved to `MainActivityViewModel`.

@sambarboza I think `UpgradeToEvolutionFragment` should listen to `WalletCoinsReceivedEvent` and `WalletCoinsSentEvent` in order to enable `Upgrade` button when `wallet.canAffordIdentityCreation()` (similarly as in `WalletFragment`), however instead of registering `OnCoinsSentReceivedListener` and `CoinsSentEventListener` in both fragments you could create a dedicated `LiveData` in `MainActivityViewModel` based on awesome `ContactsBasedLiveData` designed by @HashEngineering (instead of `platformRepo.addContactsUpdatedListener` you will have `wallet.addCoinsReceivedEventListener`, `wallet.addCoinsSentEventListener`).

Hopefully, all the changes I made are fully correct since I'm in hurry...

wallet/src/de/schildbach/wallet/ui/MainActivity.kt
```diff
@@ -102,6 +115,21 @@ class MainActivity : AbstractBindServiceActivity(), ActivityCompat.OnRequestPerm
...
     }
 
+    fun initViewModel() {
+        viewModel = ViewModelProvider(this)[MainActivityViewModel::class.java]
+        viewModel.blockchainStateData.observe(this, Observer {
+            // just to trigger data loading
+        })
+        viewModel.isPlatformAvailableData.observe(this, Observer {
+            // just to trigger data loading
+        })
```
tomasz-ludek, line 125: Those empty observers could be replaced by other way of triggering data loading but let's leave it as it is
I'll fix it when working on "JoinDash" shortcut when https://github.com/dashevo/dash-wallet/pull/516 in merged (it will be merged after this PR)

wallet/src/de/schildbach/wallet/ui/WalletFragment.kt
```diff
@@ -119,27 +132,7 @@ class WalletFragment : Fragment() {
         showHideJoinDashPayAction()
     }
 
-    fun setBlockchainState(blockchainState: BlockchainState?) {
-        this.blockchainState = blockchainState
-        if (isDetached || !isVisible) {
```
tomasz-ludek: this shouldn't be needed when using LiveDatas

wallet/src/de/schildbach/wallet/util/WalletExtensions.kt
tomasz-ludek: @HashEngineering when I sent 0.01 to the new wallet and try to create an Identity it fails due to insufficient funds exception...

tomasz-ludek:
I couldn't resist myself from implementing further improvements in MainActivityViewModel.
Please have a look https://github.com/dashevo/dash-wallet/pull/519/commits/3cc66d32afe546523d074e6215bb63b195b9ce90
The most important part is `MainActivityViewModel.isAbleToCreateIdentityData` which combines all the needed data all together.
`WalletBalanceBasedLiveData` based on mentioned before Eric's `ContactsBasedLiveData`.

HashEngineering (approved):
Looks good with the new LiveData's.  Works well.

PR #520, opened 2020-10-02
DashPay: Add create/update profile support

HashEngineering:

NMA-570, 538

https://github.com/dashevo/android-dashpay/pull/25
https://github.com/dashevo/android-dpp/pull/10
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/EditProfileActivity.kt
```diff
@@ -116,8 +161,25 @@ class EditProfileActivity : BaseMenuActivity() {
...
+            dashPayProfile = it
+            showProfileInfo(it)
+        }
+    }
+
+    private fun setEditingState(isEditing: Boolean) {
+        if (!isEditing) {
+            editProfileViewModel.dashPayProfileData.observe(this, dashPayProfileObserver)
+        } else {
+            editProfileViewModel.dashPayProfileData.removeObserver(dashPayProfileObserver)
```
HashEngineering, line 181: The purpose of this method is to support the updating the profile if it was changed on another device.

However, one the user starts typing, then the observer is removed.  

Originally, I had the observer turned back on after the update operation was successful and finished, but then the observer is called with the previous values in the profile and then the screen would change all the text fields back to the previous values.  Though, if one goes back to the More screen, then he newly changed Display Name would show up.

For now, after editing, then the View is no longer live.  This may actually be a bad idea to have the view being live.  It is perhaps unlikely that a user would update their profile on two devices at the same time.

wallet/src/de/schildbach/wallet/ui/MoreFragment.kt
```diff
@@ -66,14 +78,54 @@ class MoreFragment : Fragment(R.layout.activity_more) {
...
+                            Toast.makeText(requireActivity(), "Processing update", Toast.LENGTH_LONG).show()
+                            update_profile_status_container.visibility = View.VISIBLE
+                            editProfile.visibility = View.GONE
+                        }
+                        Status.CANCELED -> {
+                            update_profile_status_container.visibility = View.VISIBLE
+                            update_status_text.text = "Cancelled" //hard coded text
+                            editProfile.visibility = View.VISIBLE
+                        }
+                    }
```
HashEngineering, line 124: This observer seems to be called after a update operation is finished in the EditProfileActivity and the user returns to the More Screen.

wallet/src/de/schildbach/wallet/ui/EditProfileActivity.kt
```diff
@@ -104,6 +113,41 @@ class EditProfileActivity : BaseMenuActivity() {
             }
 
         })
+
+        editProfileViewModel.updateProfileRequestState.observe(this, Observer {
+            if (it != null) {
+                when (it.status) {
+                    Status.SUCCESS -> {
+                        Toast.makeText(this@EditProfileActivity, "Update successful", Toast.LENGTH_LONG).show()
+                        //setEditingState(false) // if this line is here, then the Profile changes to the previous value (crazy?)
```
HashEngineering, line 122: This is where I thought we could turn the observer back ON, but it the data loaded was old.

sambarboza (approved):

HashEngineering:
This will be merged after all other current PR's.

HashEngineering:
This branch has been merged with the evonet-develop branch and a crash was fixed.

PR #523, opened 2020-10-06
DashPay: Fix bug preventing viewing contacts

HashEngineering:

A == should have been >= otherwise the contacts screen was not displayed even though there were contacts.

None
## Screenshots / Videos

N/A
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/data/BlockchainIdentityBaseData.kt
+++ b/wallet/src/de/schildbach/wallet/data/BlockchainIdentityBaseData.kt
@@ -30,4 +30,7 @@ data class BlockchainIdentityBaseData(val id: Int,
         get() = creationState > BlockchainIdentityData.CreationState.NONE &&
                 creationState < BlockchainIdentityData.CreationState.DONE &&
                 creationStateErrorMessage == null
+
+    val creationComplete: Boolean
+        get() = creationState >= BlockchainIdentityData.CreationState.DONE
 }
--- a/wallet/src/de/schildbach/wallet/ui/MainActivityViewModel.kt
+++ b/wallet/src/de/schildbach/wallet/ui/MainActivityViewModel.kt
@@ -40,7 +40,7 @@ class MainActivityViewModel(application: Application) : AndroidViewModel(applica
     val canAffordIdentityCreationLiveData = CanAffordIdentityCreationLiveData(walletApplication)
 
     val hasIdentity: Boolean
-        get() = blockchainIdentity?.creationState == BlockchainIdentityData.CreationState.DONE
+        get() = blockchainIdentity?.creationComplete!!
 
     val isAbleToCreateIdentityData = MediatorLiveData<Boolean>().apply {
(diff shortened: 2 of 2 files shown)


wallet/src/de/schildbach/wallet/ui/MainActivityViewModel.kt
```diff
@@ -40,7 +40,7 @@ class MainActivityViewModel(application: Application) : AndroidViewModel(applica
     val canAffordIdentityCreationLiveData = CanAffordIdentityCreationLiveData(walletApplication)
 
     val hasIdentity: Boolean
-        get() = blockchainIdentity?.creationState == BlockchainIdentityData.CreationState.DONE
+        get() = blockchainIdentity?.creationComplete!!
```
HashEngineering, line 43: not sure why we need !! here, but the compiler complains about it.

wallet/src/de/schildbach/wallet/data/BlockchainIdentityBaseData.kt
```diff
@@ -30,4 +30,7 @@ data class BlockchainIdentityBaseData(val id: Int,
         get() = creationState > BlockchainIdentityData.CreationState.NONE &&
                 creationState < BlockchainIdentityData.CreationState.DONE &&
                 creationStateErrorMessage == null
+
+    val creationComplete: Boolean
+        get() = creationState >= BlockchainIdentityData.CreationState.DONE
```
HashEngineering, line 35: This getter could be used in other places instead of comparing to the value of DONE.

wallet/src/de/schildbach/wallet/ui/MainActivityViewModel.kt
tomasz-ludek: This i due to the safe call operator `?.` after `blockchainIdentity`. It would return `null` if `blockchainIdentity` is `null`. Your fix should be interpreted as `(blockchainIdentity?.creationComplete)!!` so that the not-null assertion operator `!!` refers to the whole result, not only to the `creationComplete` part.
Safer solution, that will not cause NPE when getting the `hasIdentity` field for `blockchainIdentity` being `null` would be:
```
val hasIdentity: Boolean
    get() = blockchainIdentity?.creationComplete ?: false
```

wallet/src/de/schildbach/wallet/data/BlockchainIdentityBaseData.kt
tomasz-ludek: Good idea!

tomasz-ludek (changes requested):

wallet/src/de/schildbach/wallet/ui/MainActivityViewModel.kt
HashEngineering: Thanks for the explanation.

HashEngineering:
@tomazs, thanks for fixing this.

PR #544, opened 2020-11-03
Downgrade zxing to avoid crash on Android 6

tomasz-ludek:

NMA-659
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/build.gradle
+++ b/wallet/build.gradle
@@ -38,7 +38,7 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:2.0.1'
     implementation "org.dashj:dashj-core:$dashjVersion"
     implementation 'com.google.guava:guava:29.0-android'
-    implementation 'com.google.zxing:core:3.4.1'
+    implementation 'com.google.zxing:core:3.3.3'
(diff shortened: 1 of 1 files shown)


wallet/build.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:2.0.1'
     implementation "org.dashj:dashj-core:$dashjVersion"
     implementation 'com.google.guava:guava:29.0-android'
-    implementation 'com.google.zxing:core:3.4.1'
+    implementation 'com.google.zxing:core:3.3.3'
```
HashEngineering, line 41: Let's add a comment here saying that 3.3.3 is the maximum to support Android 4.4 to 6.x

Otherwise, we may forget the next time we want to update libraries.

HashEngineering (changes requested):
The change needs a comment that says that `3.3.3 is the max for Android <= 6`

wallet/build.gradle
tomasz-ludek: Good idea!
Done

HashEngineering (approved):
Excellent

PR #547, opened 2020-11-06
NMA-662 | Upload profile picture to ImgUr

sambarboza:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/build.gradle
+++ b/wallet/build.gradle
@@ -48,6 +48,7 @@ dependencies {
     implementation "com.squareup.okhttp3:logging-interceptor:$ok_http_version"
 
     implementation 'com.squareup.moshi:moshi:1.11.0'
+    implementation 'com.squareup.moshi:moshi-kotlin:1.11.0'
     implementation 'org.slf4j:slf4j-api:1.7.30'
     implementation('com.github.tony19:logback-android-classic:1.1.1-6') {
         exclude(group: 'com.google.android', module: 'android')
@@ -111,6 +112,9 @@ ext {
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
+def props = new Properties()
+file("../local.properties").withInputStream { props.load(it) }
+
 android {
     compileSdkVersion 29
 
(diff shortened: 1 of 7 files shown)


wallet/build.gradle
```diff
@@ -142,6 +146,10 @@ android {
     }
 
     buildTypes {
+        all {
+            buildConfigField("String", "IMGUR_CLIENT_ID", props.getProperty("IMGUR_CLIENT_ID"))
+            buildConfigField("String", "IMGUR_CLIENT_SECRET", props.getProperty("IMGUR_CLIENT_SECRET"))
```
HashEngineering, line 151: This will allow Travis CI to pass, though builds without a local.properities will not have working Imgur functionality.
```
           buildConfigField("String", "IMGUR_CLIENT_ID", props.getProperty("IMGUR_CLIENT_ID", "000000000000000"))
            buildConfigField("String", "IMGUR_CLIENT_SECRET", props.getProperty("IMGUR_CLIENT_SECRET", "00000000000000000000000000000000000000"))
```

HashEngineering: Actually, this doesn't build, at least not for me.

HashEngineering (changes requested):
This PR needs to be merged with `evonet-develop` -- most of the conflicts require that both pieces of code be kept.  Then we merge this.

tomasz-ludek:
This is rather a straight forward thing and the implementation looks good. There is however one thing that we should consider. Basically Retrofit is designed for REST APIs requests, not for files downloading/uploading. I know it works, but probably more reliable solution would be the direct user of okhttp, as discussed here: https://stackoverflow.com/questions/25462523/retrofit-api-to-retrieve-a-png-image (it is downloading, but uploading will be similar).
I left a reference implementation, which worked well and is easy to use: https://github.com/dashevo/dash-wallet/blob/evonet-develop/wallet/src/de/schildbach/wallet/ui/dashpay/EditProfileViewModel.kt#L113
As I said however, I'm not pushing to use it, since our current implementation works good.

wallet/build.gradle
sambarboza: Fixed on https://github.com/dashevo/dash-wallet/pull/547/commits/8956f5ad0e8f43f6765c4a2adefcb9c1cdbde31a

sambarboza:
@HashEngineering I rebased this branch on `evonet-develop` to fix the conflicts.
@tomasz-ludek we'll change the way we upload the image on another ticket.

tomasz-ludek (approved):

HashEngineering (changes requested):
I made this PR for this PR to see if Travis CI will build successfully: https://github.com/dashevo/dash-wallet/pull/548

For some reason, not sure why, this PR does not successfully upload images from the gallery or a camera to **imgur**.  I had this failure on an emulator and a real device.  

My latest commit hash is 9ef244daad5f6847b25498f7796657a454509826 and the local.properties file is set up with the proper authentication values.

HashEngineering (approved):

PR #570, opened 2020-12-07
DashPay: Only add new ContactRequests to the db/listeners

HashEngineering:

NMA-715
Originally, the app would always get the most recent contact request during a query, but this will modify it so that if it does get a contact request that we already have, then it is ignored.  Additionally, after a short period of time of receiving a contact request, it won't be included in future queries.

as a result, the contacts updated listeners won't be triggered every 15 seconds, when there are no new contact requests.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/data/DashPayContactRequestDao.kt
+++ b/wallet/src/de/schildbach/wallet/data/DashPayContactRequestDao.kt
@@ -19,6 +19,9 @@ interface DashPayContactRequestDao {
     @Query("SELECT * FROM dashpay_contact_request WHERE toUserId = :toUserId")
     suspend fun loadFromOthers(toUserId: String): List<DashPayContactRequest>?
 
+    @Query("SELECT EXISTS (SELECT * FROM dashpay_contact_request WHERE userId = :userId AND toUserId = :toUserId)")
+    suspend fun exists(userId: String, toUserId: String): Boolean
+
     @Query("SELECT MAX(timestamp) FROM dashpay_contact_request")
     suspend fun getLastTimestamp() : Long
 
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
@@ -810,8 +810,15 @@ class PlatformRepo private constructor(val walletApplication: WalletApplication)
             Context.propagate(walletApplication.wallet.context)
             var encryptionKey: KeyParameter? = null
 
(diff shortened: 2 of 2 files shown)


wallet/src/de/schildbach/wallet/data/DashPayContactRequestDao.kt
```diff
@@ -19,6 +19,9 @@ interface DashPayContactRequestDao {
     @Query("SELECT * FROM dashpay_contact_request WHERE toUserId = :toUserId")
     suspend fun loadFromOthers(toUserId: String): List<DashPayContactRequest>?
 
+    @Query("SELECT EXISTS (SELECT * FROM dashpay_contact_request WHERE userId = :userId AND toUserId = :toUserId)")
```
tomasz-ludek, line 22: Nice!

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -810,8 +810,15 @@ class PlatformRepo private constructor(val walletApplication: WalletApplication)
...
             var encryptionKey: KeyParameter? = null
 
-            var lastContactRequestTime = if (dashPayContactRequestDao.countAllRequests() > 0)
-                dashPayContactRequestDao.getLastTimestamp() - DateUtils.MINUTE_IN_MILLIS * 12
+            var lastContactRequestTime = if (dashPayContactRequestDao.countAllRequests() > 0) {
+                val lastTimeStamp = dashPayContactRequestDao.getLastTimestamp()
+                // if the last contact request was received in the past 10 minutes, then query for
+                // contact requests that are 10 minutes before it.  If the last contact request was
+                // more than 10 minutes ago, then query all contact requests that came after it.
+                if (lastTimeStamp < System.currentTimeMillis() - DateUtils.MINUTE_IN_MILLIS * 10)
```
tomasz-ludek, line 818: I would prefer to use `TimeUnit.MINUTES.toMillis(10)` instead of `DateUtils.MINUTE_IN_MILLIS * 10` but it isn't a part of this story so we can let well alone ;)

tomasz-ludek (approved):
Looks good!

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
HashEngineering: Looks like our app is using both of these methods depending on the author.

PR #580, opened 2020-12-23
DashPay: Align some UI elements with iOS

HashEngineering:

NMA-607
User Profile Screen:

- [x] Transaction filter is missing on Android

- [x] Android uses long month for Dash transactions (I think this should be short month)

- [x] Android includes the year if it is the current year (I think the year should be omitted if it is the current year)

- [x] Android says --> [username] sent a contact request to you / iOS --> [username] has sent you a contact request

- [x] iOS does not show any icons for the contact request activity

- [ ] Android and iOS go out 8 decimals for the Dash amount; the designs show only 2 decimals (8 decimals are fine for the detail and confirmation screens but maybe this only needs to go out 4 decimals to fit better on small screens)

- [x] The username looks bold font on iOS and regular on Android (username should be bold when there is no display name and display name should be bold and username not bold when display name is available)?

- [x] The initial in the default profile picture is much bigger on Android vs iOS (Circle should be 130/130 / Font should be montserat light (60 font size))

Enter Amount Screen

- [x] iOS is still showing the "This transaction should settle instantly at no extra fee" (this should be removed)

Confirmation Screen

- [x] Android Header --> "Amount Sent" / iOS Header --> "Sent Successfully"

- [x] iOS Dash amount has the same font size / Android has smaller font after the 2nd decimal (should be same font for entire amount)

- [x] Android is using long month format / iOS is using short month (I think this should be short month)

- [x] Android extends to two lines for the date unnecessarily 

- [x] The size/aspect ratio of the Dash symbol needs to be updated on iOS

**Home Screen**

- [x] Android uses the first letter of the Display Name of the profile for the user avatar for transactions instead of the first letter of the username.

User Registration

- [x] The validation on Android should be changed from “Alphabets and Numbers only” to “Letters and numbers only” 

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/util/GraphicUtils.kt
```diff
@@ -0,0 +1,28 @@
+package de.schildbach.wallet.util
+
+import android.content.Context
+import android.util.DisplayMetrics
+
+object GraphicUtils {
```
tomasz-ludek, line 6: There is a better way to do both things without context:
```
public int pxToDp(int px) {
    DisplayMetrics displayMetrics = getContext().getResources().getDisplayMetrics();
    return Math.round(px / (displayMetrics.xdpi / DisplayMetrics.DENSITY_DEFAULT));
}
private fun dpToPx(dp: Int): Int {
    val density = Resources.getSystem().displayMetrics.density
    return (dp.toFloat() * density).roundToInt()
}
```

tomasz-ludek (approved):
Looks good!

PR #595, opened 2021-01-14
Avoid crash when platform is dead

tomasz-ludek:
Avoid crash when platform is dead.
Move coroutine scope definition out of PlatformRepo.
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


sambarboza (approved):

HashEngineering:
When testing this to do another story, I found that this PR prevented the first crash when the app was started for the first time, but subsequently, after the app was restarted, this PR didn't fix the crashes.

What I did to resolve this was the following.  Unfortunately, I don't have any logs from the crash, but the exception that caused the crash was `MaxRetriesReachedException`

```
Index: wallet/src/de/schildbach/wallet/ui/MainActivityViewModel.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt	(revision a459da091418858dee203b5093ceb8509783e6cb)
+++ wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt	(date 1610830336467)
@@ -1135,6 +1135,7 @@
         if (blockchainIdentityDataDao.load() == null) {
             log.info("PreDownloadBlocks: checking for existing associated identity")
 
+                try {
             val identity = getIdentityFromPublicKeyId()
             if (identity != null) {
                 log.info("PreDownloadBlocks: initiate recovery of existing identity ${identity.id.toString()}")
@@ -1144,6 +1145,10 @@
                 log.info("PreDownloadBlocks: no existing identity found")
                 // resume Sync process, since there is no Platform data to sync
                 finishPreBlockDownload()
+                    }
+                } catch (ex: MaxRetriesReachedException) {
+                    log.info("Platform is dead...")
+                    finishPreBlockDownload()
             }
         }
         // update contacts, profiles and other platform data
```

HashEngineering (commented):
See comment above.  More work needs to be done on this.  Perhaps we can simulate the crash.

tomasz-ludek:
Already fixed in some other PRs

PR #631, opened 2021-02-27
NMA-801 - DPP 0.18 Upgrade

HashEngineering:
Makes modifications to use the v18 network.  Currently this code only works on Palinka (v18), not testnet (v17)

## Related PR's and Dependencies

https://github.com/dashevo/dashj/pull/108
https://github.com/dashevo/android-dpp/pull/16
https://github.com/dashevo/dapi-client-android/pull/18
https://github.com/dashevo/android-dashpay/pull/41

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #632, opened 2021-03-03
NMA-818 DashPay:  Hide Invite Entry Points

HashEngineering:

To allow for QA testing before all invitation features are completed, I would like to be able to show/hide the entry points.

Acceptance Criteria

- The invitation entry points should be hidden by default

- Invitation notifications should be suppressed if the entry points are hidden

- The user should be able to show/hide the entry points after shaking the device in the About screen

Screens affected

- Contacts Screen (with no contacts)

- Search User Screen (before clicking on the search box)

- Search User Screen (after entering a search with no results)

- More Screen

- Notifications Screen

About Screen show say “Developer mode activated” and “Developer mode deactivated” when the screen is shaken.  Activated means that the Invite UI appears at all of the 5 above screens.  Deactivated means that the Invite UI should be hidden (this is the default state when the app is installed)

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


HashEngineering:
This PR does not depend on #631

tomasz-ludek (approved):
Looks good!
I added a tiny improvement to avoid opening invitations when developerMode is off.

PR #635, opened 2021-03-08
NMA-816 | Accept Invite | Verify Invite

sambarboza:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
@@ -92,7 +92,7 @@ class SendInviteWorker(context: Context, parameters: WorkerParameters)
             Result.success(workDataOf(
                     KEY_TX_ID to cftx.txId.bytes,
                     KEY_USER_ID to cftx.creditBurnIdentityIdentifier.toStringBase58(),
-                    KEY_INVITE_ID to Address.fromPubKeyHash(wallet.params, cftx.creditBurnPublicKeyId.bytes).toBase58(),
+                    KEY_INVITE_ID to Address.fromPubKeyHash(wallet.params, cftx.txId.bytes).toBase58(),
                     KEY_DYNAMIC_LINK to dynamicLink.uri.toString(),
                     KEY_SHORT_DYNAMIC_LINK to shortDynamicLink.shortLink.toString()
             ))
--- a/wallet/src/de/schildbach/wallet/ui/MainActivity.kt
+++ b/wallet/src/de/schildbach/wallet/ui/MainActivity.kt
@@ -151,26 +151,6 @@ class MainActivity : AbstractBindServiceActivity(), ActivityCompat.OnRequestPerm
         checkLowStorageAlert()
         detectUserCountry()
         walletApplication.startBlockchainService(true)
-
-        val txId = "ad7e7cc5611f6f2277cbccba69c3cf9dba59f03fbf31b04682d890215585bb6c"
-        viewModel.validateInvitation(txId).observe(this, Observer {
-            when {
-                it.status == Status.LOADING -> {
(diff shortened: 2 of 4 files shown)


wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
```diff
@@ -92,7 +92,7 @@ class SendInviteWorker(context: Context, parameters: WorkerParameters)
             Result.success(workDataOf(
                     KEY_TX_ID to cftx.txId.bytes,
                     KEY_USER_ID to cftx.creditBurnIdentityIdentifier.toStringBase58(),
-                    KEY_INVITE_ID to Address.fromPubKeyHash(wallet.params, cftx.creditBurnPublicKeyId.bytes).toBase58(),
+                    KEY_INVITE_ID to Address.fromPubKeyHash(wallet.params, cftx.txId.bytes).toBase58(),
```
HashEngineering, line 95: This change will throw an exception.  The original code was using the address associated with the Private Key that will be used to create the identity as an invitation id.  That address must be created using a 20-byte public key hash (sometimes called a Public Key Id) rather than a 32-byte transaction hash.

We may not need `KEY_INVITE_ID` because it was created in an earlier iteration of a previous story before the `cftx` seen in the previously set values was added to the code.   `KEY_INVITE_ID` was created to uniquely identify the worker that sends the transaction, though  `KEY_USER_ID` is also unique and could probably be used instead.


The line that needs to be changed to support your PR is this:
SendInviteWorker.kt: Line 90
```
val dynamicLink = createDynamicLink(dashPayProfile!!, cftx.getTxId().toString())
```
Or something similar to pass the Credit Funding Transaction Id to the function that creates the link.  

In future stories, we can add other information to the link to support the remaining functionality (like creating the username).  That process requires the private key that goes with this transaction asset lock output and we may need to including InstantSendLock data.  Some of the data required is subject to change based on the current limitation of Platform v18 and below.

HashEngineering (changes requested):
The comment above mentions the changes that are required.

wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
tomasz-ludek: True, the data that is required for invite verification (and later use) need to be added to deep link as a query parameter.
I created a dedicated method for this purpose:

```link = Constants.Invitation.appLinkUrl(username, nameLabel, cftx)```

https://github.com/dashevo/dash-wallet/blob/dashpay-verify-invitation/wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt#L113

HashEngineering: In that case, perhaps `createDynamicLink`'s second parameter should be a `CreditFundingTransaction`, rather than a string.  

And then this call's second parameter is also a `CreditFundingTransaction` instead of a string.
`Constants.Invitation.appLinkUrl(username, nameLabel, cftx)`

From this function, the cftx object will have all the necessary information needed to to create the url.  For now, the txid is sufficient.

tomasz-ludek: Sounds good to me.

tomasz-ludek: It is a simple change that can be made in my story (NMA-265).
Let's merge this PR if there is no other issues.

HashEngineering: Ok, great, this will be merged.

HashEngineering (approved):
This will be approved since outstanding issues will be resolved in a different PR.

PR #650, opened 2021-04-05
NMA-493 DashPay: Send Invite | History

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


common/src/main/res/values/dimens.xml
```diff
@@ -33,6 +33,4 @@
 
     <dimen name="space_medium">20dp</dimen>
 
-    <dimen name="identity_complete_text_size">18dp</dimen>
```
HashEngineering, line 36: This is moved to be with the test of these items.

wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt
```diff
@@ -587,22 +579,28 @@ class CreateIdentityService : LifecycleService() {
         //         However, a default empty profile will be saved to the local database.
         val emptyProfile = DashPayProfile(blockchainIdentity.uniqueIdString, blockchainIdentity.currentUsername!!)
         platformRepo.updateDashPayProfile(emptyProfile)
-        if (blockchainIdentityData.creationState < CreationState.DONE) {
-            platformRepo.updateIdentityCreationState(blockchainIdentityData, CreationState.DONE)
-            if (walletApplication.wallet.balance.isGreaterThan(Constants.DASH_PAY_FEE)) {
-                delay(1000L) //1s delay as required on NMA-491
```
HashEngineering, line 593: I don't think this delay worked as intended, so it was removed.

The delay was supposed to work such that 1 second after returning to the home screen after creating a username, the notification count would increase by 1.

This delay would delay when the user alert was added to the database and it would also delay username creation completion by 1 second.  To follow the original requirement, seems like a lot of work is required.

tomasz-ludek (approved):
Looks good!

PR #657, opened 2021-04-13
NMA-826 | Syncing Status | New UI

sambarboza:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


wallet/res/layout/custom_spinner_item.xml
```diff
@@ -1,11 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
HashEngineering, line 1: Deleting this file will cause merge problems when we pull `master` into `dashpay`.  DashPay currently has three Screens that still use this file.

Those screens will need to migrate away from the Spinner, but that may not happen before this story is merged into the dashpay branch.

HashEngineering (approved):

.gitignore
```diff
@@ -25,3 +25,4 @@ out/
 gradle-wrapper.jar
 wallet/prod
 wallet/.cxx
+wallet/cpp
```
HashEngineering, line 28: There should not be any files here.  If there are then we missed some other parts of deleting all the cpp source and submodules

PR #724, opened 2021-08-21
[NMA-959] Firebase events for invitations

Syn-McJ:
In order to collect usage data for the wallet, we need to add firebase events for some user actions related to invitations.
Additionally, I have added Hilt as a dependency injection framework and performed some refactoring related to migration from kotlin synthetics to view bindings.

Some notes on DI:
- Changes to gradle files necessary for Hilt are described [here](https://dagger.dev/hilt/gradle-setup) (plus some jetpack extensions), the rest is cleanup, removing duplicated dependencies, etc.
- More details on how to use Hilt in android can be found [here](https://developer.android.com/training/dependency-injection/hilt-android) and [here](https://developer.android.com/codelabs/android-hilt#0).
- CoroutineWorkers support injection in the constructor, but I faced some problems - I think the worker itself needs to be injected with Hilt for that to work. So for now I'm injecting the service into a public field. (`@HiltWorker` annotation is still required though). I've checked all the workers like UpdateProfile, SendInvite, etc, everything is working fine.
- `AnrSupervisorRunnable `is a tricky case - it's not a valid entry point, so the injection of crashlytics has to be done with some workarounds or the error should be propagated to be logged elsewhere. I skipped this case for now.
- I replaced some of the usages of `ViewModelProvider` with `by viewModels()` and `by activityViewModels()` extensions from jetpack - which is a newer and more concise way of injecting a view model. It's not necessary for Hilt to work though.

On view bindings:
- Be careful with the \<merge\> tag. To access elements through viewBindings, you'll need to bind layout with this tag separately, but even this didn't work for me in the case of list adapters for some reason. More investigation is required. Some info on this can be found [here](https://betterprogramming.pub/exploring-viewbinding-in-depth-598925821e41).

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/invite/InviteFilterSelectionDialog.kt
```diff
@@ -41,51 +40,45 @@ class InviteFilterSelectionDialog(private val owner: ViewModelStoreOwner) : Base
...
-            third.setOnClickListener {
-                sharedViewModel.filterBy.call(InvitesHistoryViewModel.Filter.CLAIMED)
-                dismiss()
-            }
+        binding.first.setOnClickListener {
+            sharedViewModel.setFilter(InvitesHistoryViewModel.Filter.ALL)
+            dismiss()
         }
-
-        dialog?.setOnShowListener { dialog ->
```
Syn-McJ, line 73: Removed since BaseBottomSheetDialogFragment already has this code

wallet/src/de/schildbach/wallet/ui/invite/InviteFilterSelectionDialog.kt
```diff
@@ -41,51 +40,45 @@ class InviteFilterSelectionDialog(private val owner: ViewModelStoreOwner) : Base
...
+            sharedViewModel.setFilter(InvitesHistoryViewModel.Filter.CLAIMED)
+            dismiss()
         }
     }
 
-    override fun onActivityCreated(savedInstanceState: Bundle?) {
-        super.onActivityCreated(savedInstanceState)
-        sharedViewModel = activity?.run {
-            ViewModelProvider(owner)[InvitesHistoryFilterViewModel::class.java]
-        } ?: throw IllegalStateException("Invalid Activity")
```
Syn-McJ, line 88: Replaced `ViewModelProvider `with `viewModels({ owner })` and removed this method since it doesn't seem to be needed anymore? Let me know if I'm missing something

wallet/src/de/schildbach/wallet/ui/dashpay/ContactsFragment.kt
```diff
@@ -108,116 +106,122 @@ class ContactsFragment : BottomNavFragment(R.layout.fragment_contacts_root), Tex
...
             }
-        }
 
-        search_for_user.setOnClickListener {
-            onSearchUser()
-        }
-
-        search_for_user_suggestions.setOnClickListener {
-            onSearchUser()
-        }
```
Syn-McJ, line 145: Removed since `onSearchUser `is already called by `ContactSearchResultsAdapter.Listener`

wallet/build.gradle
```diff
@@ -64,35 +70,27 @@ dependencies {
...
     implementation 'com.github.MikeOrtiz:TouchImageView:3.0.3'
 
-    // Import the BoM for the Firebase platform
-    implementation platform('com.google.firebase:firebase-bom:26.1.1')
-
     // Declare the dependencies for the Crashlytics and Analytics libraries
     // When using the BoM, you don't specify versions in Firebase library dependencies
-    implementation 'com.google.firebase:firebase-crashlytics'
-    implementation 'com.google.firebase:firebase-analytics'
+    implementation platform("com.google.firebase:firebase-bom:$firebaseVersion")
```
HashEngineering, line 93: Are crashlytics and analytics included elsewhere?

wallet/src/de/schildbach/wallet/ui/dashpay/ContactsFragment.kt
```diff
@@ -108,116 +106,122 @@ class ContactsFragment : BottomNavFragment(R.layout.fragment_contacts_root), Tex
...
+        }
 
-        dashPayViewModel.sendContactRequestState.observe(viewLifecycleOwner, Observer {
+        dashPayViewModel.sendContactRequestState.observe(viewLifecycleOwner) {
             imitateUserInteraction()
             contactsAdapter.sendContactRequestWorkStateMap = it
-        })
-        dashPayViewModel.contactsUpdatedLiveData.observe(viewLifecycleOwner, Observer<Resource<Boolean>> {
+        }
+        dashPayViewModel.contactsUpdatedLiveData.observe(viewLifecycleOwner) {
```
HashEngineering, line 202: I like this simplification.

wallet/build.gradle
Syn-McJ: Yes, crashlytics and analytics go to the common module since AnalyticsService is located there, given that we're gonna use events and possibly crash logging in liquid and uphold modules too. BoM is still left here for inappmessaging and dynamic links, although it might be the case that it is only required in the common module now.

wallet/src/de/schildbach/wallet/ui/dashpay/ContactsFragment.kt
```diff
@@ -83,6 +79,9 @@ class ContactsFragment : BottomNavFragment(R.layout.fragment_contacts_root), Tex
 
     override val navigationItemId = R.id.contacts
 
+    private var _binding: FragmentContactsRootBinding? = null
+    private val binding get() = _binding!!
```
Syn-McJ, line 83: Small fix for ContactsFragment: replaced lateinit bindings with nullable bindings. Lateinit bindings [aren't recommended](https://developer.android.com/topic/libraries/view-binding#fragments) for fragments since fragments outlive their views and a memory leak can be introduced if binding isn't nullified in `onDestroyView()`

wallet/src/de/schildbach/wallet/Constants.java
```diff
@@ -374,4 +374,32 @@ public void log(final String message) {
...
+            String CREATED_TAG = "invite_created_send_with_tag";
+            String DETAILS_TAG = "invite_details_send_again_with_tag";
+            String CREATED_PREVIEW = "invite_created_preview";
+            String DETAILS_PREVIEW = "invite_details_preview";
+            String CREATE_HISTORY = "invite_create_from_history";
+            String HISTORY_FILTER = "invite_history_filter";
+            String DETAILS = "invite_details";
+            String NEW_WALLET = "invite_new_wallet";
+        }
+    }
```
tomasz-ludek, line 404: There will be much more Events in the future so I would argue to move this to a separate file (e.g. _AnalyticsConstants.kt_)

Syn-McJ: Makes sense. Do you think this should be a single file for all events in a common module or better to break it up by modules?

wallet/AndroidManifest.xml
```diff
@@ -430,6 +430,13 @@
                 android:resource="@xml/file_provider" />
         </provider>
 
+        <provider
+            android:name="androidx.work.impl.WorkManagerInitializer"
+            android:authorities="${applicationId}.workmanager-init"
+            android:enabled="false"
+            android:exported="false"
+            tools:replace="android:authorities" />
```
Syn-McJ, line 438: Turning off default worker initialization is required for hilt injection to work. Recheck all the CoroutineWorkers, everything is working fine.

wallet/src/de/schildbach/wallet/ui/dashpay/ContactsFragment.kt
HashEngineering: As long as this works.  Some time ago we had one of these `setOnClickListener`s missing and one of the buttons didn't call `onSearchUser`

wallet/src/de/schildbach/wallet/ui/invite/InviteFilterSelectionDialog.kt
HashEngineering: We didn't know about `viewModels({ owner })`

wallet/src/de/schildbach/wallet/ui/dashpay/ContactsFragment.kt
Syn-McJ: Rechecked again, working fine. There is another way to handle events from the adapter, you declare an action like `var onActionClick: (View) -> Unit = {}`, call it on click of your list element, and simply assign a method to it from the consumer. That way there is no need to deal with interface implementations.

wallet/src/de/schildbach/wallet/Constants.java
HashEngineering: Good question.  We can move this to new file(s) in the next PR's.

Event names will only be used in a certain module.  Maybe it's easier to put them all in a single file in `common`.

wallet/src/de/schildbach/wallet/ui/dashpay/ContactsFragment.kt
HashEngineering: ok, then we can do this in a new PR / Jira story.

HashEngineering (approved):
One suggestion regarding handling events from an adapter will be noted in our refactor log.

Events appear in Firebase.

Approved.

PR #734, opened 2021-08-30
[ NMA-988]: Default fiat currency

ClaudeHangui:
Update the design of the screen which displays the list of supported fiat currencies
PS: There are no translations for the label **Select Currency**

## Checklist:


wallet/src/de/schildbach/wallet/ui/ExchangeRatesFragment.java
```diff
@@ -390,19 +399,31 @@ void setExchangeRates(List<de.schildbach.wallet.rates.ExchangeRate> exchangeRate
         }
     }
 
+    private Integer getFlagFromCurrencyCode(String currencyCode) {
+        if (currencyCode.equalsIgnoreCase("try")) return R.drawable.turk;    //we can't use 'try' as the recource name for turkish flag as try is a reserved keyword in java. A workaround is to use rename the turkish flag resource
```
tomasz-ludek, line 403: We should avoid such weird exceptions. I would suggest to add a prefix to each resource name (e.g. "currency_code_try", or something similar). Additionally, names with such prefix would be more self-descriptive.

tomasz-ludek (changes requested):

wallet/src/de/schildbach/wallet/ui/ExchangeRatesFragment.java
ClaudeHangui: The implementation of this method had been updated accordingly

ClaudeHangui (commented):
- Our design for the exchange rates screen takes into account the 2 entry points we have: **Settings** and **SendPayment** flows. The action to display this new UI from the **SendPayment** flow has not yet been implemented
- We removed unused variable dependencies and refactored the class as much as we could

HashEngineering (approved):
looks good and works well

HashEngineering (changes requested):
There is a merge conflict with the Firebase Events for Settings.

ClaudeHangui:
Conflicts have been fixed

PR #735, opened 2021-09-01
refactor: set application taskAffinity="" instead of each activity (NMA-997)

HashEngineering:

NMA-997

Documentation related to the Strandhogg v1 fix indicates that all activities or `application` need the `taskAffinity=""` value set.

If we do this, will some of our `magical` fixes to get around a new task being created with intents from other apps (web browsers, custom tabs, invites) still be necessary?

When this is merged into the `dashpay` branch, there may be other activities with `taskAffinity=""` that could be changed.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/AndroidManifest.xml
+++ b/wallet/AndroidManifest.xml
@@ -60,12 +60,12 @@
         android:label="@string/app_name"
         android:largeHeap="true"
         android:theme="@style/My.Theme"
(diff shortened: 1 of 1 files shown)


tomasz-ludek (approved):
Nice :)

PR #736, opened 2021-09-02
[NMA-977] DashPay GitHub Actions

Syn-McJ:
GitHub actions for DashPay.

`dashwallet.yml` - workflow for pushing and pull requests in the master branch. Runs tests and builds for PR and for push it tests, builds and distributes apk to Firebase.
`manual_distribution.yml` - workflow for triggering build and Firebase distribution manually from Actions UI. This doesn't show up in Actions right now.
`fastlane/Fastfile` - fastlane file with pipelines for building, testing and distributing.

https://github.com/dashevo/dash-wallet/pull/741

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


fastlane/Appfile
```diff
@@ -0,0 +1,2 @@
+json_key_file("") # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
+package_name("hashengineering.darkcoin.wallet") # e.g. com.krausefx.app
```
HashEngineering, line 2: is this supposed to be the package name of the `prod` flavor?

HashEngineering: or should it be the staging flavor?

Syn-McJ: Package name here doesn't actually affect our pipelines right now. It is used for interacting with Google Play. For example, fastlane can download screenshots and metadata from the store in a separate folder, and then you can use lanes to easily update them.

PR #739, opened 2021-09-03
DashPay: Onboarding back button & No Authentication for Invites (NMA-990, 886)

HashEngineering:

* Create New Wallet with Invite → Home Screen, press back goes to the onboarding process.  We need to close all the activities before the home screen in this case.

* During onboarding, pressing back does not go back to the Onboarding screen (the onboarding screen was finished earlier).  Instead, the app closes down.
* If onboarding is complete (wallet is initialized), and the Lock Screen is on or the app is autologout, then activating an invite will start the Use Invite Flow without Authentication.
* 
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/RestoreWalletFromSeedActivity.kt
```diff
@@ -160,12 +160,4 @@ class RestoreWalletFromSeedActivity : RestoreFromFileActivity() {
...
         }
     }
-
-    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
-        super.onActivityResult(requestCode, resultCode, data)
-        if (requestCode == SET_PIN_REQUEST_CODE && resultCode == Activity.RESULT_OK) {
-            setResult(Activity.RESULT_OK)
-            finish()
-        }
-    }
```
HashEngineering, line 170: Using `finishAffinity` will remove the need to setResult to tell previous Activities to close.

wallet/src/de/schildbach/wallet/ui/invite/InviteHandler.kt
```diff
@@ -108,7 +125,7 @@ class InviteHandler(val activity: AppCompatActivity, private val analytics: Anal
                         }
                         walletApplication.wallet != null -> {
                             log.info("the invite is valid, starting AcceptInviteActivity with invite: ${invite.link}")
-                            mainTask.startActivity(activity.applicationContext, AcceptInviteActivity.createIntent(activity, invite, false), null)
+                            mainTask.startActivity(activity.applicationContext, MainActivity.createIntent(activity, invite), null)
```
HashEngineering, line 128: If there is a wallet created, then we will activate the Home Screen (MainActivity) with the invite for processing after the user has entered their PIN (if the app is logged out)

HashEngineering: `mainTask.startActivity` is being used instead of `activity.startActivity` to get around some duplicate instance problems that happen as a result of setting `taskAffinity=""` from previous work.

However, in this story, I am also adding `taskAffinity=""` to the application level, so all activities will have the same affinity.  This may help reduce or eliminate some duplicate instance problems and render other fixes we had to being unnecessary.  There is another story to address duplicate instances.

wallet/src/de/schildbach/wallet/ui/MainActivity.kt
```diff
@@ -435,12 +445,32 @@ class MainActivity : AbstractBindServiceActivity(), ActivityCompat.OnRequestPerm
...
+        val acceptInviteIntent = AcceptInviteActivity.createIntent(this, invite, false)
+        startActivity(acceptInviteIntent)
+    }
+
+    override fun onUnlocked() {
+        if (pendingInvite != null) {
+            handleInvite(pendingInvite!!)
+            pendingInvite = null // clear the invite
+        }
+    }
```
HashEngineering, line 458: here we have the code to activate an invite after the user enters a PIN/fingerprint to dismiss the Lock Screen.

wallet/src/de/schildbach/wallet/ui/invite/OnboardFromInviteActivity.kt
```diff
@@ -120,8 +131,16 @@ class OnboardFromInviteActivity : AppCompatActivity() {
...
     }
+
+    override fun onBackPressed() {
+        if (mode == Mode.STEP_3) {
+            startActivity(goNextIntent)
+            finish()
+        } else {
+            super.onBackPressed()
+        }
+    }
```
HashEngineering, line 145: Before adding this code, the back button on Step 3 (after the wallet is created) would close the app.  Therefore it would skip over the Secure Wallet step and possibly forget to create the username (not sure about this though).

The simple fix was to have the back button go to the next step in the process.  If the user pressed back on the next screen, they will go to the Home Screen.

wallet/src/de/schildbach/wallet/ui/invite/InviteHandler.kt
```diff
@@ -185,21 +214,21 @@ class InviteHandler(val activity: AppCompatActivity, private val analytics: Anal
                 (errorMessage.contains("IdentityAssetLockTransactionOutPointAlreadyExistsError")) -> {
                     showInviteAlreadyClaimedDialog(blockchainIdentityData.invite!!)
                     // now erase the blockchain data
-                    getInstance().clearBlockchainData()
+                    getInstance().clearBlockchainIdentityData()
```
HashEngineering, line 217: The method was renamed because it does not clear the blockchain, but only Blockchain Identity Data.

wallet/src/de/schildbach/wallet/ui/invite/InviteHandler.kt
Syn-McJ: Maybe better to change the message in log.info as well

Syn-McJ (approved):
Looks alright

PR #737, opened 2021-09-03
[NMA-995] Firebase Settings & Tools events

Syn-McJ:
Firebase events for Tools & Settings

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [x] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #742, opened 2021-09-05
[NMA-996] Security firebase events

Syn-McJ:
Firebase events for Security settings

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.dash.wallet.common.services.analytics
+
+object AnalyticsConstants {
+    const val CALLING_ACTIVITY = "calling_activity"
```
HashEngineering, line 20: looks like this file will have a merge conflict with #737.

Syn-McJ: that's ok, merge one - I'll fix the other.

HashEngineering (commented):
Approved

PR #749, opened 2021-09-10
fix: keep payment flow unlocked if activated from Lock Screen (NMA-1011)

HashEngineering:

There are many ways to initiate a payment flow.

We found that from the Lock Screen, after pressing Scan to Pay and scanning a QR code, then the Lock Screen was displayed again.

In this particular payment flow, the Lock Screen should not be displayed.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):
Looks good

PR #760, opened 2021-09-17
feat (dashpay): enable invites for all builds

HashEngineering:

Invite features do not depend on developer mode.

None
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
@@ -479,9 +479,11 @@ class PlatformRepo private constructor(val walletApplication: WalletApplication)
             return 0
         }
 
-        val alert = userAlertDao.load(date)
-        if (alert != null) {
-            count++
+        if (shouldShowAlert()) {
+            val alert = userAlertDao.load(date)
+            if (alert != null) {
+                count++
+            }
         }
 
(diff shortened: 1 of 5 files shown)


wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -475,13 +475,15 @@ class PlatformRepo private constructor(val walletApplication: WalletApplication)
 
     suspend fun getNotificationCount(date: Long): Int {
         var count = 0
-        // Developer Mode Feature
-        if (walletApplication.configuration.developerMode && shouldShowAlert()) {
```
Syn-McJ, line 479: `shouldShowAlert()` is not called anymore in the changed code and the method seems to be not used, is that ok?

HashEngineering: Thanks for catching this.  `shouldShowAlert()` needs to be called because we allow the user to dismiss the alert to never be shown again.  By removing `shouldShowAlert()` the user would always see that alert.

HashEngineering: Actually, `shouldShowAlert()` looks at other things to determine if that the alert should be displayed on the screen.

Syn-McJ (approved):
Looks good now

PR #762, opened 2021-09-20
- refactor EnterAmountSharedViewModel, EnterAmountViewModel: use muta…

ClaudeHangui:
- refactor EnterAmountSharedViewModel, EnterAmountViewModel: use mutableLiveData to update LiveData
- update Configuration: set preference to hold the default currency code selected during send payment
- ExchangeRatesAdapter: distinguish between currency code saved during `Send payment` and `Update Settings` flows
- update ExchangeRatesRepository: set new implementation to get exchange rate from DB using a currency code
- EnterAmountFragment.kt: update design and implementation to handle the `Send payment` flow

- Update the design and feature implementation of the `Send Payment` flow
- Usage of LiveData in the view: For a view to be updated as per a change observed in the livedata linked to it, we need to use `MutableLiveData`to be able to update the LiveData associated to it so we had to refactor the usage of our livedata objects in the fragment
- As our exchange rate is constantly updated, we had to set a new implementation in the repository so as not to return a livedata. Instead we defined an `Executor` class to handle all the operations done in background as well as publishing the result of the operations in the main thread with the help of a callback
- As per the requirements for each fiat currency, we need to display the currency code together with it's symbol. For currencies being used by multiple countries, we needed to set a `Locale` to be able to get the currency code and symbol.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -308,7 +309,7 @@ public String getExchangeCurrencyCode() {
     public void setExchangeCurrencyCode(final String exchangeCurrencyCode) {
         prefs.edit().putString(PREFS_KEY_EXCHANGE_CURRENCY, exchangeCurrencyCode).apply();
     }
-
+    //todo: preference not used
     public boolean getExchangeCurrencyCodeDetected() {
         return prefs.getBoolean(PREFS_KEY_EXCHANGE_CURRENCY_DETECTED, false);
     }
```
HashEngineering, line 315: I don't know why Android Studio marks this as unused, but it is used in `WalletActivity.detectUserCountry()`

Is it because this method is in the `common` module, but it is used in the `wallet` module?

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
```diff
@@ -50,4 +54,92 @@ public static void showToast(Context context, String messages) {
...
+                break;
+            case "usd":
+                currentLocale = Locale.US;
+                break;
+        }
+        currentCurrencySymbol = TextUtils.isEmpty(currentLocale.getLanguage()) ?
+                currencySymbol(currencyCode.toLowerCase(Locale.ROOT)) : Currency.getInstance(currentLocale).getSymbol();
+
+        return currencyCode.equalsIgnoreCase(currentCurrencySymbol) ? currencyCode :
+                currencyCode.concat(" ").concat(currentCurrencySymbol);
```
HashEngineering, line 143: I remember getting a warning about concatenating strings that were like this `a + " " + b`, where `a` and `b` were Strings.

The warning said there would be problems rendering the string in come locales that might use RTL.  Not quite sure.  so I created a string "%s %s" in the resources.  I don't know if that is necessary in this case.

HashEngineering: I have question about the designs regarding the need for this method.

wallet/src/de/schildbach/wallet/ui/EnterAmountFragment.kt
```diff
@@ -284,5 +319,18 @@ class EnterAmountFragment : Fragment() {
         input_symbol_dash.visibility = if (viewModel.dashToFiatDirectionValue) View.VISIBLE else View.GONE
         calc_amount_symbol.visibility = if (viewModel.dashToFiatDirectionValue) View.VISIBLE else View.GONE
         calc_amount_symbol_dash.visibility = if (viewModel.dashToFiatDirectionValue) View.GONE else View.VISIBLE
+        input_select_currency_toggle.visibility = if (viewModel.dashToFiatDirectionValue) View.GONE else View.VISIBLE
+        calc_select_currency_toggle.visibility = if (viewModel.dashToFiatDirectionValue) View.VISIBLE else View.GONE
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        if(resultCode == Activity.RESULT_OK && requestCode == 99) {
```
HashEngineering, line 327: let's make 99 a constant

wallet/src/de/schildbach/wallet/ui/send/SendCoinsActivity.java
```diff
@@ -167,4 +167,9 @@ public boolean onOptionsItemSelected(final MenuItem item) {
 
         return super.onOptionsItemSelected(item);
     }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+    }
```
HashEngineering, line 174: Do we need this?

wallet/src/de/schildbach/wallet/ui/EnterAmountFragment.kt
```diff
@@ -140,6 +154,17 @@ class EnterAmountFragment : Fragment() {
...
+            val intent = Intent(activity, ExchangeRatesActivity::class.java)
+            intent.putExtra(ARG_SHOW_AS_DIALOG, true)
+            startActivityForResult(intent, 99)
+        }
+
+        calc_select_currency_toggle.setOnClickListener {
+            val intent = Intent(activity, ExchangeRatesActivity::class.java)
+            intent.putExtra(ARG_SHOW_AS_DIALOG, true)
+            startActivityForResult(intent, 99)
+        }
```
HashEngineering, line 167: lets make 99 a constant

HashEngineering (changes requested):
There are a few small items to change and a design disagreement which if not resolved with the the designer and PO, we can leave it as is in the PR and change later.

common/src/main/java/org/dash/wallet/common/Configuration.java
ClaudeHangui: Just noticed it, this is really bizarre ! I tracked the `getter` and `setter` for this pref and I definitely see it being called in the `WalletActivity`; but the task I worked on is this PR is also inside the `wallet` module, so I think it's AS that is messing up things: I for one I'm also having issues referencing dependencies in my kotlin files, I need to explicitly declare the path , it doesn't import it automatically)

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
ClaudeHangui: you're right, we should be using string resources for this since we work with multiple locales. i'll fix this
Regarding the function itself, it is required to render the currency code together with the associated currency symbol and we have some currency codes not related to a specific/single country but a set of countries, Euro for example. So using the existing `GenericUtils.currencySymbol()` will not work in this case as it'll return `Euro` as the currency symbol instead of `€`. So to achieve the desired result, we needed to define a `Locale` with whatever language, but definitely a country using the currency, like Germany of France for example. So we applied this rule with all the currencies used by multiple countries.

wallet/src/de/schildbach/wallet/ui/EnterAmountFragment.kt
ClaudeHangui: Got it !

wallet/src/de/schildbach/wallet/ui/send/SendCoinsActivity.java
ClaudeHangui: Actually yes !,
Even though the fragment is the one making the `startActivityForResult()` call, the hosting activity always gets the first shot at handling the result; this makes sense when you consider the modularity of fragments.
So once we override the `onActivityResult()` in the activity, the fragment gets a shot at handling the result

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
ClaudeHangui: > I have question about the designs regarding the need for this method.

Perhaps we can have a call regarding this to clear up things

HashEngineering: We will leave this code in for now and address it in a different story concerned with formats of values.

wallet/src/de/schildbach/wallet/ui/send/SendCoinsActivity.java
HashEngineering: The reason why I don't think we need this is because this method is only called the super method from the base class.  Isn't that the same as not having this method at all?

HashEngineering (approved):

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
ClaudeHangui: I pushed some changes on this implementation: we still return currency code + symbol. Could you please check it out and confirm with me whether or not I should rollback

HashEngineering (changes requested):
Recent merges have caused some conflicts.

wallet/res/layout/enter_amount_fragment.xml
```diff
@@ -4,6 +4,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    xmlns:tools="http://schemas.android.com/tools"
```
HashEngineering, line 7: this line is duplicating and is causing the CI system to fail.
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':wallet:merge_testNet3DebugResources'.
> /home/runner/work/dash-wallet/dash-wallet/wallet/res/layout/enter_amount_fragment.xml:7:51: Error: Attribute "xmlns:tools" was already specified for element "LinearLayout".
```

ClaudeHangui: duplicate removed and pushed !

HashEngineering (approved):
Approved.

PR #756, opened 2021-09-15
[NMA-1017] Firebase events for the home screen

Syn-McJ:
## Issue being fixed or feature implemented
Firebase events for the home screen. Also moved [TransactionsFilterDialog](https://github.com/dashevo/dash-wallet/blob/master/wallet/src/de/schildbach/wallet/ui/TransactionsFilterDialog.kt) to view bindings and did some cleanup in there.

## Related PR's and Dependencies
https://github.com/dashevo/dash-wallet/pull/724
https://github.com/dashevo/dash-wallet/pull/737
https://github.com/dashevo/dash-wallet/pull/742
https://github.com/dashevo/dash-wallet/pull/753

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

- [x] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #753, opened 2021-09-13
[NMA-961] Firebase Liquid events

Syn-McJ:
## Issue being fixed or feature implemented
Liquid events for Firebase

## Related PR's and Dependencies
https://github.com/dashevo/dash-wallet/pull/724
https://github.com/dashevo/dash-wallet/pull/737
https://github.com/dashevo/dash-wallet/pull/742

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

- [x] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #763, opened 2021-09-20
[NMA-1019]  Firebase events for users and contacts

Syn-McJ:
Added events for users and contacts and turned off logging for debug builds.

Additionally, I have moved some dashpay specific fragments and activities to view bindings, and replaced `TextWatcher` implementations with `doAfterTextChanged` extension method, which allows getting rid of the interface and unneeded methods implementations.

https://github.com/dashevo/dash-wallet/pull/724
https://github.com/dashevo/dash-wallet/pull/737
https://github.com/dashevo/dash-wallet/pull/742
https://github.com/dashevo/dash-wallet/pull/753
https://github.com/dashevo/dash-wallet/pull/756

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/EditProfileViewModel.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/EditProfileViewModel.kt
@@ -322,7 +322,6 @@ class EditProfileViewModel @Inject constructor(
                         config.imgurDeleteHash = imgurUploadResponse.data.deletehash
                         val avatarUrl = imgurUploadResponse.data.link
                         Log.d("AvatarUrl", avatarUrl)
-                        dashPayProfile?.avatarUrl = avatarUrl
                         log.info("imgur: upload successful (${response.code()})")
                         profilePictureUploadLiveData.postValue(Resource.success(avatarUrl))
                     } else {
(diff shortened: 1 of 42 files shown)


wallet/src/de/schildbach/wallet/ui/dashpay/EditProfileViewModel.kt
```diff
@@ -322,7 +322,6 @@ class EditProfileViewModel @Inject constructor(
                         config.imgurDeleteHash = imgurUploadResponse.data.deletehash
                         val avatarUrl = imgurUploadResponse.data.link
                         Log.d("AvatarUrl", avatarUrl)
-                        dashPayProfile?.avatarUrl = avatarUrl
```
Syn-McJ, line 319: This doesn't seem to be needed (it's not used in the similar `uploadToGoogleDrive` method and doesn't affect the logic), but it's messing up later checks on whether the new avatar URL is different from the original. Let me know if I'm missing something.

HashEngineering: Good question.  Let me see.

HashEngineering: I believe that you are correct.  `dashPayProfile` contains the current profile data from the Room database and we shouldn't be changing any of its member fields.  It should only updated when the room database is updated.

If we change the image, then that image url is stored here `profilePictureUploadLiveData.value!!.data` after uploading.  When "Save" is pressed, then the updated profile with the new images or text is broadcast to the Platform Network and it is also saved to the Room database.

HashEngineering (approved):
Excellent.

PR #766, opened 2021-09-24
Update  Shortcut Bar

hadia:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -215,15 +217,25 @@ public void onClick(View v) {
                 } else if (v == shortcutsPane.getImportPrivateKey()) {
                     SweepWalletActivity.start(WalletActivity.this, true);
                 }
+                else if (v == shortcutsPane.getExplore()) {
+                    startActivity(new Intent(WalletActivity.this, ExploreActivity.class));
+                }
             }
         });
         showHideSecureAction();
+        refreshIfUserHasBalance();
```
HashEngineering, line 226: This doesn't seem like it will be live as the balance can change from zero to non-zero or from non-zero to zero.

This is only called within onCreate.  If the balance changes, then the activity has to be restarted to reflect the correct buttons on this shortcut bar.

wallet/src/de/schildbach/wallet/ui/widget/ShortcutsPane.kt
```diff
@@ -100,42 +111,54 @@ class ShortcutsPane(context: Context, attrs: AttributeSet) : LinearLayout(contex
...
-        } else View.GONE
-        if (!isSmallScreen) {
-            displayed.add(receiveButton)
         }
-        val numberOfButtons = if (isSmallScreen) 3 else 4
-        secondaryItems.forEach {
-            it.visibility = if (displayed.size < numberOfButtons) {
-                displayed.add(it)
-                View.VISIBLE
-            } else View.GONE
```
HashEngineering, line 138: For smaller screens, only 3 buttons should be displayed.  The story does not define what should be displayed if there are only three places for shortcut buttons.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
hadia: Fix sync part

HashEngineering (approved):
Excellent work.

PR #784, opened 2021-10-17
NMA-783:  Merchants Map

ClaudeHangui:

1. API Key Storage: As described [here](https://developers.google.com/maps/documentation/android-sdk/start), Google recommends using **Secret Gradle Plugin (SGP)** for managing API keys. With this plugin, the key is stored in (and accessed from) the _local.properties_ file. Usage of SGP required an upgrade of our AGP and Gradle versions. We did not use the latest version of SGP as this requires migrating our AGP and GP versions from 4.x to 7.x. As our app is right now, this is not a requirement 
NB: I recommend to remove the api_key value stored in the _string.xml_ file
2. Location Permission: The app as it is now requires us to get the user's location while in foreground ONLY; as such irrespective of the android version (M, Q or above) we require only the **ACCESS_FINE_LOCATION** which is more precise than the **ACCESS_COARSE_LOCATION**
3. Location update: We provide the logic to get latest user's location via a custom class. The user's coordinates are provided as a Flow

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


ClaudeHangui:
@Syn-McJ 

**The user location circle and dash markers seem a bit too big (see the design). It might be better to just set the default user location icon - won't need custom drawable then and it should look the same, only smaller.**
Well, the design shows a custom marker, so I'll just have to adjust it's size. I'll check with Brian to see if we use the default marker rather than the custom one seen on the design

**Dash icons should have a shadow according to the design - without a shadow, when the markers overlap each other it becomes just a mash of white and blue. If a shadow is too hard to make (although you might just use a png asset), then there should at least be a border of some kind**
I personally don't see any shadow, but this is something that can be fixed with an asset indeed. I'll see with Rost !

**The transparent blue circle usually indicates the accuracy of the location tracker (see the design). As I understand you're indicating 50 miles radius with it - that's not required, user will zoom in and see all of their dash markers on a blue background which is weird. Instead, the map should just be zoomed in to fit 50 miles radius on the screen.**
Good point ! I agree with the fact that zooming might have the blue circle overlap with the closest merchant markers. Removing this will improve UX. ​I'll need to validate this modification with Brian

This is just the 1st iteration related to this feature. As you know, there is another story related to the map, were I'll be dealing with improving the UI/UX regarding location permission as we discussed last week. I'll be bringing up all the aforementionned refinements

HashEngineering (approved):
Looks good.

features/exploredash/build.gradle
```diff
@@ -76,9 +76,16 @@ dependencies {
     kapt "com.github.bumptech.glide:compiler:$glideVersion"
     api 'androidx.paging:paging-runtime-ktx:3.0.1'
 
+    // Fused Location Provider, Google Maps service $ Kotlin extension for map sdk
+    implementation 'com.google.android.gms:play-services-location:18.0.0'
```
HashEngineering, line 80: We got an email about 18.0.0 being the most recent version.  
```
We're writing to inform you that we released a major update of Maps SDK for Android (version 18.0.0), 
and it is now available via Google Play services. 
```

Syn-McJ (approved):
Looks good. There are some potential improvements that should probably go into the second map story.

PR #801, opened 2021-10-29
DashPay: update firebase (NMA-1064)

HashEngineering:

Updating Firebase BOM to 28.4.2 also requires updating DashJ to use protobuf-javalite 3.14.0

https://github.com/dashevo/dashj/pull/137
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/AndroidManifest.xml
+++ b/wallet/AndroidManifest.xml
@@ -96,7 +96,6 @@
             android:configChanges="keyboard|keyboardHidden"
             android:exported="true"
             android:label="@string/app_name_short"
-            android:launchMode="singleTask"
             android:screenOrientation="portrait"
             android:theme="@style/LockScreenActivity.Child.Theme"
             android:windowSoftInputMode="stateAlwaysHidden" />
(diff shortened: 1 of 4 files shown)


wallet/AndroidManifest.xml
```diff
@@ -96,7 +96,6 @@
             android:configChanges="keyboard|keyboardHidden"
             android:exported="true"
             android:label="@string/app_name_short"
-            android:launchMode="singleTask"
```
Syn-McJ, line 99: Was there a problem with this solution?

Syn-McJ (approved):
Looks good

wallet/AndroidManifest.xml
HashEngineering: Yes, it resulted in multiple tasks or duplicate instances on the recent app lists when the app was started.

To solve the original issue, we also need to specify theses flags along with removing `singleTask`:
```
intent.addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP or Intent.FLAG_ACTIVITY_CLEAR_TOP)
```

This is one of the side effects of using `taskAffinity=""` which prevents a particular vulnerability called Strandhogg v1.  Patching up this vulnerability created many problems for us.

Syn-McJ: Gotcha.

PR #804, opened 2021-11-07
[NMA-1036 In progress] Explore Dash physical search results

Syn-McJ:
This PR unblocks map-related stories for the explore-dash (specifically NMA-785)

- Redesign of the existing search results functionality
- Search radius picker
- Filtering by screen bounds or screen center + radius
- Some refactoring

Changes related to the map:
- Moved map-related code into a separate fragment
- Decoupled location tracking and permission requests from the map
- Minor fixes for values and icon sizes

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/build.gradle
```diff
@@ -10,7 +10,7 @@ android {
     compileSdkVersion 30
 
     defaultConfig {
-        minSdkVersion 19
+        minSdkVersion 23
```
HashEngineering, line 13: excellent.  This branch was left out of the Andriod 6 minimum PR's.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -36,32 +41,41 @@ enum class NavigationRequest {
...
     private val atmDao: AtmDao,
-    private val locationUpdatesUseCase: UserLocationState
+    private val locationProvider: UserLocationState
 ) : ViewModel() {
     companion object {
         const val QUERY_DEBOUNCE_VALUE = 300L
         const val PAGE_SIZE = 100
         const val MAX_ITEMS_IN_MEMORY = 300
+        const val METERS_IN_MILE = 1609.344
+        const val METERS_IN_KILOMETER = 1000.0
```
HashEngineering, line 61: so internally distances will be in meters? And the user will be seeing km or miles?

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
```diff
@@ -0,0 +1,340 @@
...
+            // TODO: we probably shouldn't reset all markers but calculate the difference instead,
+            // otherwise when user pans the map, everything blinks
+            resetMap()
+
+            // TODO: For the 1st iteration of this feature, we shall limit the number of markers to be displayed
+            if (results.isNotEmpty()) {
+//                if (results.size < 20) {
+                    setMarkers(results)
+//                } else setMarkers(results.shuffled().subList(0, 20))
+            }
```
HashEngineering, line 124: is this displaying all markers for thousands of stores, or is results already filtered in some way to limit the number.

HashEngineering (approved):

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
Syn-McJ: Yes, internally we should keep everything in metric because it's used in most of map and location APIs. The switch should only happen on the higher level (the ViewModel) to avoid doing conversions all over the place. Right now switching isn't implemented yet and the user sees miles.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
Syn-McJ: The results are limited by `min(map bounds, chosen radius)`. It still means there can be a couple of thousand points shown in big cities when you zoom out. I have left a suggestion to limit this further in NMA-785.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
```diff
@@ -0,0 +1,340 @@
...
+        })
+    }
+
+    private fun handleClickOnMerchantMarker(results: List<SearchResult>) {
+        markerCollection?.setOnMarkerClickListener { marker ->
+            if (marker.tag == CURRENT_POSITION_MARKER_TAG) {
+                false
+            } else {
+                // TODO: this can be moved to the viewModel, which will allow to write a test for it.
+                // Also, it might be better to set Id or Tag of the marker to the Id
```
ClaudeHangui, line 184: I agree with using tags to handle clicks on markers

ClaudeHangui (approved):
Good for me

PR #805, opened 2021-11-08
fix: update locked state when activity is created and app is locked

HashEngineering:

The fix there it to ensure that LockScreenActivity has the correct state in the `isLocked` field.

Then MainActivity can use this field that will have the correct value to determine how to handle an invite.
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):
Looks good

PR #813, opened 2021-11-12
feat: use local.properties for support email (NMA-1071)

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/.github/workflows/dashwallet.yml
+++ b/.github/workflows/dashwallet.yml
@@ -76,12 +76,9 @@ jobs:
       
     - name: TestNet Build and Firebase Distribution
       if: github.event_name == 'push'
-      env:
-        SUPPORT_EMAIL: "${{ secrets.INTERNAL_SUPPORT_EMAIL }}"
-      run: bundle exec fastlane build_distribute flavor:"_testNet3" type:"release" store_pass:"${{ secrets.SIGNING_STORE_PASS }}" version_code:"${{ env.build_number }}" comment:"Up to date Dash Wallet TestNet build" app_id:"1:1039839682638:android:3202b16d460a1a88" test_group:"qa"
-      
+      run: bundle exec fastlane build_distribute flavor:"_testNet3" type:"release" store_pass:"${{ secrets.SIGNING_STORE_PASS }}" version_code:"${{ env.build_number }}" comment:"Up to date Dash Wallet TestNet build" app_id:"1:1039839682638:android:3202b16d460a1a88" test_group:"qa" SUPPORT_EMAIL:"${{ secrets.INTERNAL_SUPPORT_EMAIL }}"
+
     - name: Production Build and Firebase Distribution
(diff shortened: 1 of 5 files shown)


.github/workflows/dashwallet.yml
```diff
@@ -76,8 +76,12 @@ jobs:
...
     - name: TestNet Build and Firebase Distribution
       if: github.event_name == 'push'
+      env:
+        SUPPORT_EMAIL: "${{ secrets.INTERNAL_SUPPORT_EMAIL }}"
       run: bundle exec fastlane build_distribute flavor:"_testNet3" type:"release" store_pass:"${{ secrets.SIGNING_STORE_PASS }}" version_code:"${{ env.build_number }}" comment:"Up to date Dash Wallet TestNet build" app_id:"1:1039839682638:android:3202b16d460a1a88" test_group:"qa"
       
     - name: Production Build and Firebase Distribution
       if: github.event_name == 'push'
+      env:
+        SUPPORT_EMAIL: "${{ secrets.INTERNAL_SUPPORT_EMAIL }}"
```
Syn-McJ, line 86: `env` can be specified just once. See `run_num` in manual_distribution.yml

HashEngineering: Thanks, I will do this, since this method is working.

.github/workflows/dashwallet.yml
```diff
@@ -15,7 +15,11 @@ jobs:
     - name: Get build number from run id
       run: |
           echo "build_number=$((70000))" >> $GITHUB_ENV
-    
+
+    - name: Get the support email
+      env:
+        SUPPORT_EMAIL: "${{ secrets.INTERNAL_SUPPORT_EMAIL }}"
```
HashEngineering, line 21: This was found to work, but should I use the same method as line 17 to get an environment variable "registered"

.github/workflows/manual_distribution.yml
```diff
@@ -47,6 +47,10 @@ jobs:
           run_num: ${{ github.run_number }}
       run: |
           echo "build_number=$((70000+$run_num))" >> $GITHUB_ENV
+
+    - name: Get the support email
+      env:
+        SUPPORT_EMAIL: "${{ secrets.INTERNAL_SUPPORT_EMAIL }}"
```
HashEngineering, line 53: same question as above

HashEngineering (commented):
@Syn-McJ I have a few questions

.github/workflows/dashwallet.yml
Syn-McJ: @HashEngineering I don't think so, I was using echo to calculate build number (offset + run_num), so it might actually not be needed anymore here. The way you set env variable here should be fine.

Syn-McJ (approved):
Looks good, `echo` is probably not needed (see my answer).

PR #803, opened 2021-11-05
DashPay: Support Core 0.18 (NMA-1025)

HashEngineering:
## Issue being fixed or feature implemented

Code in this PR for the most part is removing an obsolete reference in DashJ.
## Related PR's and Dependencies

see: 
https://github.com/dashevo/dashj/pull/144
https://github.com/dashevo/android-dpp/pull/29
https://github.com/dashevo/dapi-client-android/pull/34
https://github.com/dashevo/android-dashpay/pull/56
https://github.com/dashevo/android-dpp/pull/26
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #815, opened 2021-11-15
fix: propagate Context into current thread during invite creation (NMA-1090)

HashEngineering:
This will fix invitation creation bugs

Fix rate cases of invitation creation failures

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
@@ -80,13 +80,13 @@ class SendInviteWorker @AssistedInject constructor(
     }
 
     private val platformRepo = PlatformRepo.getInstance()
+    private val wallet = platformRepo.walletApplication.wallet!!
 
     override suspend fun doWorkWithBaseProgress(): Result {
         val password = inputData.getString(KEY_PASSWORD)
                 ?: return Result.failure(workDataOf(KEY_ERROR_MESSAGE to "missing KEY_PASSWORD parameter"))
 
         val encryptionKey: KeyParameter
-        val wallet = WalletApplication.getInstance().wallet!!
         org.bitcoinj.core.Context.propagate(wallet.context)
         try {
             encryptionKey = wallet.keyCrypter!!.deriveKey(password)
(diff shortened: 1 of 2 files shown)


wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
```diff
@@ -122,6 +122,9 @@ class SendInviteWorker @AssistedInject constructor(
 
     private fun createDynamicLink(dashPayProfile: DashPayProfile, cftx: CreditFundingTransaction, aesKeyParameter: KeyParameter): DynamicLink {
         log.info("creating dynamic link for invitation")
+        // dashj Context does not work with coroutines well, so we need to call Context.propogate
+        // in each suspend method that uses the dashj Context
+        org.bitcoinj.core.Context.propagate(wallet.context)
```
Syn-McJ, line 127: The application context is injected in the constructor with Hilt already, maybe you can use that instead?

Syn-McJ: Alternatively, you can inject the entire wallet object. It will require providing it in a module, something like this:
```
@Module
@InstallIn(SingletonComponent::class)
object ConfigurationModule {
    @Provides
    fun provideWallet(): WalletApplication = WalletApplication.getInstance()

    @Provides
    fun provideConfiguration(@ApplicationContext context: Context): Configuration =
        Configuration(PreferenceManager.getDefaultSharedPreferences(context), context.resources)
}
```

Syn-McJ (approved):
Looks good, I noted a possible use of DI here

wallet/src/de/schildbach/wallet/ui/dashpay/work/SendInviteWorker.kt
HashEngineering: > The application context is injected in the constructor with Hilt already, maybe you can use that instead?

The Context here is the DashJ context object.  While the code uses `wallet.context` to obtain this context, it can also be obtained with `Context.get()`

For now I would like to leave this PR the way it is and consider using DI starting with the above example as part of PR #803 or another task.

PR #834, opened 2021-12-10
[NMA-1110] (Explore Dash) Distance from user

Syn-McJ:
We want to show distance from user to a merchant/atm if location services are enabled.

Calculate distance for all items if location services are enabled and display it in UI.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/extensions/LocaleExt.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.dash.wallet.features.exploredash.ui.extensions
+
+import java.util.*
+
+val Locale.isMetric: Boolean
+    get() {
+        return !isO3Country.equals("usa", true) &&
+               !isO3Country.equals("mmr", true)
```
HashEngineering, line 25: I was curious about this list of non-metric countries.  When I was in England, I noticed that they are still using miles on distance on roads, though most of their measurements were metric.

image
image
https://en.wikipedia.org/wiki/Road_speed_limits_in_the_United_Kingdom

Also, Liberia is noted as not using the metric system in some older articles.

Syn-McJ: Yes, I know of Britain's case - it seems to be very mixed there, officially metric but uses imperial a lot. I assumed it's acceptable to set any option for them as a default, but it seems to be that road distances are mostly imperial still. We can add them to this list until we see any complaints.
As for Liberia - yes, seems like it's also in the list of "special" countries that use imperial exclusively, I missed that. Will fix.

Syn-McJ: Fixed

ClaudeHangui (approved):
Looks alright for me !

HashEngineering (approved):

PR #835, opened 2021-12-10
feat: search nearby locations

ClaudeHangui:
This task is all about displaying the list of nearby locations on the map. Most of the AC's required have already been done in a previous story

- hide the map result if user filter by state
- provide the ability to the user to re-center the map (via a button) to their current location if they pan the map

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -615,6 +616,17 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
+        if (viewModel.isLocationEnabled.value == true &&
+            viewModel.filterMode.value != FilterMode.Online && filters.territory.isEmpty()){
+            bottomSheet.isDraggable = true
+            bottomSheet.state = BottomSheetBehavior.STATE_HALF_EXPANDED
+            bottomSheetWasExpanded = false
+        } else {
+            bottomSheet.isDraggable = false
+            bottomSheet.state = BottomSheetBehavior.STATE_EXPANDED
+            bottomSheetWasExpanded = true
+        }
```
Syn-McJ, line 629: Unfortunately, this creates some problems. First of all, `resolveAppliedFilters` method is only for setting up the bottom panel that shows what filters are applied, it shouldn't contain any other logic (maybe we should rename this method if its purpose isn't clear).

Second, there are numerous other places in the SearchFragment that change `isDraggable` and `state` and those places will ignore `territory` and this snippet completely.

The way I think this should be set up is we should have a single method that checks all conditions for the bottom sheet draggable status (similar to the `shouldShowFiltersPanel` method). Then, all assignments like `bottomSheet.isDraggable =` should use this method instead of a direct assignment. Conditions that should be checked are:
- is location enabled
- is online filter mode
- screen state

Similar method can be added for the `bottomSheet.state`. There, apart from the above-mentioned conditions merchant type should also be checked, since the sheet is always expanded for Online merchants.

Changes in the territory can continue to be observed through `appliedFilters`, or you can add a livedata for territory and observe it directly.

features/exploredash/src/main/res/drawable/round_corners_bgd_white.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
+    <corners android:radius="8dp"/>
+    <solid android:color="@color/dash_white"/>
+</shape>
```
Syn-McJ, line 5: This resource already exists in the wallet module, perhaps it's better to move it to the common instead of duplicating. Or you can use `rounded_background` with a theme, the purpose of `rounded_background` is to avoid creating multiple rounded background resources but instead use just one with different themes.

features/exploredash/src/main/res/drawable/ic_compass.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="20dp"
+    android:viewportHeight="20" android:viewportWidth="20"
+    android:width="20dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="#008DE4" android:pathData="M10.0007,6.6666C8.159,6.6666 6.6673,8.1582 6.6673,9.9999C6.6673,11.8416 8.159,13.3333 10.0007,13.3333C11.8423,13.3333 13.334,11.8416 13.334,9.9999C13.334,8.1582 11.8423,6.6666 10.0007,6.6666ZM17.4506,9.1666C17.0673,5.6916 14.309,2.9332 10.834,2.5499V0.8333H9.1673V2.5499C5.6923,2.9332 2.934,5.6916 2.5506,9.1666H0.834V10.8333H2.5506C2.934,14.3083 5.6923,17.0666 9.1673,17.4499V19.1666H10.834V17.4499C14.309,17.0666 17.0673,14.3083 17.4506,10.8333H19.1673V9.1666H17.4506ZM10.0007,15.8333C6.7757,15.8333 4.1673,13.2249 4.1673,9.9999C4.1673,6.7749 6.7757,4.1666 10.0007,4.1666C13.2256,4.1666 15.834,6.7749 15.834,9.9999C15.834,13.2249 13.2256,15.8333 10.0007,15.8333Z"/>
+</vector>
```
Syn-McJ, line 5: This icon is already added as `ic_current_location`, you can reuse that instead by applying a blue tint to it.

features/exploredash/src/main/res/layout/fragment_search.xml
```diff
@@ -58,7 +58,19 @@
             android:layout_height="match_parent"
             android:layout_marginTop="?attr/actionBarSize"
             app:layout_behavior="org.dash.wallet.features.exploredash.ui.bottomsheetbehavior.ExploreMapBottomSheetBehavior" />
-
+        <ImageButton
+            android:id="@+id/recenter_map_btn"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:background="@drawable/round_corners_bgd_white"
+            android:src="@drawable/ic_compass"
```
Syn-McJ, line 66: Can reuse `ic_current_location` instead with `app:tint`

Syn-McJ (changes requested):
This breaks behavior of the bottom sheet somewhat, please review the requested changes.

common/src/main/java/org/dash/wallet/common/ui/OffsetDialogFragment.kt
```diff
@@ -23,22 +23,20 @@ import android.view.ViewGroup
...
-import androidx.fragment.app.activityViewModels
+import androidx.lifecycle.ViewModelProvider
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import com.google.android.material.bottomsheet.BottomSheetDialog
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment
-import dagger.hilt.android.AndroidEntryPoint
 import org.dash.wallet.common.R
 import org.dash.wallet.common.UserInteractionAwareCallback
 
-@AndroidEntryPoint
```
ClaudeHangui: @AndroidEntryPoint-annotated classes cannot have type parameters
So we had to get the `LockScreenViewModel` using the provider

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dialogs/FiltersDialog.kt
```diff
@@ -287,7 +287,7 @@ class FiltersDialog: OffsetDialogFragment<ConstraintLayout>() {
     }
 
     private fun applyFilters() {
-        viewModel.selectedTerritory = selectedTerritory
```
ClaudeHangui, line 290: had to define a custom setter for the territory

ClaudeHangui:
@Syn-McJ I took upon your suggestion to update my work:

- a method to define the state of the bottom sheet
- a method to define whether the bottom sheet should be draggable or not
- I made the choice of observing the territory on its own rather than relying `appliedFilters` .

Syn-McJ (approved):
Changes look good

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -101,11 +101,10 @@ class ExploreViewModel @Inject constructor(
     val currentUserLocation = _currentUserLocation.asLiveData()
 
     private val _selectedTerritory = MutableStateFlow("")
-    var selectedTerritory: String
```
ClaudeHangui, line 104: As we changed this to a Livedata, as such, since it's value is updated in `FiltersDialog` class, we needed to define a custom setter

HashEngineering (approved):
Looks good.

PR #837, opened 2021-12-16
bugfix (Explore Dash): expanded bottom sheet in details mode

Syn-McJ:
There is a problem I missed with https://github.com/dashevo/dash-wallet/pull/835: if we select a territory, the bottom sheet stays expanded for physical locations. This shouldn't be the case, we should always show the map for a point with coordinates (unless opened from the Online tab).

Changed the `setBottomSheetState` method and simplified conditions check to make it more clear.

The logic that we have for the bottom sheet and some other components in explore-dash got quite complicated and not too intuitive, as evident from difficulties other developers have while changing it. This is the result of having multiple different logical screens that live in a single physical one. I have some ideas for refactoring this to make it less convoluted, but for now, we just need a fix for the existing problem.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
+++ b/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
@@ -400,7 +400,8 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
 
     private fun transitToDetails() {
         val item = viewModel.selectedItem.value ?: return
-        val isOnline = item.type == MerchantType.ONLINE
+        val isOnline = viewModel.filterMode.value == FilterMode.Online ||
+                item.type == MerchantType.ONLINE
 
         val bottomSheet = BottomSheetBehavior.from(binding.contentPanel)
         bottomSheetWasExpanded = bottomSheet.state == BottomSheetBehavior.STATE_EXPANDED
@@ -432,7 +433,7 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
         val bottomSheet = BottomSheetBehavior.from(binding.contentPanel)
         bottomSheet.isDraggable = isBottomSheetDraggable(viewModel.isLocationEnabled.value == true)
         bottomSheet.expandedOffset = resources.getDimensionPixelOffset(R.dimen.default_expanded_offset)
-        bottomSheet.state = setBottomSheetState(expandedInSearchScreen = bottomSheetWasExpanded)
+        bottomSheet.state = setBottomSheetState(bottomSheetWasExpanded)
 
         if (binding.searchResults.itemDecorationCount < 1) {
             binding.searchResults.addItemDecoration(searchResultsDecorator)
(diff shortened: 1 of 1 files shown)


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -400,7 +400,8 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
 
     private fun transitToDetails() {
         val item = viewModel.selectedItem.value ?: return
-        val isOnline = item.type == MerchantType.ONLINE
+        val isOnline = viewModel.filterMode.value == FilterMode.Online ||
+                item.type == MerchantType.ONLINE
```
Syn-McJ, line 404: When the item is of Both type, we treat it like Online if it is opened from the Online tab.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -637,17 +638,18 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
-                                    expandInMerchantLocationScreen: Boolean = false): Int {
-
-        return if (viewModel.filterMode.value == FilterMode.Online
-            || viewModel.selectedTerritory.value?.isNotEmpty() == true
-            || viewModel.isLocationEnabled.value == false || isOnlineMerchant
-            || expandedInSearchScreen || expandInMerchantLocationScreen){
-            BottomSheetBehavior.STATE_EXPANDED
-        } else {
-            BottomSheetBehavior.STATE_HALF_EXPANDED
+    private fun setBottomSheetState(forceExpand: Boolean = false): Int {
```
Syn-McJ, line 641: No need for multiple forcing conditions.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -637,17 +638,18 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
+        val isDetails = screenState == ScreenState.DetailsGrouped || screenState == ScreenState.Details
+        val nearbySearch = viewModel.filterMode.value != FilterMode.Online &&
+                           viewModel.selectedTerritory.value.isNullOrEmpty() &&
+                           viewModel.isLocationEnabled.value == true
+
+        return when {
+            forceExpand -> BottomSheetBehavior.STATE_EXPANDED
+            isDetails -> BottomSheetBehavior.STATE_HALF_EXPANDED
+            nearbySearch -> BottomSheetBehavior.STATE_HALF_EXPANDED
+            else -> BottomSheetBehavior.STATE_EXPANDED
```
Syn-McJ, line 652: Changed `if` to `when` - hopefully, it makes this check clearer.

ClaudeHangui (approved):
Ok by me
thanks for the simplification of the `setBottomSheetState` method !

HashEngineering (approved):

PR #839, opened 2021-12-16
fix: update map zoom level as per radius filter change

ClaudeHangui:
Any change in the `radius` filter was not reflected zoom level of the map

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
+++ b/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
@@ -107,13 +107,13 @@ class ExploreViewModel @Inject constructor(
 
     // Can be miles or kilometers, see isMetric
     private val _selectedRadiusOption = MutableStateFlow(DEFAULT_RADIUS_OPTION)
-    var selectedRadiusOption: Int
-        get() = _selectedRadiusOption.value
-        set(value) { _selectedRadiusOption.value = value }
-
+    var selectedRadiusOption = _selectedRadiusOption.asLiveData()
+    fun setSelectedRadiusOption(selectedRadius: Int){
+        _selectedRadiusOption.value = selectedRadius
+    }
     // In meters
     val radius: Double
-        get() = if (isMetric) selectedRadiusOption * Const.METERS_IN_KILOMETER else selectedRadiusOption * Const.METERS_IN_MILE
+        get() = if (isMetric) selectedRadiusOption.value!! * Const.METERS_IN_KILOMETER else selectedRadiusOption.value!! * Const.METERS_IN_MILE
 
     // Bounded only by selected radius
     private var radiusBounds: GeoBounds? = null
@@ -220,7 +220,7 @@ class ExploreViewModel @Inject constructor(
                                     }
                                     .flatMapLatest { bounds ->
                                         _appliedFilters.postValue(
-                                            FilterOptions(query, territory, payment, selectedRadiusOption)
+                                            FilterOptions(query, territory, payment, selectedRadiusOption.value!!)
                                         )
                                         getPagingFlow(query, territory, payment, mode, bounds, sortByDistance)
                                             .cachedIn(viewModelScope)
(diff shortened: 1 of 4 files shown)


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -107,13 +107,13 @@ class ExploreViewModel @Inject constructor(
 
     // Can be miles or kilometers, see isMetric
     private val _selectedRadiusOption = MutableStateFlow(DEFAULT_RADIUS_OPTION)
-    var selectedRadiusOption: Int
-        get() = _selectedRadiusOption.value
```
ClaudeHangui, line 111: A custom setter is needed to update the radius and the changes are observed via a Livedata

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -107,13 +107,16 @@ class ExploreViewModel @Inject constructor(
...
-        set(value) { _selectedRadiusOption.value = value }
-
+    val selectedRadiusOption = _selectedRadiusOption.asLiveData()
+    fun setSelectedRadiusOption(selectedRadius: Int){
+        _selectedRadiusOption.value = selectedRadius
+    }
     // In meters
     val radius: Double
-        get() = if (isMetric) selectedRadiusOption * Const.METERS_IN_KILOMETER else selectedRadiusOption * Const.METERS_IN_MILE
```
ClaudeHangui, line 116: Test cases were failing due to the value of `radius` not yet updated, since it depends on `selectedRadiusOption` which is now a livedata. We therefore needed to provide a default radius value

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -220,7 +223,7 @@ class ExploreViewModel @Inject constructor(
                                     }
                                     .flatMapLatest { bounds ->
                                         _appliedFilters.postValue(
-                                            FilterOptions(query, territory, payment, selectedRadiusOption)
+                                            FilterOptions(query, territory, payment, selectedRadiusOption.value!!)
```
Syn-McJ, line 226: `radius` getter has a null check for the `selectedRadiusOption`, but here it's just assumed that it's not null. Which is likely, but not guaranteed.

It's better to use the non-null value that you already have access to in the flatMap (just need to declare it):
```
_selectedRadiusOption.flatMapLatest { selectedRadius ->
```

This one is guaranteed to be non-null.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -534,11 +534,13 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
         val searchLocation = if (viewModel.selectedTerritory.value?.isNotEmpty() == true) {
             viewModel.selectedTerritory.value
         } else {
-            val radiusOption = viewModel.selectedRadiusOption
-            resources.getQuantityString(
-                if (viewModel.isMetric) R.plurals.radius_kilometers else R.plurals.radius_miles,
-                radiusOption, radiusOption
-            )
+            val radiusOption = viewModel.selectedRadiusOption.value
+            radiusOption?.let {
```
Syn-McJ, line 538: So the `searchLocation` is nullable now and we're calling `getString` with null potentially? This is wrong, need to use the default radius value if `radiusOption` is null.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dialogs/FiltersDialog.kt
```diff
@@ -186,7 +186,7 @@ class FiltersDialog: OffsetDialogFragment<ConstraintLayout>() {
         binding.radiusLabel.isVisible = true
         binding.radiusCard.isVisible = true
 
-        selectedRadiusOption = viewModel.selectedRadiusOption
+        selectedRadiusOption = viewModel.selectedRadiusOption.value!!
```
Syn-McJ, line 189: See the comment above. Here, we should use default radius value if `selectedRadiusOption` is null.

Syn-McJ (changes requested):
@ClaudeHangui looks good but some minor changes needed with null-value handling.

ClaudeHangui:
@Syn-McJ please check out my latest changes

Syn-McJ:
> @Syn-McJ please check out my latest changes

@ClaudeHangui I don't see any changes?

ClaudeHangui:
@Syn-McJ  snap, I forgot to push.
it's done now sorry

Syn-McJ (approved):
Looks great

HashEngineering (approved):
Looks great.

PR #840, opened 2021-12-16
[NMA-1115] (Explore Dash) Possible fix for sync worker issues

Syn-McJ:
Sync worker fails to run sometimes with a firebase error “client is offline” even though the connection is present.

It's not 100% clear what is the problem, but it might be related to the usage of `await()` method from the `kotlinx-coroutines-play-services`, even though this issue was apparently [fixed a while ago](https://github.com/firebase/firebase-android-sdk/issues/2385). I moved all of the firebase data fetching methods from `await()` back to `addListenerForSingleValueEvent` just in case this is the problem.

Another possibility is that this is simply due to connection problems, hence I've added an exponential backoff retrying policy.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks good.

PR #844, opened 2021-12-23
NMA-1134 : fix the local currency

hadia:
Fix the Local currency format to match figma 
image

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/adapter/BuyAndSellDashServicesAdapter.kt
+++ b/wallet/src/de/schildbach/wallet/adapter/BuyAndSellDashServicesAdapter.kt
@@ -11,8 +11,6 @@ import androidx.core.view.isVisible
 import de.schildbach.wallet.data.BuyAndSellDashServicesModel
 import de.schildbach.wallet_test.databinding.ItemServiceListBinding
 import org.bitcoinj.core.Coin
-import org.bitcoinj.utils.ExchangeRate
-import org.bitcoinj.utils.MonetaryFormat
 import org.dash.wallet.common.Configuration
 import org.dash.wallet.common.Constants
 import org.dash.wallet.common.ui.BaseAdapter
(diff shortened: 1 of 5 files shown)


wallet/src/de/schildbach/wallet/adapter/BuyAndSellDashServicesAdapter.kt
```diff
@@ -40,16 +46,26 @@ class BuyAndSellDashServicesAdapter( val config: Configuration,val onClickListen
...
                 )
+
                 if(data.balance!=null){
                     binding.serviceBalance.setAmount(data.balance)
                 }
                 if(data.localBalance!=null){
-                    binding.serviceFiatAmount.setAmount(data.localBalance)
+                    binding.serviceFiatAmount.text = Constants.PREFIX_ALMOST_EQUAL_TO+  view.context.getString(
+                        R.string.fiat_balance_with_currency, fiatFormat.format(data.localBalance),
+                        currencySymbol
```
Syn-McJ, line 67: Whether a currency symbol goes before or after an amount depends on locale of the device, but here it seems like it is always going to be after the amount.

Syn-McJ (changes requested):

Syn-McJ (commented):
See my comment

wallet/src/de/schildbach/wallet/adapter/BuyAndSellDashServicesAdapter.kt
Syn-McJ: Actually, I'm not sure if this is required given that we ignore this everywhere in the app except for the Enter Amount screens. @HashEngineering

hadia: In the figma design it's before the amount

Syn-McJ: Figma designs don't represent all the possible states of UI. Sometimes it has currency before the amount, sometimes after. Generally, it should depend on the device locale. NumberFormat class handles this automatically, but we probably need to use MonetaryFormat which doesn't.

Syn-McJ (approved):
Great work

HashEngineering (approved):

PR #848, opened 2021-12-25
[NMA-1111] (Coinbase UI) Payment method picker + Enter Amount

Syn-McJ:
As a part of Coinbase integration, we need a way to pick a payment method out of all provided by Coinbase API and enter the amount to buy in Dash or a chosen fiat. 

- Implemented a payment method picker as a separate component that can be reused in other integration modules.
- Implemented an enter amount + currency picker component that can be reused in other modules.
- Added a way to inject the ExchangeRatesRepository with Hilt.

`AmountView` and `EnterAmountFragment` borrow from the old `EnterAmountFragment` in the wallet, but I've tried to simplify it since the old one has a lot of repeating code and nested checks that are unnecessary. It's likely that I've missed some things that were considered in the old implementation, in which case we have time to ride the new one out in Coinbase and other modules and fix the bugs if there are any.

`WalletDataProvider` could be provided and injected with Hilt for fetching the exchange rates, but at this point, I figured we might as well inject `ExchangeRatesRepository` directly. The `ExchangeRatesRepository` doesn't need to be moved to the `common` module, only an interface is needed and the implementation can be provided with Hilt out of the `wallet`. 

`ExchangeRate` did need to be moved to the `common` however, at which point I decided to convert it to Kotlin for a better null-safety. The database schema wasn't changed, so this will not affect users in any way.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/data/ExchangeRate.kt
```diff
@@ -1,10 +1,115 @@
...
+        dest.writeString(currencyName)
+        dest.writeSerializable(currency)
+    }
+
+    companion object CREATOR : Parcelable.Creator<ExchangeRate> {
+        override fun createFromParcel(parcel: Parcel): ExchangeRate {
+            return ExchangeRate(parcel)
+        }
 
-data class ExchangeRate(
```
Syn-McJ, line 5: The old `ExchangeRate` class is renamed to `ExchangeRateData` and marked as deprecated since it's just a wrapper there is no need for it anymore.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/NumericKeyboardView.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    constructor(context: Context, attrs: AttributeSet, defStyleAttr: Int) : super(context, attrs) {
+        init(attrs, defStyleAttr, defStyleAttr)
+    }
+
+    private fun init(attributeSet: AttributeSet, defStyleAttr: Int, defStyleRes: Int) {
+        context.withStyledAttributes(attributeSet, R.styleable.NumericKeyboardView, defStyleAttr, defStyleRes) {
+            isFunctionEnabled = getBoolean(R.styleable.NumericKeyboardView_nk_functionEnabled, isFunctionEnabled)
+            isDecSeparatorEnabled = getBoolean(R.styleable.NumericKeyboardView_nk_decSeparatorEnabled, isDecSeparatorEnabled)
```
Syn-McJ, line 77: I figured it might be useful to set `isFunctionEnabled` and `isDecSeparatorEnabled` from the layout

common/src/main/res/color/radiobutton_icon_color.xml
```diff
@@ -14,14 +14,7 @@
...
-        <shape android:shape="rectangle">
-            <corners android:radius="?attr/cornerRadius" tools:radius="10dp" />
-            <solid android:color="@android:color/white" />
-        </shape>
-    </item>
-</ripple>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="@color/white" android:state_selected="true"/>
+    <item android:color="@color/steel_gray_250" android:state_selected="false"/>
+</selector>
```
Syn-McJ, line 20: This diff doesn't make sense, `rounded_ripple_transparent` was deleted since `rounded_ripple_background` can do the same thing.

common/src/main/res/drawable/bg_dash_gradient.xml
```diff
@@ -1,21 +1,7 @@
...
-    </aapt:attr>
-  </path>
-</vector>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+  <gradient
+      android:type="linear"
+      android:angle="0"
+      android:startColor="#FF3199DF"
+      android:endColor="#FF006CBD" />
+</shape>
```
Syn-McJ, line 7: `vector` gradient caused crash on API 23

common/src/main/res/layout/app_bar_with_icon.xml
```diff
@@ -43,8 +41,6 @@
         <TextView
             android:id="@+id/connected"
             style="@style/Overline"
-            android:textSize="10sp"
-            android:textColor="@color/gray_900"
```
Syn-McJ, line 47: The style already has the correct settings for text size and color.

common/src/main/res/values/styles.xml
```diff
@@ -862,13 +931,125 @@
...
+    <style name="TertiaryButtonTheme.Small.Red">
+        <item name="titleTextColor">@color/red_300</item>
+    </style>
+
+    <style name="TertiaryButtonTheme.Large">
+        <item name="cornerRadius">10dp</item>
+    </style>
+
+    <style name="TertiaryButtonTheme.Large.Red">
+        <item name="titleTextColor">@color/red_300</item>
```
Syn-McJ, line 1046: These changes might cause some merge conflicts with explore-dash. I'll resolve them after one of the branches merged with the `master`.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseServicesViewModel.kt
```diff
@@ -16,25 +16,25 @@
...
 import javax.inject.Inject
 
 @HiltViewModel
 class CoinbaseServicesViewModel @Inject constructor(
-    application: Application,
     private val coinBaseRepository: CoinBaseRepository,
+    private val exchangeRatesProvider: ExchangeRatesProvider,
     val config: Configuration
-) : AndroidViewModel(application) {
+) : ViewModel() {
```
Syn-McJ, line 37: We don't actually need a viewmodel with application. Context can be injected with Hilt if needed.

wallet/src/de/schildbach/wallet/BaseWalletApplication.kt
```diff
@@ -60,29 +60,19 @@ abstract class BaseWalletApplication : MultiDexApplication(), WalletDataProvider
...
-            return@switchMap MutableLiveData<List<ExchangeRate>>().apply {
+            return@switchMap MutableLiveData<List<ExchangeRateData>>().apply {
                 value = it.map { exchangeRate ->
-                    ExchangeRate(exchangeRate.currencyCode, exchangeRate.rate, exchangeRate.getCurrencyName(this@BaseWalletApplication), exchangeRate.fiat)
-                }
-            }
-        }
-    }
-
-    override fun currencyCodes(): LiveData<List<String>> {
```
Syn-McJ, line 81: `currencyCodes()` method wasn't used. If needed, it's better to re-add as `Flow<List<String>>`

wallet/src/de/schildbach/wallet/ui/LockScreenActivity.kt
```diff
@@ -512,5 +515,19 @@ open class LockScreenActivity : SecureActivity() {
         }
         bottomSheetDialog?.dismiss()
         dialogFragment?.dismiss()
+
+        dismissOffsetDialogs(supportFragmentManager)
+    }
+
+    private fun dismissOffsetDialogs(fragmentManager: FragmentManager) {
```
Syn-McJ, line 522: There is another problem with dismissing dialogs currently - if the user unlocked the app and then opened a dialog, `activatingLockScreen` can be triggered after initialization, which will lead to that dialog being dismissed immediately.

I have added an alternative method for the `OffsetDialogFragment` exclusively to avoid side effects, but we should consider dismissing all dialogs this way.

wallet/schemas/de.schildbach.wallet.AppDatabase/2.json
```diff
@@ -1,102 +0,0 @@
-{
-  "formatVersion": 1,
-  "database": {
-    "version": 2,
-    "identityHash": "77eb849e83b9d0cc8bb55a8b6f0b9d19",
```
HashEngineering, line 5: Why is this file deleted?

I thought we had to keep these files.

.gitignore
```diff
@@ -29,4 +29,5 @@ vendor/
 fastlane/report.xml
 google-services.json
 output-metadata.json
-service.properties
\ No newline at end of file
+service.properties
+wallet/schemas/*/*.json
```
HashEngineering, line 33: So we don't need these files in our repo?

I thought these files were needed for backward compatibility with older database versions.

HashEngineering (commented):
I have questions regarding the room schema .json files

.gitignore
Syn-McJ: These files are actually generated on build. We don't need to keep them, and in some other branches, they were already excluded in .gitignore.

Right now our database is actually performing a destructive migration on every database version increment. That means everything is erased and re-created again. For keeping data when updating the database version, we need to write migrations.

Syn-McJ: Re-added since we still might need them for testing migrations.

HashEngineering (approved):
Approved since my concerns were addressed.

PR #851, opened 2021-12-27
[NMA-1105] Currency selector problems

Syn-McJ:
After making a payment with non-default currency, opening the currency selector again shows the wrong currency.

The way I see it, the problem here is that we have a bunch of code in the `ExchangeRatesFragment` to decide which currency to show as selected. Because this code is independent of the one in the Enter Amount screen, it can get unsynchronized and we will have different currencies.

To prevent this problem from occurring again, we should move the responsibility of deciding selected currency to the caller of `ExchangeRatesFragment`, it itself should not make this decision unless no currency is specified. This way, whichever currency the caller has as default, `ExchangeRatesFragment` will have as selected.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


ClaudeHangui (approved):
Ok by me!

HashEngineering (approved):
Looks great.  21 lines added and removed, and the ExchangeRatesFragment code is more simple.

PR #855, opened 2021-12-29
[NMA-1100] the local currency is not changed in the first time after installation

Syn-McJ:
If a user in the US tries to change the default currency immediately after the installation, it will get reset back to USD after exiting the settings. This happens because `onResume` method is called in `MainActivity` and it triggers country detection again (this only happens in DashPay for some reason).

`detectUserCountry` will not run if `exchangeCurrencyCodeDetected` is set. This property is used to detect unintended changes in the currency and show a dialog asking which one to keep. We can set it preemptively after the user changed the default currency, assuming that if it's changed explicitly, we don't need to confirm it with a dialog.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Approved.

PR #858, opened 2022-01-03
[NMA-1144] (Explore Dash) Distance for online-only merchants & sync fixes

Syn-McJ:
Online merchants should not have a distance showing up in the search results, and should not appear in the Nearby tab if the location is disabled.
Also, there were some issues with synchronization that lead to `last_updated` being saved locally even if some of the data sources had failed to synchronize.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Look good

PR #859, opened 2022-01-04
[NMA-1145] (Explore Dash) Clearing results on switching filter tab

Syn-McJ:
When the user switches the filter tab, e.g. from All to Online, the results of the All tab are still showing for a short while before Online results are loaded. This might introduce some confusion.

Clearing search results immediately after switching a tab.

Additionally, I've changed `ExistingWorkPolicy` to `Keep` for the explore sync worker so that the ongoing work doesn't get interrupted (which seems to be a cause of failed synchronization in some cases).

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
looks good

PR #861, opened 2022-01-04
[NMA-1141 && NMA-1139 && NMA-1153] Explore Dash fixes

Syn-McJ:
We want to show a full details card for `both` merchants even if opened from the Online tab.
Additionally, when details are opened and then the user goes back to the search screen, the search query is erased but search results are still filtered.
The last fix is grouping in the Online tab, which was by name but should be by the distance.

- Removed all conditions that limit the full details card from opening in `FilterMode.Online` and only left checks for `MerchantType.Online`.
- Now setting search text explicitly when transitioning to the SearchResults screen mode.
- Changed grouping in Online tab to be by a Pythagorean formula like in all the other cases.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks good.

HashEngineering (approved):
New changes are fine.

PR #865, opened 2022-01-07
[NMA-1157] Last sync times in about screen

Syn-McJ:
We want to show the time of the last data load and last sync in the About section. I have also added logging of non-fatal crashes so that we can see sync problems in Firebase immediately.
Additionally, I've updated most of the strings with 2022 copyright year and added manufacturer as a field for searching over ATMs as we wanted but didn't do for the last build.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks Good

PR #867, opened 2022-01-11
fix: use context parameter instead of calling requireContext

HashEngineering:
There is a crash if CheckPinDialog is created during a wait period from entering the incorrect PIN.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):
Looks good

ClaudeHangui (approved):
Good !

PR #868, opened 2022-01-11
fix: missing wallet (NMA-1119)

HashEngineering:

We don't know why some installations have missing wallets.  Here we will put a dialog to warn the user and ask them to submit a report to the Dash Support Desk.

Further investigation will be needed to find the cause of a missing wallet, which may be due to an error with creating a new wallet, restoring a recovery phrase or restoring a file.

N/A
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


common/src/main/res/drawable/ic_error.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
HashEngineering, line 1: This file was taken from `dashpay`.

wallet/src/de/schildbach/wallet/rates/ExchangeRatesRepository.kt
```diff
@@ -109,7 +109,7 @@ class ExchangeRatesRepository private constructor() {
     }
 
     // This will return null if not found
-    fun getExchangeRate(currencyCode: String): ExchangeRate? {
+    fun getExchangeRate(currencyCode: String?): ExchangeRate? {
```
HashEngineering, line 112: it is possible for the currencyCode to be null and this was causing some crashes. 

This class was translated from Java and currencyCode was not specified as `@NotNullable`

wallet/src/de/schildbach/wallet/ui/SetPinActivity.kt
```diff
@@ -115,6 +117,23 @@ class SetPinActivity : InteractionAwareActivity() {
...
+            val fancyAlertDialogViewModel = ViewModelProvider(this)[FancyAlertDialogViewModel::class.java]
+            fancyAlertDialogViewModel.onPositiveButtonClick.observe(this) {
+                alertDialog = ReportIssueDialogBuilder.createReportIssueDialog(this,
+                    WalletApplication.getInstance()).buildAlertDialog()
+                alertDialog.show()
+            }
+            dialog.show(supportFragmentManager, "serious_error_dialog")
+        }
+
         if (walletApplication.wallet.isEncrypted) {
```
HashEngineering, line 137: This will display a dialog that tells the user to contact support.  What if wallet is `null` will the dialog be displayed? or will it close because the app crashes on line 137?

wallet/src/de/schildbach/wallet/ui/send/SendCoinsActivityViewModel.kt
```diff
@@ -37,7 +37,6 @@ open class SendCoinsActivityViewModel(application: Application) : AndroidViewMod
     }
 
     val walletApplication = application as WalletApplication
-    val wallet = walletApplication.wallet!!
```
HashEngineering, line 40: This line was causing a crash if the wallet was `null`.  `wallet` is not used by this view model.

The fact that `wallet` is `null` is a big problem, so removing the crash here will probably mean that another crash will appear elsewhere.

Syn-McJ (commented):
@HashEngineering one comment that might need addressing.

Syn-McJ (approved):
Looks good

PR #870, opened 2022-01-13
[NMA-1160] NPE in fireabase repository

Syn-McJ:
We're getting NPE when using `!!` operator in firebase repository when fetching last_update field and then doing `dataSnapshot.getValue<Long>()!!`. 
Why does last_update returns null in the first place is unclear and should be investigated, but for now we should catch this problem and log it as non-fatal crash.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks great.

PR #871, opened 2022-01-16
[NMA-1163] fix: removing LockScreenViewModel to avoid Hilt errors

Syn-McJ:
LockScreenViewModel and Hilt are causing some problems, likely because the viewModel is registered from different activities that aren't derived from LockScreenViewModel.

- `DialogFragments` are dismissed automatically by enumerating them from the `LockScreenActivity` and calling `dismiss()` directly.
- For `AlertDialogs`, `LockScreenViewModel` is replaced with an analogous singleton service that shouldn't create any problems. This is a temporary measure until we replace `AlertDialogs` with `DialogFragments`, to have a single system of showing/dismissing all dialogs.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/WalletTransactionsFragment.java
```diff
@@ -391,7 +391,7 @@ protected CharSequence collectWalletDump() {
     public void onTransactionRowClicked(Transaction tx) {
         TransactionDetailsDialogFragment transactionDetailsDialogFragment =
                 TransactionDetailsDialogFragment.newInstance(tx.getTxId());
-        transactionDetailsDialogFragment.show(getChildFragmentManager(), null);
+        transactionDetailsDialogFragment.show(getParentFragmentManager(), null);
```
Syn-McJ, line 394: We should show dialogs with `parentFragmentManager` so that it is dismissable from the `LockScreenActivity`.

HashEngineering (approved):
Looks good.

PR #874, opened 2022-01-21
[NMA-1174] Support interruptions when preloading Explore data

tomasz-ludek:
Removed the code related to syncing Explore data with Firebase Realtime Database
Switched from JSON to ProtoByf when preloading Explore data.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ExploreSyncWorker.kt
```diff
@@ -75,91 +71,78 @@ class ExploreSyncWorker constructor(val appContext: Context, workerParams: Worke
...
+                .apply()
 
-        if (synced != dataSize) {
-            throw FirebaseException("sync is broken, expected $dataSize records, got $synced")
+            log.info("Sync Explore Dash finished")
+
+            return@withContext Result.success()
+
+        } catch (ex: Exception) {
+            analytics.logError(ex)
```
Syn-McJ, line 142: It's better if we log some details here, otherwise, it'll be hard to say from crashlytics whether the worker is interrupted just once or several times. I suggest logging something like "syncing from ${merchantDataSizeDB}"

Syn-McJ (changes requested):
Looks good works well, just one slight improvement for logs, please.

wallet/src/de/schildbach/wallet/ExploreSyncWorker.kt
tomasz-ludek: Good point, fixed

Syn-McJ (approved):

HashEngineering:
In a future PR, should we include this in our project (`syncdata.proto`) and then have the Proto source auto generated?

In **DashJ**, we have the `.proto` files and the generated java source as part of the repo.  This is how the BitcoinJ devs do it (or did in the past)

In `dapi-client`, we have only the `.proto` files as part of the repo.

tomasz-ludek:
@HashEngineering the Protos generation is implemented in https://github.com/dashevo/explore-dash-sync (not yet pushed) and is very similar to dashj . Next week I'll push the code and ask you for the review. We will need to find the best approach to sync the proto files between dash-wallet and explore-dash-sync.

HashEngineering (approved):

HashEngineering:
Last commit fixing crash from injection:
https://github.com/dashevo/dash-wallet/pull/875

PR #876, opened 2022-01-23
[NMA-1179] Coinbase common components

Syn-McJ:
For the CrowdNode stories, we need some common components from the Coinbase branch, specifically:
- Enter Amount screen
- Dialogs with the new design

- For the Enter Amount screen, changes largely repeat those in #848 (minus the payment method picker). Additionally, there were two `WalletDataProvider` interfaces serving the same purpose so I combined them in one.
- For the dialogs, I have created an `AdaptiveDialog` that accommodates all of the various new dialogs UIs we have, as well as provides an option to specify a custom layout.

Usage in Kotlin:
```
val dialog = AdaptiveDialog.new(
        R.drawable.ic_location,
        getString(title),
        getString(message),
        getString(R.string.permission_deny),
        getString(R.string.permission_allow)
)

// with callback:
dialog.show(requireActivity()) { result ->
...        
}

// as a coroutine:
val result = dialog.showAsync(requireActivity())
```

In Java:
```
AdaptiveDialog dialog = AdaptiveDialog.Companion.simple("Take this action?", "No", "Yes");
dialog.show(activity, aBoolean -> {
    Log.i(TAG, "AdaptiveDialog result: " + aBoolean);
    return Unit.INSTANCE;
});
```

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
```diff
@@ -22,19 +22,22 @@ import androidx.lifecycle.LiveData
...
 import org.bitcoinj.core.Coin
 import org.bitcoinj.core.Transaction
-import org.dash.wallet.common.data.ExchangeRate
+import org.dash.wallet.common.data.ExchangeRateData
 import org.dash.wallet.common.data.Resource
 
 interface WalletDataProvider {
 
-    fun freshReceiveAddress(): Address
+    fun currentReceiveAddress(): Address
```
ClaudeHangui, line 30: Is there a specific reason for having this function, given that `freshReceiveAddress()` serves the same purpose ?

ClaudeHangui (approved):
Great job !

common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
Syn-McJ: `currentReceiveAddress` is used by Uphold, I just left it as it is when merging two interfaces. But you're right, it might be better if Uphold used `freshReceiveAddress` as Liquid does. Maybe @HashEngineering knows the context.

HashEngineering: We should use `freshReceiveAddress` in Uphold as well.  We forgot to change it previously.

Otherwise it is possible that more than one transaction will be sent to the same address.  Ideally, we want all received transactions to be sent to separate addresses, at least as far as the app can control.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
```diff
@@ -179,4 +183,19 @@ public static String formatFiatWithoutComma(String fiatValue){
...
+        Currency currency = Currency.getInstance(fiat.currencyCode);
+        numberFormat.setCurrency(currency);
+        String currencySymbol = currency.getSymbol(getDeviceLocale());
+        boolean isCurrencyFirst = numberFormat.format(1.0).startsWith(currencySymbol);
+
+        if (isCurrencyFirst) {
+            return currencySymbol + " " + format.format(fiat);
+        } else {
+            return format.format(fiat) + " " + currencySymbol;
+        }
```
HashEngineering, line 199: I wonder if this works correctly with RTL languages.

HashEngineering (changes requested):
There are merge conflicts, otherwise this looks good.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
Syn-McJ: I'll check

Syn-McJ: Well, the result matches what android APIs are producing, f.e. "100 ₫" which then gets reverted to "₫ 100" when assigning this to a `TextView`, which I assume is what needs to be happening.
However, we do have some minor problems with RTL all around the app. We probably should have a story to investigate and fix those.

HashEngineering: That is fine.

PR #878, opened 2022-01-26
chore: coinbase-master merge conflicts

Syn-McJ:
The latest changes to the master with extracting common components from coinbase had created some merge conflicts.
Most of these are minor, however, some changes were required to the following classes:
- `SetPinActivity`
- `CancelTransactionDialog` and other coinbase dialogs (dialogs are now dismissed automatically instead of using `LockScreenViewModel`).

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/enter_amount/AmountView.kt
```diff
@@ -147,8 +154,8 @@ class AmountView(context: Context, attrs: AttributeSet) : ConstraintLayout(conte
     }
 
     private fun updateDashSymbols() {
-        binding.inputCurrencyToggle.isVisible = showCurrencySelector
-        binding.resultCurrencyToggle.isVisible = showCurrencySelector
+        binding.inputCurrencyToggle.isVisible = showCurrencySelector && !dashToFiat
```
ClaudeHangui, line 157: Do we need again to add `&& !dashToFiat` here  given that `showCurrencySelector` already checks that for us ?

ClaudeHangui (approved):
Looks good to me, besides a slight concern regarding my comment on `AmountView` class

common/src/main/java/org/dash/wallet/common/ui/enter_amount/AmountView.kt
Syn-McJ: @ClaudeHangui yes, because `updateDashSymbols` can be called independently from `showCurrencySelector`. Those are dynamic properties that can be changed after `AmountView` initialization, and we can get both toggles hidden or both visible.
And also in general, if we assign visibility to toggles, I feel it should be done in the same way everywhere just to keep the intention clear and avoid confusion for anyone who looks at it in the future.

HashEngineering (approved):

PR #879, opened 2022-01-26
feat: track coinbase buy events

ClaudeHangui:
This PR focuses on tracking buy events related to a Coinbase user

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/enter_amount/AmountView.kt
```diff
@@ -43,6 +46,7 @@ class AmountView(context: Context, attrs: AttributeSet) : ConstraintLayout(conte
         .noCode().minDecimals(2).optionalDecimals()
 
     private var onCurrencyToggleClicked: (() -> Unit)? = null
+    private val analytics = FirebaseAnalyticsServiceImpl.getInstance()
```
Syn-McJ, line 49: @ClaudeHangui it's better to use Hilt for injecting the analytics service. `getInstance()` method was added when we didn't have DI in the wallet.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/AmountView.kt
```diff
@@ -104,6 +108,10 @@ class AmountView(context: Context, attrs: AttributeSet) : ConstraintLayout(conte
         updateCurrency()
         binding.convertDirectionBtn.setOnClickListener {
             dashToFiat = !dashToFiat
+            analytics.logEvent(if (dashToFiat) AnalyticsConstants.Coinbase.ENTER_AMOUNT_DASH
+                else AnalyticsConstants.Coinbase.ENTER_AMOUNT_FIAT,
+                bundleOf()
+            )
```
HashEngineering, line 114: Since this is a common component, then would not this log all currency direction changes as being from the coinbase module?

Syn-McJ: Agree, this is related to my previous comment. The fact that we cannot easily inject AnalyticsService into this component is the first indication that logs are made at the wrong level. DashToFiat should be propagated through the viewModel into the coinbase and logged in there.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseServicesFragment.kt
```diff
@@ -32,14 +36,15 @@ import org.dash.wallet.integration.coinbase_integration.databinding.FragmentCoin
...
+import javax.inject.Inject
 
 @AndroidEntryPoint
 class CoinbaseServicesFragment : Fragment(R.layout.fragment_coinbase_services) {
     private val binding by viewBinding(FragmentCoinbaseServicesBinding::bind)
     private val viewModel by viewModels<CoinbaseServicesViewModel>()
     private var loadingDialog: FancyAlertDialog? = null
     private var currentExchangeRate: org.dash.wallet.common.data.ExchangeRate? = null
-
+    @Inject lateinit var analyticsService: FirebaseAnalyticsServiceImpl
```
HashEngineering, line 47: When injecting, wouldn't it be:
```
     @Inject lateinit var analyticsService: AnalyticsService
```

HashEngineering (changes requested):
one comment left above.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/AmountView.kt
HashEngineering: Should that be the case for all of the component classes?  The Fragment that is using them and responding to events from the component should log the event, rather than the component itself.

Syn-McJ: Yes, that should be the case. Common components should only be responsible for their own functionality and shouldn't know or care about event logging.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountFragment.kt
```diff
@@ -90,6 +95,12 @@ class EnterAmountFragment: Fragment(R.layout.fragment_enter_amount) {
 
         binding.keyboardView.onKeyboardActionListener = keyboardActionListener
         binding.continueBtn.setOnClickListener {
+            if (!binding.amountView.showCurrencySelector) {
+                analyticsService.logEvent(
+                    AnalyticsConstants.Coinbase.CONTINUE_DASH_PURCHASE,
```
Syn-McJ, line 100: @ClaudeHangui this is still wrong. Common components should only be responsible for their own functionality and shouldn't know or care about event logging. Events should be propagated down to the end-user (conbase module in this case) and logged in there. 

Otherwise, we'll end up with a tree of if-else that checks for coinbase vs crowdnode vs etc logs. This breaks both single-responsibility and open-closed principles.

Syn-McJ (changes requested):

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountFragment.kt
ClaudeHangui: you guys are absolutely right !

hadia (approved):

common/src/main/java/org/dash/wallet/common/ui/payment_method_picker/PaymentMethodPicker.kt
```diff
@@ -90,6 +94,7 @@ class PaymentMethodPicker(context: Context, attrs: AttributeSet): ConstraintLayo
                     showSearch = false
                 ) { _, index, dialog ->
                     dialog.dismiss()
+                    analytics.logEvent(AnalyticsConstants.Coinbase.CHANGE_PAYMENT_METHOD, bundleOf())
```
HashEngineering, line 97: Here remains a logged event inside a common component.

HashEngineering (changes requested):

common/src/main/java/org/dash/wallet/common/ui/payment_method_picker/PaymentMethodPicker.kt
ClaudeHangui: Fixed in last push

HashEngineering (approved):
excellent.

HashEngineering (approved):
Excellent, this is ready to merge.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountFragment.kt
HashEngineering: This has been addressed, so I will merge this story.

PR #882, opened 2022-01-28
fix: do not use ProcessPhoenix for reset wallet

HashEngineering:

The goal is to eliminating a Crash when Resetting the Wallet
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/ResetWalletDialog.kt
+++ b/wallet/src/de/schildbach/wallet/ui/ResetWalletDialog.kt
@@ -38,7 +38,12 @@ class ResetWalletDialog : BaseDialogFragment() {
                 negativeAction = {
                     analytics.logEvent(AnalyticsConstants.Security.RESET_WALLET, bundleOf())
                     (activity as? AbstractBindServiceActivity)?.unbindServiceServiceConnection()
+                    // 1. wipe the wallet
+                    // 2. start OnboardingActivity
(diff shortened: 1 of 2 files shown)


wallet/src/de/schildbach/wallet/ui/ResetWalletDialog.kt
```diff
@@ -38,7 +38,12 @@ class ResetWalletDialog : BaseDialogFragment() {
                 negativeAction = {
                     analytics.logEvent(AnalyticsConstants.Security.RESET_WALLET, bundleOf())
                     (activity as? AbstractBindServiceActivity)?.unbindServiceServiceConnection()
+                    // 1. wipe the wallet
+                    // 2. start OnboardingActivity
+                    // 3. close the backstack (Home->More->Security)
                     WalletApplication.getInstance().triggerWipe(context)
+                    startActivity(OnboardingActivity.createIntent(requireContext()))
+                    activity?.finishAffinity()
```
HashEngineering, line 46: I wonder if there might be a race condition here.  Are these statements executed sequentially in the same thread? `triggerWipe` does call into the `BlockchainServiceImpl` -- will these be in the same USER thread?  The `OnboardingActivity`, to work as intended, must find that there is not wallet object or files that are deleted by `triggerWipe`.

HashEngineering:
This was included in PR #885

PR #884, opened 2022-02-01
[NMA-1154] (Explore Dash) Nearby search results shouldn't show if location is disabled

Syn-McJ:
When we don't have location permission, there is no point in the Nearby tab as it's hardly different from the All tab in this case. 

Instead of showing search results, we display a message and a button that initiates a location request flow - the same as in the filters dialog.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks great.

PR #885, opened 2022-02-02
[NMA-1190] fix: Hilt crash in OnboardingActivity

Syn-McJ:
There is a crash in the `OnboardingActivity` on the "Wallet is permanently locked" screen, when the user tries to initiate a reset dialog. It happens because `BaseDialogFragment` is annotated as `@AndroidEntryPoint`, while the `OnboardingActivity` isn't.

A couple of changes are made:
- We could annotate `OnboardingActivity` as `@AndroidEntryPoint`, however, the fact that the `BaseDialogFragment` is annotated creates various side effects around the app (like this bug). It's better to avoid  `@AndroidEntryPoint` in the dialogs, so I have removed it and made related changes.
- Given that the `BaseDialogFragment` is obsolete and unnesasary (`DialogFragments` are dismissed automatically now), I have uncoupled the `ResetWalletDialog` from it.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/ui/BaseDialogFragment.kt
+++ b/common/src/main/java/org/dash/wallet/common/ui/BaseDialogFragment.kt
@@ -6,14 +6,16 @@ import android.os.Bundle
 import android.util.Log
 import androidx.appcompat.app.AlertDialog
 import androidx.fragment.app.DialogFragment
-import dagger.hilt.android.AndroidEntryPoint
-import javax.inject.Inject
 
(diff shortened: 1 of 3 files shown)


ClaudeHangui (approved):
Good for me !

HashEngineering:
This will be using PR: https://github.com/dashevo/dash-wallet/pull/882

HashEngineering (approved):
Looks, great.

PR #886, opened 2022-02-07
feat(explore-dash): track database sync status

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DataSyncStatus.kt
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2021 Dash Core Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
Syn-McJ, line 4: I thought we were using GPL? The year could use an update too :)

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -175,6 +183,34 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
+                            syncStatus.isVisible = true
+                            syncStatus.setBackgroundResource(R.color.red_300)
+                            progress.isVisible = false
+                            when (syncProgress.exception) {
+                                is FileNotFoundException -> syncMessage.text = getString(R.string.sync_in_progress_file_not_found)
+                                else -> syncMessage.text = getString(R.string.sync_in_progress_error)
+                            }
+                        }
+                    }
+                    Status.CANCELED -> binding.syncStatus.isVisible = true
```
Syn-McJ, line 210: Should sync status be visible in this case and if yes, don't we need to set some message?

wallet/src/de/schildbach/wallet/ExploreSyncWorker.kt
```diff
@@ -117,14 +127,21 @@ class ExploreSyncWorker constructor(val appContext: Context, workerParams: Worke
...
+            var merchantsLoaded = 0
             exploreRepository.getMerchantData(merchantDataSizeDB).collect {
                 counter++
                 merchantCache.add(it)
                 if (merchantCache.size == PAGE_SIZE) {
                     merchantDao.save(merchantCache)
                     merchantCache.clear()
                     log.info("MERCHANT $counter of ${merchantDataSize - merchantDataSizeDB} sync took $tableSyncWatch")
+
+                    delay(10000)
```
Syn-McJ, line 139: This should probably be removed if it was used for testing.

wallet/src/de/schildbach/wallet/di/DataSyncStatusModule.kt
```diff
@@ -0,0 +1,37 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.components.SingletonComponent
+import de.schildbach.wallet.ExploreDataSyncProgress
+import org.dash.wallet.features.exploredash.repository.DataSyncStatus
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object DataSyncStatusModule {
```
Syn-McJ, line 32: Admittedly I have no idea myself how to better organize Hilt modules, but I think for this small service it might be unnecessary to create a new file. `DataProviderModule` might be suitable enough to have `DataSyncStatus` in it.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -175,6 +183,34 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
         viewModel.allMerchantLocations.observe(viewLifecycleOwner) { merchantLocations ->
             merchantLocationsAdapter.submitList(merchantLocations)
         }
+
+        syncProgress.observe(viewLifecycleOwner) { syncProgress ->
+            if (syncProgress != null) {
```
Syn-McJ, line 188: This doesn't look like it can be `null`

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -68,6 +71,11 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
     @Inject
     lateinit var configuration: Configuration
 
+    @Inject
+    lateinit var dataSyncStatus: DataSyncStatus
+
+    val syncProgress by lazy { dataSyncStatus.getSyncProgress() }
```
Syn-McJ, line 77: Why not just access it from the service directly? Seems like it will produce the same effect.

Syn-McJ (changes requested):
This works well apart from some UI issues. We definitely need Ros to check how it would look on the online tab.

However, I have a couple of notes that might be more of higher-level nitpicking, yet I feel like not without reasoning:

1. It's better to avoid using LiveData in service classes like `DataSyncStatus`. The reasons for this are:
- LiveData is invoked on the UI thread (unless you do some additional dancing with it), and it might not be what we want. This service should be consumable from any part of the app (e.g. a background logger) and we should be able to process `syncProgress` in a background thread.
- Something like `Flow` provides more flexibility for mapping and filtering. 

In this case, LiveData can be replaced with Flow quite easily.

2. Given that we use MVVM, our viewModel should contain the state of UI. Here `DataSyncStatus` is injected and observed directly from the fragment which should be avoided if possible (I know we have `Configuration` injected above but I think this also should be refactored in the future). 
Additionally, we have other fragments (map fragment) that might care about the sync progress message and adjust their behavior according to it. The map fragment is using the same viewModel so it's beneficial to keep sync progress state in there.

Those two points together would mean that `syncProgress` flow needs to be mapped to a LiveData in the viewModel and only then observed in the fragment, which means more code needs to be written for this feature. But I feel like this would be a cleaner structure and better matching to the rest of the code.

wallet/src/de/schildbach/wallet/ExploreSyncWorker.kt
HashEngineering: Correct.  this is for testing only

wallet/src/de/schildbach/wallet/di/DataSyncStatusModule.kt
HashEngineering: I will move it to `DataProviderModule`

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DataSyncStatus.kt
HashEngineering: I copied the wrong license.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
HashEngineering: This was to get around a NPE.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -175,6 +177,37 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
+                            else -> syncMessage.text = getString(R.string.sync_in_progress_error)
+                        }
+                    }
+                }
+                Status.CANCELED -> {
+                    binding.apply {
+                        syncStatus.isVisible = true
+                        syncMessage.text = getString(R.string.sync_in_progress_canceled)
+                    }
+                }
```
HashEngineering, line 208: currently, a canceled status is not tracked from the Sync Worker.  Not sure which exception to look for, but this could be useful on devices that are so slow they have to start over.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -175,6 +177,37 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
             merchantLocationsAdapter.submitList(merchantLocations)
         }
+        viewModel.syncProgressLiveDataFromFlow.observe(viewLifecycleOwner) { syncProgress ->
+
+            when (syncProgress.status) {
+                Status.LOADING -> {
+                    binding.apply {
+                        syncStatus.isVisible = true
+                        progress.isVisible = true
+                        progress.progress = syncProgress.data?.toInt() ?: 0
```
HashEngineering, line 187: The progress bar is indeterminate, which means this progress value is probably not used for anything.

wallet/src/de/schildbach/wallet/ExploreDataSyncProgress.kt
```diff
@@ -0,0 +1,33 @@
...
+        val log = LoggerFactory.getLogger(ExploreDataSyncProgress::class.java)
+    }
+
+    private val _syncProgressFlow = MutableStateFlow(Resource.loading(0.00))
+
+    override fun getSyncProgressFlow(): Flow<Resource<Double>> = _syncProgressFlow
+
+    override suspend fun setSyncError(exception: Exception) {
+        log.info("explore data sync failure", exception)
+        _syncProgressFlow.emit(Resource.error(exception))
```
HashEngineering, line 22: using `emit()` here was a guess, but it does work.

wallet/src/de/schildbach/wallet/ExploreSyncWorker.kt
```diff
@@ -117,14 +127,21 @@ class ExploreSyncWorker constructor(val appContext: Context, workerParams: Worke
...
+                    merchantsLoaded += PAGE_SIZE
+                    log.info("{} - {}", merchantsLoaded, merchantDataSize)
+                    syncStatus.setSyncProgress(10.0 + 0.9 * (merchantsLoaded.toDouble() / merchantDataSize) * 100.0)
                 }
+                // this math may not be correct
```
HashEngineering, line 144: this comment needs to be deleted.

wallet/src/de/schildbach/wallet/ExploreSyncWorker.kt
```diff
@@ -117,14 +127,21 @@ class ExploreSyncWorker constructor(val appContext: Context, workerParams: Worke
...
                 merchantCache.add(it)
                 if (merchantCache.size == PAGE_SIZE) {
                     merchantDao.save(merchantCache)
                     merchantCache.clear()
                     log.info("MERCHANT $counter of ${merchantDataSize - merchantDataSizeDB} sync took $tableSyncWatch")
+
+                    delay(10000)
+                    merchantsLoaded += PAGE_SIZE
+                    log.info("{} - {}", merchantsLoaded, merchantDataSize)
+                    syncStatus.setSyncProgress(10.0 + 0.9 * (merchantsLoaded.toDouble() / merchantDataSize) * 100.0)
```
HashEngineering, line 142: I made up a formula where 10% is for ATM's and 90% is for merchants.  If the device gets interrupted at least once time, then this percentage will be incorrect.  

With future improvements in syncing, perhaps we cannot determine a % at all with copying a single sqlite file.

Currently the UI is using an indeterminate progress bar (rotating circle), so these numbers are just for experimentation at this point.  They seemed to go from 0% to 100% in the logs.

Syn-McJ (changes requested):
Looks great, however there is a broken unit test - because the ViewModel has a new injection now, it needs to be mocked in the test.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
HashEngineering: Yes, I added some text for this case also, though, currently there is nothing that triggers a canceled state.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
HashEngineering: This is a habit from DashPay and this check was removed.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
HashEngineering: Found a way to inject the DataSyncStatus service into the viewModel, so this code was removed.

HashEngineering:
The tests are hopefully fixed, but there is no specific test for the `DataSyncStatus` flow or values.

Syn-McJ (approved):
Looks great

HashEngineering:
This PR still needs to handle a few more states that the SearchFragment can be in regarding how the user is searching for stores and ATM's

tomasz-ludek (approved):
Well done!

Syn-McJ (approved):

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -653,4 +653,12 @@ class ExploreViewModel @Inject constructor(
...
         addSource(syncStatusService.getSyncProgressFlow().asLiveData(), this::setValue)
     }
+
+    // this is a hack. it was added to clear the exception so the SearchFragment
+    // would stop displaying a Red Error
+    fun clearSyncStatus() {
+        viewModelScope.launch {
+            syncStatusService.setSyncProgress(100.0)
+        }
+    }
```
HashEngineering, line 663: @Syn-McJ I created this function to clear exceptions from the status flow, so that UI would not continue to display an error.

This is not the correct way to do it -- the data shouldn't be changed, rather the UI should instead ignore the error.

Why even do this?

The sync operation happens when the app starts and if the network is offline, the sync operation will fail and the exception goes into the status flow.

When the user goes to Search in Explore Dash, the ViewModel will determine the status and find the exception.  There is a timer that tells the Red error to disappear in 15 seconds.  However, if the user leaves and returns to the Search, the same red error appears about a network error.

During this time the network may again be online, but the app does not attempt to resume sync -- so the error state persists until the app is restarted.

Perhaps a red error message for network offline should be put in a different story and PR that handles network interruptions.

Syn-McJ: Agree about handling network interruptions. Here, we can perhaps have a `syncStatusService.clear()` method or something similar to make it less hacky.

Syn-McJ (approved):
Looks good

PR #887, opened 2022-02-08
fix: copying text on the About screen

Syn-McJ:
There is a small bug that prevents text on the About screen from being copied. `ClipboardManager` gets text from itself instead of a TextView.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Yes, I will put these commits in my PR before it goes to testing.  

This PR will stay open until #886 is reviewed.

HashEngineering:
This was put in the 7.4.6 release ahead of #886, so it will be merged now.

PR #888, opened 2022-02-10
Bugfix: non encrypted wallet

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
+++ b/common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
@@ -47,6 +47,7 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
         private const val POS_BUTTON_ARG = "positive_text"
         private const val NEG_BUTTON_ARG = "negative_text"
 
+        @JvmStatic
         fun simple(
             message: String,
             negativeButtonText: String,
@@ -61,7 +62,7 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
                 positiveButtonText
             )
         }
-
+        // not usable from Java
         fun new(
             @DrawableRes icon: Int,
             title: String,
@@ -79,6 +80,7 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
             )
         }
 
+        @JvmStatic
         fun custom(
             @LayoutRes layout: Int,
             @DrawableRes icon: Int?,
--- a/wallet/src/de/schildbach/wallet/livedata/EncryptWalletLiveData.kt
+++ b/wallet/src/de/schildbach/wallet/livedata/EncryptWalletLiveData.kt
@@ -90,9 +90,11 @@ class EncryptWalletLiveData(application: Application) : MutableLiveData<Resource
                 // For the new key, we create a new key crypter according to the desired parameters.
                 val keyCrypter = KeyCrypterScrypt(scryptIterationsTarget)
                 val newKey = keyCrypter.deriveKey(password)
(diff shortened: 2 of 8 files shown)


wallet/src/de/schildbach/wallet/livedata/EncryptWalletLiveData.kt
```diff
@@ -90,9 +90,11 @@ class EncryptWalletLiveData(application: Application) : MutableLiveData<Resource
                 // For the new key, we create a new key crypter according to the desired parameters.
                 val keyCrypter = KeyCrypterScrypt(scryptIterationsTarget)
                 val newKey = keyCrypter.deriveKey(password)
-                wallet.encrypt(keyCrypter, newKey)
+                // skip this for now
+                //wallet.encrypt(keyCrypter, newKey)
+                //throw KeyCrypterException("fake error")
```
HashEngineering, line 95: This needs to be removed.

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
```diff
@@ -61,7 +62,7 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
                 positiveButtonText
             )
         }
-
+        // not usable from Java
         fun new(
```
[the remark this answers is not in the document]
Syn-McJ: My bad. I was thinking of renaming it to "create" anyway.

wallet/src/de/schildbach/wallet/ui/SetPinActivity.kt
```diff
@@ -436,6 +433,37 @@ class SetPinActivity : InteractionAwareActivity() {
...
+        }
+        val dialog = AdaptiveDialog.new(
+            R.drawable.ic_error,
+            getString(title),
+            getString(message),
+            getString(R.string.button_cancel),
+            getString(R.string.button_ok)
+        )
+        dialog.show(this) {
+            if (it != null && it) {
```
Syn-McJ, line 459: Can be checked as `if (it == true)`

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -795,9 +796,38 @@ private void resetBlockchain() {
...
+                    "no other details are available without the user submitting a report");
+            AdaptiveDialog dialog = AdaptiveDialog.custom(R.layout.dialog_adaptive,
+                    R.drawable.ic_error,
+                    getString(R.string.wallet_encryption_error_title),
+                    getString(R.string.wallet_not_encrypted_error_message),
+                    getString(R.string.button_cancel),
+                    getString(R.string.button_ok)
+            );
+            dialog.show(this, reportIssue -> {
+                if (reportIssue != null) {
```
Syn-McJ, line 817: Is dismissing this dialog without choosing a button is the same as choosing "Cancel"? In this case, the null result should be accounted for. A three-state result (true, false, null) is made to accommodate specific negative action in case it is different from dismissing.

Syn-McJ (approved):
Looks good. A have one concern regarding the dialog being dismissed without any action chosen.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
HashEngineering: Dismissing is the same as pressing cancel.  Though I made a change to this dialog to make it not cancel-able (a button must be pressed).  If that is the case, then there should be no `null` result, right?

Syn-McJ: Yes, this should do.

Syn-McJ (approved):
Looks good

PR #889, opened 2022-02-11
[NMA-1021] (CrowdNode) Creating account

Syn-McJ:
For the CrowdNode integration, we need functionality for creating an account.

- For sending transactions, I have adapted [SendCoinsTask](https://github.com/dashevo/dash-wallet/blob/master/wallet/src/de/schildbach/wallet/ui/send/SendCoinsTask.kt) which wasn't used.
- For receiving transactions, I have created a mechanism for observing transactions that match a certain filter.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/services/SecurityModel.kt
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.dash.wallet.common.services
+
+import androidx.fragment.app.FragmentActivity
+
+interface SecurityModel {
+    suspend fun requestPinCode(activity: FragmentActivity): String?
```
Syn-McJ, line 23: `CheckPinDialog` is a bit heavy on wallet dependencies, so I inject this functionality with DI instead.

wallet/src/de/schildbach/wallet/ui/VerifySeedActivity.kt
```diff
@@ -152,8 +153,13 @@ class VerifySeedActivity : InteractionAwareActivity(), VerifySeedActions {
         goingBack = false
     }
 
-    private fun goHome() {
-        startActivity(Intent(this, WalletActivity::class.java))
+    private fun goBack() {
+        val navigateToHome = intent.getBooleanExtra(NAVIGATE_TO_HOME, true)
```
Syn-McJ, line 157: We don't want to navigate to Home from the CrowdNode entry point.

wallet/src/de/schildbach/wallet/ui/notifications/NotificationManagerWrapper.kt
```diff
@@ -0,0 +1,59 @@
...
+import android.content.Intent
+import androidx.core.app.NotificationCompat
+import androidx.core.app.TaskStackBuilder
+import dagger.hilt.android.qualifiers.ApplicationContext
+import de.schildbach.wallet.Constants
+import de.schildbach.wallet_test.R
+import org.dash.wallet.common.services.NotificationService
+import javax.inject.Inject
+
+class NotificationManagerWrapper @Inject constructor(
```
Syn-McJ, line 32: `NotificationManager` can be injected directly, but if we ever decide to use some in-app notification mechanism instead of system notifications, this will allow making changes easily.

common/src/main/java/org/dash/wallet/common/transactions/ByAddressCoinSelector.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun select(
+        target: Coin,
+        candidates: MutableList<TransactionOutput>
+    ): CoinSelection {
+        val filtered = candidates.filter { output ->
+            val script = output.scriptPubKey
+            (ScriptPattern.isP2PKH(script) || ScriptPattern.isP2SH(script)) &&
+                    script.getToAddress(address.parameters) == address
+        }
```
HashEngineering, line 39: For now all addresses in the wallet that are spendable are P2PKH (Pay to Public Key Hash).  It's possible that at a later point in time, we would support P2SH (which is Pay to Script Hash) -- one such application would be a multisig wallet where multiple parties would be signing the transaction.  That is a long way off, though, if ever.  For now we can still check `isP2SH` as it will be future proof.

common/src/main/java/org/dash/wallet/common/transactions/CoinsFromAddressTxFilter.kt
```diff
@@ -0,0 +1,55 @@
...
+): TransactionFilter {
+    var toAddress: Address? = null
+        private set
+
+    override fun matches(tx: Transaction): Boolean {
+        val actualValue = if (includeFee && tx.fee != null) coins - tx.fee else coins
+        val networkParameters = fromAddress.parameters
+
+        for (input in tx.inputs) {
+            input.outpoint.connectedOutput?.let { connectedOutput ->
```
HashEngineering, line 38: This will only find coins that from an address regarding sent transactions.

Received transactions will have `connectedOutput = null` unless it was an internal transaction (sent from and to our wallet).  That is probably fine considering how this is used.

common/src/main/java/org/dash/wallet/common/transactions/CoinsFromAddressTxFilter.kt
```diff
@@ -0,0 +1,55 @@
...
+                val script = connectedOutput.scriptPubKey
+
+                if ((ScriptPattern.isP2PKH(script) || ScriptPattern.isP2SH(script)) &&
+                    script.getToAddress(networkParameters) == fromAddress
+                ) {
+                    val output = tx.outputs.firstOrNull { it.value == actualValue }
```
HashEngineering, line 44: one thing I am curious about here is if two outputs had the same value in Coin.  One of them is the one you are searching for, the other is for change.  I haven't yet looked at how these filters are used.

common/src/main/java/org/dash/wallet/common/transactions/LockedTransaction.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import org.bitcoinj.core.Sha256Hash
+import org.bitcoinj.core.Transaction
+import org.dash.wallet.common.transactions.TransactionFilter
+
+class LockedTransaction(private val topUpTxId: Sha256Hash): TransactionFilter {
+    override fun matches(tx: Transaction): Boolean {
+        return tx.txId == topUpTxId && tx.confidence.isTransactionLocked
+    }
+}
```
HashEngineering, line 28: Locked means that the transaction was locked by InstantSend.

That means a separate message was received by the network that has to be verified.  There is a small chance that this message is not received, or that the message is not verified.  I need to see how this is used later.

common/src/main/res/values-ca/strings.xml
```diff
@@ -0,0 +1,19 @@
...
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<resources>
+    <string name="button_retry">Re-intenta</string>
+</resources>
```
HashEngineering, line 19: I am curious where these translations came from.  `ca` is a language we have nothing for in `common` currently.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
```diff
@@ -0,0 +1,209 @@
...
+            if (showNotificationOnResult) {
+                notificationService.showNotification(
+                    "crowdnode_ready",
+                    appContext.getString(R.string.crowdnode_account_ready),
+                    notificationIntent
+                )
+            }
+        } catch (ex: Exception) {
+            log.info("CrowdNode error: $ex")
+            analyticsService.logError(ex, "status: ${signUpStatus.value}")
```
HashEngineering, line 117: Excellent

common/src/main/java/org/dash/wallet/common/transactions/LockedTransaction.kt
HashEngineering: In the event that InstantSend is not working, then we may want to broaden the definition of lock as is done in other parts of the app (seen by other peers && confidence is BUILDING)

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/ui/CrowdNodeViewModel.kt
```diff
@@ -17,13 +17,127 @@
...
+            config.crowdNodeAccountAddress = existingAddress.toBase58()
+            return existingAddress
+        }
+
+        val savedAddress = config.crowdNodeAccountAddress
+
+        return if (savedAddress.isNullOrEmpty()) {
+            return createNewAccountAddress()
+        } else {
+            Address.fromString(Context.get().params, savedAddress)
```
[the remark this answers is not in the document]
HashEngineering: We probably should.  `Context.get()` will fail with an exception if called on a thread where `Context.propagate()` has not been called.  Since this is called in a ViewModel, then its probably called in a user thread, which is fine.

if in the future we want to support multiple networks in the same app, then `Context.get().params` may return the wrong network.  So its better that we use the network of the wallet.

common/src/main/res/values-ca/strings.xml
HashEngineering: Nevermind, these strings were moved to `common` from other places.

In Transifex, when we push the update via the source (removing text from one file, adding it to another in english), then in the translations the "new" text in `common` will be auto translated from what was in "memory" from before.

wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
```diff
@@ -0,0 +1,133 @@
...
+                // Re-encrypt wallet with new key.
+                try {
+                    wallet.changeEncryptionKey(newKeyCrypter, key, newKey)
+                    key = newKey
+                    log.info("scrypt upgrade succeeded")
+                } catch (x: KeyCrypterException) {
+                    log.info("scrypt upgrade failed: {}", x.message)
+                }
+            }
+        }
```
HashEngineering, line 128: I was wondering why this was here, but it does come from `DeriveKeyTask`

In the parent app, Bitcoin Wallet, the number of scrypt hash function iterations has increased over time.  For Krip Phones, the number of iterations is set to a smaller number than normal.  If we ever increase that number in a future version or if a user takes a backup file made from Krip Phone or a previous version with a lower number of iterations, then this code would change the key that encrypts the wallet.

wallet/src/de/schildbach/wallet/transactions/WalletTransactionObserver.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.bitcoinj.core.Context
+import org.bitcoinj.core.Transaction
+import org.bitcoinj.utils.Threading
+import org.bitcoinj.wallet.Wallet
+import org.dash.wallet.common.transactions.TransactionFilter
+
+@ExperimentalCoroutinesApi
+class WalletTransactionObserver(private val wallet: Wallet) {
+    fun observe(vararg filters: TransactionFilter): Flow<Transaction> = callbackFlow {
+        Context.propagate(Context(Constants.NETWORK_PARAMETERS)) // TODO: check (and in balance obs)
```
HashEngineering, line 36: wallet has `getContext()` also.

common/src/main/java/org/dash/wallet/common/transactions/CoinsFromAddressTxFilter.kt
HashEngineering: After looking at `toAddress`, I see that it is not checked or used. 

With this method, when a transaction is received (the response), we are checking its inputs to see if it matches one we sent (the request).  My other concern is what if the response does not use the request as the input.

In both cases here, there is a rule being used that might work 99.9999% of the time.

Syn-McJ: I see your point. I suppose this filter does what it describes it does, but if we're deriving from it, we should consider if we want `toAddress` to match our own or not. So the actual changes might be better added to the `CrowdNodeAcceptTermsResponse` and `CrowdNodeWelcomeToApiResponse` filters.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -212,15 +218,12 @@ public void onCoinsReceived(final Wallet wallet, final Transaction tx, final Coi
...
-                    final boolean isReplayedTx = confidenceType == ConfidenceType.BUILDING && (replaying || isRestoringBackup);
+            handler.post(() -> {
+                final boolean isReceived = amount.signum() > 0;
+                final boolean isReplayedTx = confidenceType == ConfidenceType.BUILDING && (replaying || isRestoringBackup);
 
-                    if (isReceived && !isReplayedTx)
-                        notifyCoinsReceived(address, amount, tx.getExchangeRate());
-                }
+                if (isReceived && !isReplayedTx && ignoreCrowdNodeFilter.matches(tx))
+                    notifyCoinsReceived(address, amount, tx.getExchangeRate());
```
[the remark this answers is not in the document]
Syn-McJ: Yes, we can make a more specific filter for the notifications to allow "coins received" in some cases. Or, we might decide to show a specific notification for crowdnode withdrawals.

HashEngineering (approved):
Excellent work.

HashEngineering (approved):

PR #892, opened 2022-02-16
[NMA-1196] Empty results bugfix + ui fixes

Syn-McJ:
When Explore is opened in airplane mode, the map isn't initialized and it doesn't report bounds. This is why DB query isn't triggered. This is a bad design on my part since search results shouldn't depend on the map like this.

- Removed null check for the search bounds.
- Also made some minor UI fixes.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Excellent.

PR #896, opened 2022-02-21
[NMA-1195 & NMA-1197 & NMA-1201] (CrowdNode) Background signup & bug fixes

Syn-McJ:
We want to continue CrowdNode signup process even if the user closed the app.

- Running signup in a CoroutineWorker with `setForeground` to assure that we the worker can broadcast transactions.
- `HiltWorkerFactory` is added to simplify DI in workers and to have Wallet in line with DashPay.
- Fixed the "Insufficient money" issue for the top-up transaction by transferring only 0.009 Dash out of the minimum required 0.01.
- Broadened the conditions for the `LockedTransaction` filter to fix the stuck signup process.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/transactions/LockedTransaction.kt
+++ b/common/src/main/java/org/dash/wallet/common/transactions/LockedTransaction.kt
@@ -19,12 +19,16 @@ package org.dash.wallet.common.transactions
 
 import org.bitcoinj.core.Sha256Hash
 import org.bitcoinj.core.Transaction
-import org.dash.wallet.common.transactions.TransactionFilter
+import org.bitcoinj.core.TransactionConfidence
 
-// TODO: In the event that InstantSend is not working, then we may want to broaden the definition of
-// lock as is done in other parts of the app (seen by other peers && confidence is BUILDING)
 class LockedTransaction(private val topUpTxId: Sha256Hash): TransactionFilter {
     override fun matches(tx: Transaction): Boolean {
-        return tx.txId == topUpTxId && tx.confidence.isTransactionLocked
+        val confidence = tx.confidence
+        val type = confidence.confidenceType
+        val isLocked = confidence.isTransactionLocked ||
+                type == TransactionConfidence.ConfidenceType.BUILDING ||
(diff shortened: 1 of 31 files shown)


features/exploredash/build.gradle
```diff
@@ -53,7 +53,6 @@ dependencies {
     // Core
     implementation 'androidx.core:core-ktx:1.6.0'
     implementation 'androidx.appcompat:appcompat:1.3.1'
-    implementation 'androidx.work:work-runtime-ktx:2.6.0'
```
Syn-McJ, line 56: `ExploreSyncWorker` isn't in the explore-dash module, so this isn't needed atm. However, we might want to move the worker into the module since it would only be logical.

common/src/main/java/org/dash/wallet/common/transactions/LockedTransaction.kt
```diff
@@ -19,12 +19,16 @@ package org.dash.wallet.common.transactions
...
 class LockedTransaction(private val topUpTxId: Sha256Hash): TransactionFilter {
     override fun matches(tx: Transaction): Boolean {
-        return tx.txId == topUpTxId && tx.confidence.isTransactionLocked
+        val confidence = tx.confidence
+        val type = confidence.confidenceType
+        val isLocked = confidence.isTransactionLocked ||
+                type == TransactionConfidence.ConfidenceType.BUILDING ||
+                (type == TransactionConfidence.ConfidenceType.PENDING && confidence.numBroadcastPeers() > 1)
+
+        return tx.txId == topUpTxId && isLocked
```
HashEngineering, line 32: I thought `Transaction.getConfidence` could return `null` -- but perhaps it is not an issue if it didn't happen during previous development.

HashEngineering: Though, if `confidence` was `null` -- then there is a bug in `dashj`

HashEngineering (approved):

HashEngineering (approved):

PR #897, opened 2022-02-22
Feat(coinbase): error handling

ClaudeHangui:
The purpose of this story was to handle the case where a coinbase user would attempt to make a dash purchase for an amount over the authorised limit defined in his account

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/coinbase-integration/src/main/res/layout/dialog_withdrawal_limit_info.xml
```diff
@@ -0,0 +1,219 @@
...
+            android:id="@+id/header_separator"
+            android:layout_width="0dp"
+            app:layout_constraintStart_toStartOf="@id/top_background"
+            app:layout_constraintEnd_toEndOf="@id/top_background"
+            android:layout_height="1dp"
+            android:layout_marginTop="12dp"
+            android:background="@color/divider_color"
+            app:layout_constraintTop_toBottomOf="@id/coinbase_header_label" />
+
+        <ImageView
```
ClaudeHangui, line 60: After a discussion with Ross, he pointed out that this `coinbase` label is actually an image

common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -579,4 +582,36 @@ public void setCoinBaseUserAccountId(String accountId) {
...
+        return prefs.getString(PREFS_KEY_COINBASE_USER_INPUT_AMOUNT, null);
+    }
+
+    public void setCoinbaseUserWithdrawalRemaining(String amount){
+        prefs.edit().putString(PREFS_KEY_COINBASE_USER_WITHDRAWAL_REMAINING, amount).apply();
+    }
+
+    public String getCoinbaseUserWithdrawalRemaining() {
+        return prefs.getString(PREFS_KEY_COINBASE_USER_WITHDRAWAL_REMAINING, getCoinbaseUserWithdrawalLimitAmount());
+    }
```
Syn-McJ, line 616: We keep adding things to this Configuration class which will balloon it sooner or later beyond manageable. I'm thinking we should split it into module-specific configuration classes + a global one in the future.

common/src/main/java/org/dash/wallet/common/ui/payment_method_picker/PaymentMethodPicker.kt
```diff
@@ -90,6 +95,7 @@ class PaymentMethodPicker(context: Context, attrs: AttributeSet): ConstraintLayo
                     showSearch = false
                 ) { _, index, dialog ->
                     dialog.dismiss()
+                    analytics.logEvent(AnalyticsConstants.Coinbase.CHANGE_PAYMENT_METHOD, bundleOf())
```
Syn-McJ, line 98: This is the same mistake as before with the `EnterAmount` fragment. This component should not care about analytics, and it might be used in other integration modules. `AnalyticsService` should be injected.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
```diff
@@ -208,4 +208,9 @@ public static String fiatToStringWithoutCurrencyCode(Fiat fiat) {
     public static String getCoinIcon(String code) {
         return  "https://raw.githubusercontent.com/jsupa/crypto-icons/main/icons/"+code.toLowerCase()+".png";
     }
+
+    public static String getLocaleCurrencyCode(){
+        NumberFormat numberFormat = NumberFormat.getCurrencyInstance(getDeviceLocale());
+        return  numberFormat.getCurrency().getCurrencyCode();
+    }
```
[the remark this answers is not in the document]
ClaudeHangui: Ohh I see, will take a look at it now !

common/src/main/res/drawable/rounded_corner_grey_bgd.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <solid android:color="@color/steel_gray_50" />
+    <corners android:radius="14dp" />
```
Syn-McJ, line 5: Instead of adding a whole bunch of drawables for various backgrounds, you could use the `rounded_background.xml` with a theme, like this:

```
<View
...
android:background="@drawable/rounded_background"
android:theme="@style/SearchInputBackground"
```

in styles:

```
<style name="SearchInputBackground">
    <item name="backgroundColor">@color/gray_40</item>
    <item name="strokeColor">@color/gray_40</item>
    <item name="cornerRadius">8dp</item>
</style>
```

Syn-McJ (changes requested):
@ClaudeHangui would be nice to remove logging from the `PaymentMethodPicker` and move it down to Coinbase module.

common/src/main/java/org/dash/wallet/common/ui/payment_method_picker/PaymentMethodPicker.kt
ClaudeHangui: this is a merge error from my side !

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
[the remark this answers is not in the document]
ClaudeHangui: @HashEngineering updated the implementation of this function as per your suggestions.

common/src/main/java/org/dash/wallet/common/ui/payment_method_picker/PaymentMethodPicker.kt
ClaudeHangui: fixed !

common/src/main/java/org/dash/wallet/common/Configuration.java
HashEngineering: Yes, I was wondering the same thing.

The common module doesn't need to know about settings/configuration that are specific to other modules.  When we do make a change for module-specific vs global, we must take care not to break or lose any settings.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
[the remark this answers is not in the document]
HashEngineering: Great work on this fix.

HashEngineering (approved):

PR #899, opened 2022-02-23
NMA-1200  Buy Swap | Fiat decimal precision

hadia:
The fiat values should only go out two decimals.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
```diff
@@ -200,6 +200,21 @@ public static String fiatToString(Fiat fiat) {
...
+        String currencySymbol = currency.getSymbol(getDeviceLocale());
+       return numberFormat.format(1.0).startsWith(currencySymbol);
+    }
+
+    public static String getLocalCurrencySymbol(String currencyCode) {
+        NumberFormat numberFormat = NumberFormat.getCurrencyInstance(getDeviceLocale());
+        Currency currency = Currency.getInstance(currencyCode);
+        numberFormat.setCurrency(currency);
+        return currency.getSymbol(getDeviceLocale());
+    }
```
HashEngineering, line 216: This will be one merge conflict with Claude's PR:

It is possible for some phones to above old Locale information for certain currencies that have changed (eg Belarus).
```
    public static String getLocaleCurrencyCode(){
        Currency currency = Currency.getInstance(getDeviceLocale());
        String newCurrencyCode = currency.getCurrencyCode();
        if (CurrencyInfo.hasObsoleteCurrency(newCurrencyCode)) {
            newCurrencyCode = CurrencyInfo.getUpdatedCurrency(newCurrencyCode);
        }
        newCurrencyCode = CurrencyInfo.getOtherName(newCurrencyCode);
        return newCurrencyCode;
    }
```

Syn-McJ (approved):
Looks fine apart from Eric's comment

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
hadia: I tried @ClaudeHangui  method we need the local currency Symbol and the one claude create return code

ClaudeHangui: @hadia you can create a data class which holds the symbol and the code, and return that in your case. The `Currency` class has a method to get the symbol; unless of course (like Belarus for example) this might have also change

HashEngineering: @hadia @ClaudeHangui I suppose it is fine.

HashEngineering (approved):

PR #900, opened 2022-02-23
[NMA-1167] (Coinbase) Authorization limit dialog

Syn-McJ:
The first time user tries to log in to Conbase, we need to show a dialog about the correct authorization limit. 

- The dialog is the same as what's implemented in https://github.com/dashevo/dash-wallet/pull/897, however, there are some problems with UI - lack of shadows in particular. We should not ignore shadows if they are in the design. It's not hard to add - just use `CardViewRaised` style or `android:elevation`. I've added `ViewRaised` style that implements the correct shadow color.

I have also moved some of the old styles from the common module to the wallet, to encourage developers to use the new typography and button styles in feature modules. This allows us to start to address a couple of UI issues we have:

- RTL problems for most of which there is a simple fix which is to add `android:textAlignment="viewStart"` to the typography styles. We can only do this for the new styles as adding it to the old ones can break things in the wallet module. However, because we keep mixing the new and the old styles this fix is not as easy.
- There are some changes we can make to typography and button styles to make the introduction of a dark theme easier.

This PR should be merged to feature-coinbase-integration if done after https://github.com/dashevo/dash-wallet/pull/897

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/coinbase-integration/src/main/res/layout/dialog_coinbase_buy_dash.xml
```diff
@@ -43,46 +43,24 @@
...
-    <androidx.appcompat.widget.AppCompatImageView
-        android:id="@+id/buy_dialog_contact_coinbase_support_icon"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:paddingStart="12dp"
-        android:paddingEnd="8dp"
-        android:src="@drawable/ic_blue_support"
-        app:layout_constraintBottom_toBottomOf="@+id/buy_dialog_contact_coinbase_support"
-        app:layout_constraintEnd_toStartOf="@+id/buy_dialog_contact_coinbase_support"
-        app:layout_constraintTop_toTopOf="@+id/buy_dialog_contact_coinbase_support" />
```
Syn-McJ, line 66: This can all be done with just a button

wallet/src/de/schildbach/wallet/data/BuyAndSellDashServicesModel.kt
```diff
@@ -20,7 +21,7 @@ data class BuyAndSellDashServicesModel(
     ) {
         LIQUID(org.dash.wallet.integration.liquid.R.string.liquid, org.dash.wallet.common.R.drawable.ic_liquid),
         UPHOLD(org.dash.wallet.integration.uphold.R.string.uphold_account, org.dash.wallet.common.R.drawable.ic_uphold),
-        COINBASE(org.dash.wallet.integration.coinbase_integration.R.string.coinbase, org.dash.wallet.common.R.drawable.ic_coinbase)
+        COINBASE(org.dash.wallet.integration.coinbase_integration.R.string.coinbase, R.drawable.ic_coinbase)
```
Syn-McJ, line 24: I see no reason to put module-specific resources in the Common. If we do, it will grow to enormous proportions. Separation of resources is also a benefit in modular structure.

ClaudeHangui (approved):
All good for me !
thanks for the heads up in the PR description regarding the design

wallet/src/de/schildbach/wallet/ui/BuyAndSellIntegrationsActivity.kt
```diff
@@ -61,17 +63,18 @@ import java.util.*
 import kotlin.concurrent.schedule
 
 @AndroidEntryPoint
-class BuyAndSellLiquidUpholdActivity : LockScreenActivity(), FancyAlertDialog.FancyAlertButtonsClickListener {
+class BuyAndSellIntegrationsActivity : LockScreenActivity(), FancyAlertDialog.FancyAlertButtonsClickListener {
```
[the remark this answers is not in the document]
HashEngineering: Yes, great idea to change this class name.

HashEngineering (approved):

PR #902, opened 2022-02-25
fix: remove unused coinbase preferences

ClaudeHangui:
- display appropriate error message for input amount greater than withdrawal limit
- handle crash caused by ArithmeticException after user input amount to be sent

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseBuyDashOrderReviewViewModel.kt
+++ b/integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseBuyDashOrderReviewViewModel.kt
@@ -21,7 +21,6 @@ import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import org.bitcoinj.utils.Fiat
-import org.dash.wallet.common.Configuration
 import org.dash.wallet.common.WalletDataProvider
 import org.dash.wallet.common.data.SingleLiveEvent
 import org.dash.wallet.integration.coinbase_integration.model.*
@@ -33,7 +32,6 @@ import javax.inject.Inject
 @HiltViewModel
 class CoinbaseBuyDashOrderReviewViewModel @Inject constructor(
     private val coinBaseRepository: CoinBaseRepository,
-    private val userPreference: Configuration,
     private val walletDataProvider: WalletDataProvider
 ) : ViewModel() {
     private val _showLoading: MutableLiveData<Boolean> = MutableLiveData()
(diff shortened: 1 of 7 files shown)


Syn-McJ (approved):
Looks good

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseBuyDashOrderReviewViewModel.kt
```diff
@@ -80,22 +76,38 @@ class CoinbaseBuyDashOrderReviewViewModel @Inject constructor(
...
                 _showLoading.value = false
-                when (result.value) {
+                val response = result.value
+                when (response.code()) {
                     200, 201 -> {
-                        _transactionCompleted.value = true
+                        _transactionCompleted.value = TransactionState(true, null)
                     }
-                    else -> _transactionCompleted.value = false
+                    400 -> {
```
ClaudeHangui, line 90: For Coinbase's 40x responses, it appears the `safeApiCall()`sometimes sends ResponseResource of type `Success`, with an appropriate error body (if any) which is mainly why we couldn't see the api error message sent by coinbase

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseConversionPreviewViewModel.kt
```diff
@@ -85,16 +85,24 @@ class CoinbaseConversionPreviewViewModel @Inject constructor(
...
                 _showLoading.value = false
-                when (result.value) {
+                val response = result.value
+                when (result.value.code()) {
                     200, 201 -> {
-                        _transactionCompleted.value = true
+                        _transactionCompleted.value = TransactionState(true, null)
                     }
-                    else -> _transactionCompleted.value = false
+                    400 -> {
```
ClaudeHangui, line 93: We needed to replicate this behavior for the `Buy Swap` flow since the end result expected its the same when transferring dash to wallet

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/CoinBaseRepository.kt
```diff
@@ -147,7 +147,7 @@ interface CoinBaseRepositoryInt {
     suspend fun getActivePaymentMethods(): ResponseResource<List<PaymentMethodsData>>
     suspend fun placeBuyOrder(placeBuyOrderParams: PlaceBuyOrderParams): ResponseResource<PlaceBuyOrderUIModel>
     suspend fun commitBuyOrder(buyOrderId: String): ResponseResource<CommitBuyOrderUIModel>
-    suspend fun sendFundsToWallet(sendTransactionToWalletParams: SendTransactionToWalletParams): ResponseResource<Int>
+    suspend fun sendFundsToWallet(sendTransactionToWalletParams: SendTransactionToWalletParams): ResponseResource<Response<SendTransactionToWalletResponse?>>
```
Syn-McJ, line 150: The response type of this method seems somewhat convoluted. There are 3 nested "Responses" of some kind.

When you have a suspend method, there isn't really a reason to use a class like `ResponseResource` in my opinion. The exception is propagated, so you have response/error functionality already, just need to try/catch it. `Loading` object isn't used in Coinbase right now, and the loading state can be implemented in other ways.

Syn-McJ (approved):
Looks good. I have left a comment on return types to consider.

PR #881, opened 2022-01-27
[NMA-1150] ci: Explore Dash Auto Loads

tomasz-ludek:
Use Proto Buffers to load the explore data from file stored on Google Cloud Storage

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #903, opened 2022-02-28
fix: change migration from 8to10 to 9to10 since v7.4.1-5 were version 8

HashEngineering:

in v7.4.6 the Explore Dash Sync System was improved and the Explore Dash data was moved to its own database.

A conversation here was about which versions should be used to remove the Explore Dash data from the primary database:
https://github.com/dashevo/dash-wallet/pull/881#discussion_r804623660

The PR originally had 4 to 9.

However, it was discovered in testing that the explore dash data was not removed after upgrading.

These are the database versions used in previous releases:

v7.4.6 - version 9 - https://github.com/dashevo/dash-wallet/blob/v7.4.6/wallet/src/de/schildbach/wallet/AppDatabase.java#L21
v7.4.5 - version 8 - https://github.com/dashevo/dash-wallet/blob/v7.4.5/wallet/src/de/schildbach/wallet/AppDatabase.java#L32
v7.4.1 - version 8 - https://github.com/dashevo/dash-wallet/blob/v7.4.1/wallet/src/de/schildbach/wallet/AppDatabase.java#L32
v7.3.5 - version 2 - https://github.com/dashevo/dash-wallet/blob/v7.3.5/wallet/src/de/schildbach/wallet/AppDatabase.java#L17

Therefore v7.4.1 to v7.4.5 should have a migration from 8 to 9 (which does work using debug builds).

There are some v7.4.6 installs that did not have an 8to9 migration to drop the tables.  Should we make a version 10 for that and then a migration 8 to 10, 9 to 10?  Or is there a different way?

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/AppDatabase.java
+++ b/wallet/src/de/schildbach/wallet/AppDatabase.java
@@ -18,7 +18,7 @@ import de.schildbach.wallet.rates.ExchangeRatesDao;
 @Database(entities = {
         ExchangeRate.class,
         BlockchainState.class
-    }, version = 9)
+    }, version = 10)
 @TypeConverters({RoomConverters.class})
 public abstract class AppDatabase extends RoomDatabase {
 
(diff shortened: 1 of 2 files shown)


Syn-McJ (approved):
Great

PR #905, opened 2022-03-01
[NMA-1031] (CrowdNode) Portal

Syn-McJ:
UI for the portal and calculating current balance using CrowdNode rest API.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeWebApi.kt
```diff
@@ -15,15 +15,15 @@
...
+import retrofit2.Response
+import retrofit2.http.*
 
-@AndroidEntryPoint
-class PortalFragment : Fragment(R.layout.fragment_portal) {
-    private val binding by viewBinding(FragmentPortalBinding::bind)
-}
\ No newline at end of file
+interface CrowdNodeWebApi {
+    @GET("odata/apifundings/GetFunds(address='{address}')")
```
Syn-McJ, line 25: For now, we'll be using this endpoint, but perhaps in the future, there'll be something more suitable for the balance specifically.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/RemoteDataSource.kt
```diff
@@ -0,0 +1,54 @@
...
+import okhttp3.logging.HttpLoggingInterceptor
+import org.dash.wallet.common.BuildConfig
+import org.dash.wallet.common.Configuration
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+import javax.inject.Inject
+
+class RemoteDataSource {
+    companion object {
+        private const val BASE_URL = "https://test.crowdnode.io/"
```
Syn-McJ, line 30: This class can probably be generalized to combine with the one in Coinbase later.

features/exploredash/src/main/res/values-es/strings-explore-dash.xml
```diff
@@ -34,7 +34,7 @@
     <string name="explore_option_buy">Comprar</string>
     <string name="explore_option_sell">Vender</string>
     <string name="explore_option_buy_sell">Comprar y Vender</string>
-    <string name="explore_menu_info">Información</string>
+    <string name="button_info">Información</string>
```
HashEngineering, line 37: In Transifex, what will happen is this:
1. we will push the source files, which will no longer have `explore_menu_info`, but instead `button_info`
2. `button_info` is new and no languages will have this string translated
3. Transifex will see that languages are missing `button_info`, but will see in memory how `"info"` was translated.  Then it will automatically fill out `button_info` for all languages that had a translation for `"info"`

We won't need to push the changes in this particular file to Transifex, but next time we pull translations, `button_info` should be included.

HashEngineering (approved):
This is well done, especially noting with comments the places that were added for QA and those places that may need to change later.

PR #906, opened 2022-03-01
feat: logout user invalid token

ClaudeHangui:
For any user on Coinbase whose token is invalid (for whatever reason) or expired, any attempt to access protected resource endpoints for Coinbase should force the user to be logged out 

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/remote/TokenAuthenticator.kt
```diff
@@ -25,18 +30,23 @@ import org.dash.wallet.common.Configuration
...
+    context: Context
 ) : Authenticator {
 
+    val entryPoint = EntryPointAccessors.fromApplication(context, CoinbasePortalEntryPoint::class.java)
+
     override fun authenticate(route: Route?, response: Response): Request? {
         return runBlocking {
             when (val tokenResponse = getUpdatedToken()) {
                 is ResponseResource.Success -> {
+                    closeCoinbasePortal(response)
```
ClaudeHangui, line 49: For a token invalid, revoked or expired, a 401 response code is provided with an error body. It appears our `ResponseResource` class doesn't always catch the Coinbase 40x response in the catch block. As such, we might have a response of type Success

HashEngineering (changes requested):
This has merge conflicts.

PR #907, opened 2022-03-01
feat (coinbase): handle coinbase invalid or revoked token

ClaudeHangui:
When the user's token is invalid/revoked for whatever reason, we should forcefully log out the user from Coinbase so that he can login again and get a new token

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/remote/TokenAuthenticator.kt
+++ b/integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/remote/TokenAuthenticator.kt
@@ -16,6 +16,11 @@
  */
 package org.dash.wallet.integration.coinbase_integration.repository.remote
 
+import android.content.Context
+import dagger.hilt.EntryPoint
+import dagger.hilt.InstallIn
+import dagger.hilt.android.EntryPointAccessors
+import dagger.hilt.components.SingletonComponent
 import kotlinx.coroutines.runBlocking
 import okhttp3.Authenticator
 import okhttp3.Request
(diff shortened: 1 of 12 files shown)


integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/remote/TokenAuthenticator.kt
```diff
@@ -25,33 +30,51 @@ import org.dash.wallet.common.Configuration
...
 import javax.inject.Inject
 
 class TokenAuthenticator @Inject constructor(
     private val tokenApi: CoinBaseTokenRefreshApi,
-    private val userPreferences: Configuration
+    private val userPreferences: Configuration,
+    context: Context
 ) : Authenticator {
 
+    val entryPoint = EntryPointAccessors.fromApplication(context, CoinbasePortalEntryPoint::class.java)
```
ClaudeHangui, line 43: We needed to inject our broadcast class in a non android class. This call allows us to to that.
We needed an instance of the Context class, thus, it had to be provided as a dependency

HashEngineering (approved):

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/remote/TokenAuthenticator.kt
Syn-McJ: Generally, when you find yourself needing to inject something into a class that isn't an Android component and doesn't support injection, it pays to stop and think if this is the right place to do the injection.

`TokenAuthenticator` seems like a fairly generic class that could be reused across features with some modifications, but now not only it is tied specifically to Coinbase, but also it is aware of the context, our dependency framework, and responsible for broadcasting changes. Quite a lot of SOLID principles are broken here.

Syn-McJ (changes requested):
@ClaudeHangui I have an issue with this solution for broadcasting authentication state from the `TokenAuthenticator`. First, this type of broadcasting is a lazy pattern that makes it difficult to follow the logical flow and allows listening to the broadcast from anywhere in the app which isn't a good thing. Second, it complicates the chain of dependencies. A good chain of dependencies looks like this: 

While a bad one looks like this:



Additionally, `CoinbaseActivity` already has a ViewModel that can be used for this type of communication. Perhaps we can use the exception throwing mechanism, catching it down the line closer to the viewModel and listening to the event through the viewModel.

There is also a `CoinBaseAuthRepository` class that doesn't seem to be used but could be more suitable for handling this error.

Could you please investigate if an alternative solution of this kind can be used?

PR #909, opened 2022-03-03
[NMA-1219] Updating okhttp version

Syn-McJ:
Now that the minimum Android version is 6.0 we can update okhttp to version 4.x

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -614,62 +614,6 @@ private void checkAlerts() {
         url.addEncodedQueryParameter("package", packageInfo.packageName);
         url.addQueryParameter("current", Integer.toString(packageInfo.versionCode));
 
-		/*new HttpGetThread(url.build(), application.httpUserAgent()) {
```
Syn-McJ, line 617: This has been commented out for 8 years. Probably safe to remove.

HashEngineering: Bitcoin Wallet uses this method to determine if there is a new version published to Google Play.  If so, the app can remind to upgrade using Google Play or perhaps Fdroid depending on which store they were using.

We never used this mechanism.

HashEngineering (approved):

PR #910, opened 2022-03-04
feat: logout coinbase user with invalid token

ClaudeHangui:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


ClaudeHangui:
@Syn-McJ,
`TokenAuthenticator`  is an extension of `Authenticator` class from okhttp3, used to catch unauthorised (precisely 401) errors from a server; `authenticate()` is it's callback and refreshing of the token is handled there. As you see it returns a nullable value of type `Request`. I tried to implement all the approaches we discussed and it didn't seem to work
 I couldn't find a better way of doing this; if you've got any ideas I'm opened to solutions

Syn-McJ (approved):
@ClaudeHangui got it, yeah I see how this isn't an easy case. Let's leave it for now.

HashEngineering (approved):

PR #912, opened 2022-03-10
[NMA-1030] (CrowdNode) Deposit

Syn-McJ:
The user should be able to send a specified amount of Dash to CrowdNode.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
```diff
@@ -36,30 +35,43 @@ import org.dash.wallet.common.transactions.ByAddressCoinSelector
...
+    override suspend fun sendCoins(
+        address: Address,
+        amount: Coin,
+        constrainInputsTo: Address?,
+        emptyWallet: Boolean
+    ): Transaction {
         val wallet = walletApplication.wallet ?: throw RuntimeException("this method can't be used before creating the wallet")
         Context.propagate(wallet.context)
-        val sendRequest = createSendRequest(address, amount, constrainInputsTo)
+        val sendRequest = createSendRequest(address, amount, constrainInputsTo, emptyWallet)
```
Syn-McJ, line 52: Since we need to first top-up our CrowdNode account address and then send the required amount from it to CrowdNode, we either need to top-up amount + fee or use `emptyWallet` property.

HashEngineering: Determining the fee is more involved.  

One way to include the fee is this:
1. Take the intended deposit amount and add 192 duffs (the fee for sending 1 input to one output)
2. Send that to our CrowdNode address
3. Then do `emptyWallet` to send to CrowdNode (the fee is 192 duffs/satoshis).

Not sure if the user would care about

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
```diff
@@ -63,6 +61,27 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
...
+            return custom(
+                R.layout.dialog_progress,
+                null,
+                null,
+                message,
+                "",
+                null
+            ).apply { isCancelable = false }
+        }
+
```
HashEngineering, line 84: In DashPay, the FancyAlertDialog does have some progress related feature.  Not sure if that was your inspiration here, but if we were to replace `FancyAlertDialog` with `AdaptiveDialog`, we would need these progress features in the DashPay branch.

Syn-McJ: Progress in the `FancyAlertDialog` used quite a lot in the Coinbase branch. I figured we might want to put it in the `AdaptiveDialog` as well since `FancyAlertDialog` is being faded out. I suppose we can do the same for DashPay.

wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
Syn-McJ: > The user may not care about their depositing being short my 192 duffs.

This is what I thought but then I guess it's still better to account for it. I'll change that in the withdraw story.

HashEngineering: That will be fine with me.

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
```diff
@@ -63,6 +61,27 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
             )
         }
 
+        suspend fun withProgress(massage: String, activity: FragmentActivity, action: suspend () -> Unit) {
+            val dialog = progress(massage)
```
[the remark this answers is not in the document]
Syn-McJ: Cleaned up (this and imports)

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
```diff
@@ -128,8 +147,8 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
         val iconView: ImageView? = view.findViewById(R.id.dialog_icon)
         val titleView: TextView? = view.findViewById(R.id.dialog_title)
         val messageView: TextView = view.findViewById(R.id.dialog_message)
-        val positiveButton: TextView = view.findViewById(R.id.dialog_positive_button)
-        val negativeButton: TextView = view.findViewById(R.id.dialog_negative_button)
+        val positiveButton: TextView? = view.findViewById(R.id.dialog_positive_button)
+        val negativeButton: TextView? = view.findViewById(R.id.dialog_negative_button)
```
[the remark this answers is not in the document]
Syn-McJ: Correct. Initially, I thought there should be at least 1 button in any dialog (for closing), however, the progress dialog doesn't have any buttons and other custom layouts might also not have any.

HashEngineering (approved):
Excellent

PR #913, opened 2022-03-15
feat: network unavailablity coinbase

ClaudeHangui:
We needed to handle network state changes in the various Coinbase screens. For a device offline, we needed to inform the user of his connectivity status; the entry points to some screens were disabled; when the connection is re-established we resume the default behavior(s)
In other to achieve this we needed to refactor the `ConnectivityLiveData` class used for monitoring network status. 
Regarding changing the drawables in the Main Portal (`BuyAndSellIntegrationsActivity`) on network state change, we couldn't think of a better way of informing the adapter to change the icons for the various Buy & Sell services.
   

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/ConnectivityViewModel.kt
```diff
@@ -1,11 +1,30 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+import org.dash.wallet.common.livedata.NetworkStateInt
+import javax.inject.Inject
+
+@ExperimentalCoroutinesApi
+@HiltViewModel
+open class ConnectivityViewModel @Inject constructor(private val networkStateProvider: NetworkStateInt)
```
ClaudeHangui, line 17: We centralised the value being returned by our `NetworkState` class in this ViewModel; so every (child) VM which is required to provide to it's associated view the state of the network has it from here.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseBuyDashOrderReviewFragment.kt
```diff
@@ -159,6 +165,14 @@ class CoinbaseBuyDashOrderReviewFragment : Fragment(R.layout.fragment_coinbase_b
...
             countDownTimer.start()
         }
+
+        parentFragmentManager.beginTransaction()
+            .replace(R.id.network_status_container, NetworkUnavailableFragment.newInstance())
+            .commit()
+
+        amountViewModel.isDeviceConnectedToInternet.observe(viewLifecycleOwner){ hasInternet ->
+            setNetworkState(hasInternet)
+        }
```
Syn-McJ, line 175: I'm not sure that it is the responsibility of the `EnterAmountViewModel` to track the connectivity state. Its responsibility is only to enter an amount, so we're breaking the single responsibility principle here. Perhaps it's better to derive `CoinbaseBuyDashOrderReviewViewModel` from the `ConnectivityViewModel`?

Syn-McJ (changes requested):
Looks good. A lot of new files are missing the copyright, would be nice to add. Also, I have a comment about tracking the network state in the `EnterAmountViewModel`, please check.

Syn-McJ (approved):
Looks good

HashEngineering (approved):
There will be at least one merge conflict or compile error when `master` is merged into `feature-coinbase-integration` after this PR is merged.

PR #915, opened 2022-03-16
feat: track firebase events from explore dash portal and info screens

ClaudeHangui:
Track with Firebase click events from Explore Dash Portal and Dialog Info screens

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
+++ b/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
@@ -100,4 +100,11 @@ object AnalyticsConstants {
         const val TRANSACTION_FILTER = "home_transaction_filter"
         const val SEND_RECEIVE_BUTTON = "home_send_receive_button"
     }
+
+    object ExploreDash {
+        const val WHERE_TO_SPEND = "portal_where_to_spend"
+        const val PORTAL_ATM = "portal_atm"
+        const val LEARN_MORE = "info_learn_more"
+        const val CONTINUE = "info_continue"
+    }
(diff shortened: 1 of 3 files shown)


Syn-McJ (approved):
Looks good

common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
```diff
@@ -100,4 +100,11 @@ object AnalyticsConstants {
         const val TRANSACTION_FILTER = "home_transaction_filter"
         const val SEND_RECEIVE_BUTTON = "home_send_receive_button"
     }
+
+    object ExploreDash {
+        const val WHERE_TO_SPEND = "portal_where_to_spend"
+        const val PORTAL_ATM = "portal_atm"
+        const val LEARN_MORE = "info_learn_more"
+        const val CONTINUE = "info_continue"
```
HashEngineering, line 108: We should have an "explore_dash_" prefix on these event names, otherwise they will not be sorted as a group.  Check previously created event names which have prefixes that note the screen, function or service that they apply to.

HashEngineering (changes requested):
Let's add a prefix to the names.

common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
ClaudeHangui: @HashEngineering done !

HashEngineering (approved):
Excellent work.

PR #917, opened 2022-03-18
feat: explore dash merchant search event tracking

ClaudeHangui:
Use Firebase to track user interactions related to explore dash for merchants
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
+++ b/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
@@ -100,4 +100,11 @@ object AnalyticsConstants {
         const val TRANSACTION_FILTER = "home_transaction_filter"
         const val SEND_RECEIVE_BUTTON = "home_send_receive_button"
     }
+
+    object ExploreDash {
+        const val WHERE_TO_SPEND = "portal_where_to_spend"
+        const val PORTAL_ATM = "portal_atm"
+        const val LEARN_MORE = "info_learn_more"
+        const val CONTINUE = "info_continue"
+    }
 }
--- a/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
+++ b/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
@@ -102,19 +102,19 @@ object AnalyticsConstants {
     }
 
     object ExploreDash {
-        const val WHERE_TO_SPEND = "portal_where_to_spend"
-        const val PORTAL_ATM = "portal_atm"
-        const val LEARN_MORE = "info_learn_more"
-        const val CONTINUE = "info_continue"
-        const val ONLINE_MERCHANTS = "online_merchants"
-        const val NEARBY_MERCHANTS = "nearby_merchants"
-        const val ALL_MERCHANTS = "all_merchants"
-        const val FILTER_MERCHANTS_TOP = "filter_merchants_top"
-        const val FILTER_MERCHANTS_BOTTOM = "filter_merchants_bottom"
-        const val SELECT_MERCHANT_LOCATION = "select_merchant_location"
-        const val SELECT_MERCHANT_MARKER = "select_merchant_marker"
-        const val INFO_EXPLORE_MERCHANT = "info_search"
-        const val PAN_MERCHANT_MAP = "pan_merchant_map"
-        const val ZOOM_MERCHANT_MAP = "zoom_merchant_map"
+        const val WHERE_TO_SPEND = "explore_portal_where_to_spend"
+        const val PORTAL_ATM = "explore_portal_atm"
+        const val LEARN_MORE = "explore_info_learn_more"
+        const val CONTINUE = "explore_info_continue"
+        const val ONLINE_MERCHANTS = "explore_online_merchants"
+        const val NEARBY_MERCHANTS = "explore_nearby_merchants"
+        const val ALL_MERCHANTS = "explore_all_merchants"
+        const val FILTER_MERCHANTS_TOP = "explore_filter_merchants_top"
+        const val FILTER_MERCHANTS_BOTTOM = "explore_filter_merchants_bottom"
+        const val SELECT_MERCHANT_LOCATION = "explore_select_merchant_location"
+        const val SELECT_MERCHANT_MARKER = "explore_select_merchant_marker"
+        const val INFO_EXPLORE_MERCHANT = "explore_info_search"
+        const val PAN_MERCHANT_MAP = "explore_pan_merchant_map"
+        const val ZOOM_MERCHANT_MAP = "explore_zoom_merchant_map"
     }
 }
--- a/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
+++ b/common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
@@ -106,5 +106,15 @@ object AnalyticsConstants {
         const val PORTAL_ATM = "portal_atm"
         const val LEARN_MORE = "info_learn_more"
         const val CONTINUE = "info_continue"
+        const val ONLINE_MERCHANTS = "online_merchants"
+        const val NEARBY_MERCHANTS = "nearby_merchants"
+        const val ALL_MERCHANTS = "all_merchants"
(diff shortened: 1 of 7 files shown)


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
```diff
@@ -107,6 +128,10 @@ class ExploreMapFragment : SupportMapFragment() {
                         map.cameraPosition.zoom
                     )
                 }
+
+                map.setOnCameraMoveStartedListener { reason ->
```
ClaudeHangui, line 132: When the camera motion starts we need to know the reason

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
```diff
@@ -97,6 +99,25 @@ class ExploreMapFragment : SupportMapFragment() {
             googleMap?.let { map ->
                 map.setOnCameraIdleListener {
                     val bounds = map.projection.visibleRegion.latLngBounds
+                    if (cameraMovementReason == GoogleMap.OnCameraMoveStartedListener.REASON_GESTURE){
+                        when {
+                            viewModel.hasZoomLevelChanged(previousZoomLevel, map.cameraPosition.zoom) -> {
+                                viewModel.zoomLevelChangeCallback.call()
+                            }
+                            viewModel.hasCameraCenterChanged(
```
ClaudeHangui, line 107: Check if a pan gesture was made: we rely on the coordinates (position) of the camera in the center of the map

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
```diff
@@ -97,6 +99,25 @@ class ExploreMapFragment : SupportMapFragment() {
             googleMap?.let { map ->
                 map.setOnCameraIdleListener {
                     val bounds = map.projection.visibleRegion.latLngBounds
+                    if (cameraMovementReason == GoogleMap.OnCameraMoveStartedListener.REASON_GESTURE){
+                        when {
+                            viewModel.hasZoomLevelChanged(previousZoomLevel, map.cameraPosition.zoom) -> {
```
ClaudeHangui, line 104: Check if the user zoomed (in/out) the map

Syn-McJ: 1. This entire `when` can be moved to the ViewModel. You're calling a viewModel check and then calling a viewModel event if the check is satisfied. Just call a `triggerEventIfSatisfied(param, param)` method once.
2. This entire check is needed to trigger an event that is then observed from another fragment that logs the event. I don't quite get why can't you either
    - Log it from this fragment
    - Log it from the viewModel

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -22,6 +22,7 @@ import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.*
 import androidx.paging.*
 import androidx.paging.PagingData
+import com.google.android.gms.maps.model.LatLng
```
Syn-McJ, line 25: The viewModel should not reference LatLng since it's implementation-dependent and we're working with an interface in case of `UserLocationState`. If the implementation of `UserLocationState` changed, you'll have to change a lot of related code in the viewModel as well. Please use `GeoBounds` instead.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -222,6 +230,21 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
+            if (viewModel.exploreTopic == ExploreTopic.Merchants){
+                analyticsService.logEvent(AnalyticsConstants.ExploreDash.SELECT_MERCHANT_MARKER, bundleOf())
+            }
+        }
+        viewModel.zoomLevelChangeCallback.observe(viewLifecycleOwner){
+            if (viewModel.exploreTopic == ExploreTopic.Merchants){
+                analyticsService.logEvent(AnalyticsConstants.ExploreDash.ZOOM_MERCHANT_MAP, bundleOf())
+            }
+        }
+        viewModel.cameraCenterChangeCallback.observe(viewLifecycleOwner){
```
Syn-McJ, line 243: `SearchFragment` doesn't have to know anything about the map, I don't get why this is here.

Syn-McJ (changes requested):
@ClaudeHangui there are some problems with events being passed around from one fragment to another for no reason. This can be simplified greatly.

ClaudeHangui:
@Syn-McJ there was actually some reason for me handling the tracking of some events in `SearchFragment`: I originally had those events in the `ExploreMapFragment` but I realised that we are to implement these same events for the ATM Search as well, and since the SearchFragment is aware of the `ExploreTopic` I thought it was ideal for it to handle all the map-related trackings. Guess it was the wrong choice. Thanks for the feedback !

Syn-McJ:
@ClaudeHangui I see, well there is no reason not to log events from the ViewModel - in fact, I would prefer not doing injections into fragments at all. But not all user interactions go into the ViewModel and so it might not be always convenient.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreMapFragment.kt
```diff
@@ -98,32 +96,12 @@ class ExploreMapFragment : SupportMapFragment() {
...
-                        bounds.northeast.latitude,
-                        bounds.northeast.longitude,
-                        bounds.southwest.latitude,
-                        bounds.southwest.longitude,
-                        bounds.center.latitude,
-                        bounds.center.longitude,
-                        map.cameraPosition.zoom
-                    )
+                    viewModel.previousZoomLevel = map.cameraPosition.zoom
+                    viewModel.previousCameraGeoBounds = getGeoBounds()
```
ClaudeHangui, line 104: using `searchBounds` in our `triggerPanAndZoomEvents()` gives us a NPE, I'm not sure I understand why; so I had to create another property to hold the current geobounds. Moreover, it appears this searchBounds var is no longer used in the ExploreMapFragment

Syn-McJ: @ClaudeHangui if you pass the `map` instance into `getGeoBounds`, you can make the return value non-nullable. Then you won't have to accept nullable `GeoBounds` into the `triggerPanAndZoomEvents`. Then you won't have to use the `!!` operator on it down the line.
`previousCameraGeoBounds` also either shouldn't be nullable or shouldn't be unpacked with the `!!` operator - it's better to make use of kotlins null-safety features when you can.

Syn-McJ (approved):
Looks great

common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
```diff
@@ -100,4 +100,21 @@ object AnalyticsConstants {
...
+        const val ONLINE_MERCHANTS = "online_merchants"
+        const val NEARBY_MERCHANTS = "nearby_merchants"
+        const val ALL_MERCHANTS = "all_merchants"
+        const val FILTER_MERCHANTS_TOP = "filter_merchants_top"
+        const val FILTER_MERCHANTS_BOTTOM = "filter_merchants_bottom"
+        const val SELECT_MERCHANT_LOCATION = "select_merchant_location"
+        const val SELECT_MERCHANT_MARKER = "select_merchant_marker"
+        const val INFO_EXPLORE_MERCHANT = "info_search"
+        const val PAN_MERCHANT_MAP = "pan_merchant_map"
+        const val ZOOM_MERCHANT_MAP = "zoom_merchant_map"
```
HashEngineering, line 118: Same comment as #915, there should be a prefix, otherwise within the list of events, some explore dash events are at A and others at Z.

HashEngineering (changes requested):
Add prefix to event names.

HashEngineering (approved):
excellent work

PR #922, opened 2022-03-23
feat: merchant filter event tracking

ClaudeHangui:
Track events for user interaction on filter screen in explore dash
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -740,4 +739,49 @@ class ExploreViewModel @Inject constructor(
...
+        analyticsService.logEvent(
+            if (_isLocationEnabled.value == true) AnalyticsConstants.ExploreDash.FILTER_LOCATION_ALLOWED else
+                AnalyticsConstants.ExploreDash.FILTER_LOCATION_DENIED,
+            bundleOf()
+        )
+        analyticsService.logEvent(AnalyticsConstants.ExploreDash.FILTER_APPLY_ACTION, bundleOf())
+        analyticsService.logEvent(
+            if (isDialogFullyExpanded)
+            AnalyticsConstants.ExploreDash.FILTER_SWIPE_DOWN_ACTION_OFF else
+                AnalyticsConstants.ExploreDash.FILTER_SWIPE_DOWN_ACTION_ON, bundleOf())
```
Syn-McJ, line 781: What is the purpose of these events and `isDialogFullyExpanded` value? As I understood from the story, the `Swipe Down` AC is simply for tracking the filter dialog dismissal. So it's the same as `FILTER_CANCEL_ACTION` but for dismissing.

Syn-McJ (commented):
Left some comments

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
ClaudeHangui: @Syn-McJ  regarding `the swipe down` action I'll have to confirm this with Brian since the `Cancel Action` from my understanding is to be recorded when the user click the `Cancel` button from the dialog
Thanks !

common/src/main/java/org/dash/wallet/common/services/analytics/AnalyticsConstants.kt
```diff
@@ -127,5 +127,22 @@ object AnalyticsConstants {
...
+        const val FILTER_ONE_MILE = "explore_one_mile"
+        const val FILTER_FIVE_MILE = "explore_five_miles"
+        const val FILTER_TWENTY_MILE = "explore_twenty_miles"
+        const val FILTER_FIFTY_MILE = "explore_fifty_miles"
+        const val FILTER_LOCATION_ALLOWED = "explore_location_allowed"
+        const val FILTER_LOCATION_DENIED = "explore_location_denied"
+        const val FILTER_APPLY_ACTION = "explore_apply_action"
+        const val FILTER_CANCEL_ACTION = "explore_cancel_action"
+        const val FILTER_SWIPE_DOWN_ACTION_ON = "explore_swipe_down_on"
+        const val FILTER_SWIPE_DOWN_ACTION_OFF = "explore_swipe_down_off"
```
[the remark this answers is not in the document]
ClaudeHangui: @Syn-McJ actually all events (besides the one on the `Cancel` button) are to be recorded when the user clicks the `Apply` button. This also applies to the type the location selected and whether or not the user had the gift card or dash payment methods enabled or not
Adding the prefix seems a good idea, thanks !

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
ClaudeHangui: @Syn-McJ it appears we are to handle the swipe action only for manual actions; the idea being to detect whether or not the user performed a swipe gesture in the filter screen and record this value once the button Apply is clicked

Syn-McJ: @ClaudeHangui judging by Brian's comments, we just need 1 event when the dialog is closed by dismissing. Thus it's not clear why do you need 2 events here. You probably don't even need to track `onSlide` as well. Just have a variable that is set when the user clicked on cancel, and then track the "dismissed" event. Variable set  = `explore_filter_cancel_action`, not set = `explore_filter_dismissed`

ClaudeHangui: @Syn-McJ I hear you, thanks for insisting on the matter !

Syn-McJ (approved):
Looks good

HashEngineering (approved):

HashEngineering (approved):
Excellent.  All event names are less than 40 chars.

PR #924, opened 2022-03-28
fix: finish onboarding if upgrading with unencrypted wallet

HashEngineering:

v6 and v5 wallets will go to the Set Pin screen during the upgrade process if the app doesn't have a PIN set.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):
Looks good

PR #925, opened 2022-03-28
feat: set firebase events for merchant details

ClaudeHangui:
We are required to record a set of events related to user interaction within the explore dash for merchant details

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ItemDetails.kt
```diff
@@ -41,10 +41,14 @@ import java.util.*
 class ItemDetails(context: Context, attrs: AttributeSet): LinearLayout(context, attrs) {
     private val binding = ItemDetailsViewBinding.inflate(LayoutInflater.from(context), this)
 
-    private var onSendDashClicked: (() -> Unit)? = null
+    private var onSendDashClicked: ((isPayingWithDash: Boolean) -> Unit)? = null
```
ClaudeHangui, line 44: We are required to track only when the `Pay with Dash` button is clicked

HashEngineering (approved):

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -844,4 +844,77 @@ class ExploreViewModel @Inject constructor(
...
+            )
+        }
+    }
+
+    fun trackPayWithDashEvent() {
+        analyticsService.logEvent(
+            AnalyticsConstants.ExploreDash.MERCHANT_DETAILS_PAY_WITH_DASH,
+            bundleOf()
+        )
+    }
```
Syn-McJ, line 919: Does it need to be a separate method for each event? Perhaps a `trackEvent(name: String)` method is sufficient, with the caller deciding which event name constant to provide.

Otherwise, half of the viewModel will consist of these methods.

Syn-McJ (approved):
Looks good, but methods in the viewModel can be simplified to only have one general method, instead of one for each event.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
ClaudeHangui: @Syn-McJ I set up a function to handle tracking in vm
please check it out !

Syn-McJ (approved):

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
Syn-McJ: @ClaudeHangui looks great!

PR #928, opened 2022-04-08
Feat(dashpay): log event timing for platform functions (NMA-1249)

HashEngineering:

https://github.com/dashevo/dapi-client-android/pull/41
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
@@ -32,7 +32,9 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import org.bitcoinj.core.Address
 import org.bouncycastle.crypto.params.KeyParameter
+import org.dash.wallet.common.services.analytics.AnalyticsConstants
 import org.dash.wallet.common.services.analytics.AnalyticsService
+import org.dash.wallet.common.services.analytics.AnalyticsTimer
 import org.slf4j.LoggerFactory
 import javax.inject.Inject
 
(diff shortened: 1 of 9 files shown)


wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
```diff
@@ -69,6 +71,19 @@ open class DashPayViewModel @Inject constructor(
...
+
+    fun startUsernameSearchTimer() {
+        timerUsernameSearch = AnalyticsTimer(analytics, log, AnalyticsConstants.Process.PROCESS_USERNAME_SEARCH_UI)
+    }
+
+    fun reportUsernameSearchTime(resultSize: Int, searchTextSize: Int) {
+        timerUsernameSearch?.logTiming(
+            Pair("resultCount", resultSize),
+            Pair("searchCount", searchTextSize)
+        )
```
Syn-McJ, line 84: I wonder if it's not better to use a map, like this:
```
timer.logTiming(
    mapOf(
        "resultCount" to resultSize,
        "searchCount" to searchTextSize
    )
)
```
Then in the `logTiming` you can have `forEach { (key, value) -> }` which is more verbose than `it.first, it.second`

Syn-McJ (approved):
Looks good, one small suggestion.

wallet/src/de/schildbach/wallet/ui/dashpay/DashPayViewModel.kt
HashEngineering: Thanks I will make this change because a map seems easier to work with than `vararg`s.

Syn-McJ (approved):

Syn-McJ (approved):

PR #929, opened 2022-04-11
[NMA-1223] Merge 7.4.7 to Dash Pay

Syn-McJ:
We need to merge 7.4.7 (Explore Dash) into `dashpay` branch and resolve conflicts.

A couple of things to note:
- After the merge, show/hide logic for the shortcuts was a bit of a mess - we had changes that were improving it slightly in DashPay already, but then in the wallet, the old logic was extended. I decided to refactor and simplify it further. Details are below in the diff.
- `PaymentsFragment` was derived from `BottomNavFragment` and assumed to be hosted in `MainActivity`, which created problems when I tried to put it into Explore navigation graph. I figured it's better if we refactor this to move `BottomNavFragment` related logic into one place in the `MainActivity` so that the information about the hosting activity doesn't leak into the fragments. Refactoring isn't complete here, but it is started.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/Constants.java
```diff
@@ -115,7 +113,7 @@
                 SUPPORTS_PLATFORM = true;
                 SYNC_FLAGS.add(MasternodeSync.SYNC_FLAGS.SYNC_HEADERS_MN_LIST_FIRST);
                 SYNC_FLAGS.add(MasternodeSync.SYNC_FLAGS.SYNC_BLOCKS_AFTER_PREPROCESSING);
-                FAUCET_URL = "http://faucet.krupnik.networks.dash.org/";
+                org.dash.wallet.common.Constants.FAUCET_URL = "http://faucet.krupnik.networks.dash.org/";
```
Syn-McJ, line 116: This is a bit weird, but we need `FAUCET_URL` in another module.
I think the correct approach to consider for the future is using DI and having an interface with all necessary network-specific values, while the implementation will be provided depending on the build flavor.

wallet/src/de/schildbach/wallet/ui/CreateUsernameActivity.kt
```diff
@@ -289,7 +289,9 @@ class CreateUsernameActivity : InteractionAwareActivity(), TextWatcher {
                     }
                 })
                 showProcessingState()
-                startService(CreateIdentityService.createIntentFromInvite(this, username, invite))
+                invite?.let {
```
Syn-McJ, line 292: Kotlin 1.5 requires stricter null-safety, `invite` doesn't fit into `createIntentFromInvite` since it's nullable.

wallet/src/de/schildbach/wallet/ui/DashPayUserActivity.kt
```diff
@@ -99,11 +99,14 @@ class DashPayUserActivity : LockScreenActivity(),
...
             updateContactRelationUi()
         } else {
-            val dashPayProfile = intent.getParcelableExtra(EXTRA_INIT_PROFILE_DATA) as DashPayProfile
-            viewModel.updateProfileData(dashPayProfile) // save the profile to the database for non-contacts
-            viewModel.userData = UsernameSearchResult(dashPayProfile.username, dashPayProfile, null, null)
-            viewModel.initUserData(dashPayProfile.username).observe(this) {
-                updateContactRelationUi()
+            val dashPayProfile = intent.getParcelableExtra(EXTRA_INIT_PROFILE_DATA) as DashPayProfile?
+
+            if (dashPayProfile != null){
```
Syn-McJ, line 104: Kotlin 1.5 requires stricter null-safety

wallet/src/de/schildbach/wallet/ui/InviteHandlerActivity.kt
```diff
@@ -75,8 +75,10 @@ class InviteHandlerActivity : AppCompatActivity() {
         }
 
         initViewModel()
-        if (externalInvite != null) {
-            viewModel.handleInvite(externalInvite)
+        val invite = externalInvite
+
+        if (invite != null) {
```
Syn-McJ, line 80: Kotlin 1.5 requires stricter null-safety

wallet/src/de/schildbach/wallet/ui/MainActivity.kt
```diff
@@ -281,11 +286,20 @@ class MainActivity : AbstractBindServiceActivity(), ActivityCompat.OnRequestPerm
...
+                WalletFragment::class -> R.id.bottom_home
+                ContactsFragment::class,
+                UpgradeToEvolutionFragment::class -> R.id.contacts
+                PaymentsFragment::class -> R.id.payments
+                ExploreFragment::class -> R.id.discover
+                MoreFragment::class -> R.id.more
+                else -> binding.bottomNavigation.selectedItemId
             }
+
+            binding.bottomNavigation.menu.findItem(itemId).isChecked = true
```
Syn-McJ, line 300: Instead of keeping the type of bottom navigation in the fragments themselves and having them check an appropriate bottom bar button, we can get the currently displayed fragment in the `MainActivity` and keep the `isChecked` assignment in here.

wallet/src/de/schildbach/wallet/ui/MainActivity.kt
```diff
@@ -281,11 +286,20 @@ class MainActivity : AbstractBindServiceActivity(), ActivityCompat.OnRequestPerm
...
         }
-        binding.bottomNavigation.setOnNavigationItemSelectedListener { item ->
+        binding.bottomNavigation.setOnItemSelectedListener { item ->
```
Syn-McJ, line 302: `setOnNavigationItemSelectedListener` is deprecated

wallet/src/de/schildbach/wallet/ui/dashpay/BottomNavFragment.kt
```diff
@@ -19,37 +19,33 @@ package de.schildbach.wallet.ui.dashpay
...
-            navParentView.forceHideViewToHide = forceHideBottomNav
-        } else {
-            mainActivity.bottom_navigation.visibility = if (forceHideBottomNav) View.VISIBLE else View.GONE
+        mainActivity?.let { activity ->
+            val navParentView = activity.bottom_navigation.parent.parent
+            if (navParentView is KeyboardResponsiveCoordinatorLayout) {
+                navParentView.forceHideViewToHide = forceHideBottomNav
+            } else {
+                activity.bottom_navigation.visibility = if (forceHideBottomNav) View.VISIBLE else View.GONE
+            }
```
Syn-McJ, line 48: `navigationItem.isChecked = true` logic is moved to `MainActivity`, but this one is left here for now. But I think it would be better to move it in there as well and get rid of `BottomNavFragment`. This will allow using the fragments that now are derived from it freely in other places, as well as help with the transition to the navigation component.

wallet/src/de/schildbach/wallet/ui/dashpay/CropImageActivity.kt
```diff
@@ -89,14 +89,15 @@ class CropImageActivity : LockScreenActivity() {
                         background.setImageDrawable(resource)
                         circle_crop.setImageDrawable(resource)
                         circle_crop.maxZoom = 5f
-                        if (initZoomedRect != null) {
+
+                        initZoomedRect?.let { rect ->
```
Syn-McJ, line 93: kotlin 1.5 strick null-safety (`initZoomedRect` is lazy and so a simple null-check doesn't work).

wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
```diff
@@ -130,12 +130,13 @@ class PlatformRepo private constructor(val walletApplication: WalletApplication)
 
     private var lastPreBlockStage: PreBlockStage = PreBlockStage.None
 
-    private val analytics: AnalyticsService
+    private val analytics: AnalyticsService by lazy {
+        walletApplication.analyticsService
+    }
```
Syn-McJ, line 135: This will avoid problems with Firebase Analytics initialization that we had in explore-dash

wallet/src/de/schildbach/wallet/ui/widget/ShortcutsPane.kt
```diff
@@ -79,87 +81,96 @@ class ShortcutsPane(context: Context, attrs: AttributeSet) : LinearLayout(contex
...
-                View.VISIBLE
-            } else View.GONE
+        var slotsLeft = if (isSmallScreen) 3 else 4
+        shortcuts.forEach { btn ->
+            if (btn.shouldAppear && slotsLeft > 0) {
+                addShortcut(btn)
+                slotsLeft--
+            } else {
+                removeShortcut(btn)
+            }
```
Syn-McJ, line 156: After trying to sort out our rules for displaying shortcuts, I've realized that there are 2 problems:
- Visibility of shortcuts is handled with both `addView` and `isVisible`. This leads to having 2 different methods `setup` and `refresh` while there should be only 1.
- `ShorcutsPane` mixes together the logic for each button and the logic for small/big screens. This should be split up.

The change I've made:
- Having buttons added/removed only by `addShortcut` and `removeShortcut` methods. There is no ambiguity as to how to show and hide those buttons anymore.
- Having all buttons we want to ever display in the current app version in the `shortcuts` list. This is the only place that dictates the order of buttons.
- The `shouldAppear` property in the `ShortcutButton` is the only thing that tells whether the conditions are satisfied for showing the button.
- `refresh` checks the list of buttons, checks which buttons want to be displayed and takes maximum based on the screen size.

.gitignore
```diff
@@ -29,4 +29,6 @@ vendor/
 fastlane/report.xml
 google-services.json
 output-metadata.json
-service.properties
\ No newline at end of file
+service.properties
+wallet/schemas/*/*.json
```
HashEngineering, line 33: It seems like this line keeps appearing in our `.gitignore` in various PR's - there is at least one test that uses these files.

.gitignore
```diff
@@ -29,4 +29,6 @@ vendor/
...
+wallet/assets/explore/*.bin
```
HashEngineering, line 34: Our Explore Dash system no longer uses any `.bin` files

build.gradle
```diff
@@ -16,11 +20,13 @@ buildscript {
...
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.3'
+        classpath 'com.android.tools.build:gradle:4.2.0'
         classpath 'com.google.gms:google-services:4.3.5'
-        classpath 'com.google.firebase:firebase-crashlytics-gradle:2.5.1' // Crashlytics
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        classpath 'com.google.firebase:firebase-crashlytics-gradle:2.5.1' // Crashlytics
         classpath "com.google.dagger:hilt-android-gradle-plugin:$hiltVersion"
+        classpath "androidx.navigation:navigation-safe-args-gradle-plugin:$navigationVersion"
+        classpath "com.google.android.libraries.mapsplatform.secrets-gradle-plugin:secrets-gradle-plugin:1.3.0"
```
HashEngineering, line 29: I wonder if there is something like this that could be used to project other secrets unrelated to map keys.

.gitignore
Syn-McJ: Looks like it's removed from .gitignore in the head of master. Let's leave until the next merge.

build.gradle
Syn-McJ: It seems like this plugin actually **can** be used for any other secrets in the `local.properties`: https://github.com/google/secrets-gradle-plugin

HashEngineering: Yes, I will make a Story for this.

.gitignore
HashEngineering: That is fine with me.

HashEngineering (approved):

PR #825, opened 2021-11-29
fix: reset autologout & close dialogs on autologout (NMA-985)

ClaudeHangui:
As per the story description, we were required to generalize the handling of the alert dialogs displayed througout the app. This required a lot of code refactoring

## Issue being fixed or feature implemented
- Replace the existing DialogBuilder with a more customizable builder class
- close any opened dialog after app gets locked

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [ x] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #932, opened 2022-04-18
[NMA-504 && NMA-1211] Bugfixes with "Send to Address" 

Syn-McJ:
There were a couple of problems we had with "Send to Address" button:
- Some error cases weren't showing the dialog due to changes in #825 
- If a text with Dash address is copied but it doesn't match exactly, an error is shown.

- Changed error dialogs to the `AdaptiveDialog` and wrote a bunch of tests ensuring that it shows when needed.
- Added searching for an address in the `InputParser` and wrote a bunch of tests for various copied texts.

There were several screens related to the Options menu in the `WalletActivity` which we don't need anymore, and several instrumented tests we don't use in the Wallet module. This includes the database migration test - since we have a separate DB for explore now, this test isn't needed.

Additionally, I have created a `MainActivityViewModel` for using it in the `WalletActivity`. This view model already exists in the `dashpay` branch. We can move some of the logic to it from the activity to simplify merging in the future (although we will need to resolve conflicts at least once with this view model).

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


fastlane/Fastfile
```diff
@@ -20,7 +20,7 @@ platform :android do
   lane :test do |options|
     flavor=options[:flavor].capitalize
 	type=options[:type].capitalize
-    gradle(task: "test#{flavor}#{type}UnitTest")
+    gradle(task: "test")
```
Syn-McJ, line 23: For some reason `test...UnitTest` skips tests in the `wallet` module.

wallet/AndroidManifest.xml
```diff
@@ -157,12 +157,6 @@
             android:configChanges="keyboard|keyboardHidden"
             android:screenOrientation="portrait"
             android:theme="@style/LockScreenActivity.Child.Theme"></activity>
-        <activity
-            android:name="de.schildbach.wallet.ui.RequestCoinsActivity"
```
Syn-McJ, line 161: Was used in one of the Context Menu options in the `WalletActivity`. Not needed anymore.

wallet/src/de/schildbach/wallet/ui/InputParser.java
```diff
@@ -423,27 +440,6 @@ protected void cannotClassify(final String input) {
         error(null, R.string.input_parser_cannot_classify, input);
     }
 
-    public static void dialog(final Context context,
```
Syn-McJ, line 426: `dialog()` methods weren't used and in any case, `InputParser` shouldn't be responsible for dialogs.

wallet/src/de/schildbach/wallet/ui/LockScreenActivity.kt
```diff
@@ -67,16 +68,14 @@ open class LockScreenActivity : SecureActivity() {
...
-
-    val walletApplication: WalletApplication = WalletApplication.getInstance()
-    private val configuration = walletApplication.configuration
-    private val autoLogout: AutoLogout = walletApplication.autoLogout
-    @Inject
-    lateinit var lockScreenBroadcaster: LockScreenBroadcaster
+    @Inject lateinit var walletApplication: WalletApplication
+    @Inject lateinit var lockScreenBroadcaster: LockScreenBroadcaster
+    @Inject lateinit var configuration: Configuration
+    private val autoLogout: AutoLogout by lazy { walletApplication.autoLogout }
```
Syn-McJ, line 74: Injecting these fields instead of calling `getInstance()` allows to run unit tests on derived activities.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -190,35 +184,38 @@ public void onChanged(de.schildbach.wallet.data.BlockchainState blockchainState)
             }
         });
 
-        RefreshUpdateShortcutsPaneViewModel model = new ViewModelProvider(this).get(RefreshUpdateShortcutsPaneViewModel.class);
```
Syn-McJ, line 193: Refreshing shortcuts moved into `MainActivityViewModel`.

wallet/src/de/schildbach/wallet/ui/WalletActivity.java
```diff
@@ -331,10 +328,26 @@ public void onActivityResult(final int requestCode, final int resultCode, final
     }
 
     private void handleString(String input, final int errorDialogTitleResId, final int cannotClassifyCustomMessageResId) {
```
Syn-McJ, line 330: I also wanted to move this method into the view model, but this would create problems with some of the tests. Leaving it here for now.

HashEngineering (changes requested):
There is one question about the database test file and one about payment requests.

wallet/androidTest/de/schildbach/wallet/ui/WalletActivityTest.java
```diff
@@ -1,61 +0,0 @@
...
-import static androidx.test.espresso.action.ViewActions.click;
-import static androidx.test.espresso.intent.Intents.intended;
-import static androidx.test.espresso.intent.matcher.IntentMatchers.hasComponent;
-import static androidx.test.espresso.matcher.ViewMatchers.withId;
-
-@RunWith(AndroidJUnit4.class)
-@Ignore
-public class WalletActivityTest {
-
-    @Rule
```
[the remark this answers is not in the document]
Syn-McJ: My bad, I haven't noticed that it was changed somewhat.
Not sure if it's testing the migration correctly atm - the right way would be to fill the version 8 database with some merchant data and then run the migration, making sure that merchants are deleted and nothing is broken.

But I suppose it does check the integrity of the other 2 tables, so it's ok. Would be good to actually get instrumented tests running during our CI though.

wallet/androidTest/de/schildbach/wallet/database/DatabaseMigrationTest.kt
```diff
@@ -1,94 +0,0 @@
...
-    )
-
-    @Test
-    @Throws(IOException::class)
-    fun migrateAll() {
-        // Create db and fill with data
-        testHelper.createDatabase(TEST_DB_NAME, 8).apply {
-            var values = ContentValues()
-            values.put("id", 1)
-            values.put("bestChainDate", 1633356847000)
```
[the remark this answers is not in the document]
Syn-McJ: My bad, I haven't noticed that it was changed somewhat.
Not sure if it's testing the migration correctly atm - the right way would be to fill the version 8 database with some merchant data and then run the migration, making sure that merchants are deleted and nothing is broken.

But I suppose it does check the integrity of the other 2 tables, so it's ok. Would be good to actually get instrumented tests running during our CI though.

wallet/test/de/schildbach/wallet/util/activities/WalletActivityTest.kt
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    @Test
+    fun handlePaste_paymentRequest_startsSendCoins() {
+        val amount = 1.72
+        val dashAddress = if (Constants.NETWORK_PARAMETERS == MainNetParams.get()) {
+            "XjBya4EnibUyxubEA8D2Y8KSrBMW1oHq5U"
+        } else {
+            "yeEpP1q7ow319JYbnQCq7ucwPcjkeSyuoz"
+        }
```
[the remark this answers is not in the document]
Syn-McJ: Yes, this piece is just resolving the address, the actual requests we're testing are on the lines 128 and 135. `dash:` and `pay:` are covered.

wallet/src/de/schildbach/wallet/ui/InputParser.java
```diff
@@ -173,11 +167,34 @@ public void parse() {
...
+                } catch (final AddressFormatException x) {
+                    log.info("got invalid address", x);
+                    error(x, R.string.input_parser_invalid_address);
+                }
+
+                return true;
+            }
+
+            return false;
+        }
```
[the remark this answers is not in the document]
Syn-McJ: Yes, it's my thinking too - request URL is less likely to get tangled together with other text.

Pattern matching `dash:<address>&amount=<amount>` would be more complicated indeed (and I'm no expert in regex). Additionally, it would not work for the case when the request itself has an error in it (like with that bad symbol Andrei had in his testing).

I think extracting a correct dash address should be enough here.

wallet/androidTest/de/schildbach/wallet/database/DatabaseMigrationTest.kt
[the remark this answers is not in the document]
HashEngineering: Merchant data is in a different database.  My most recent changes were to get this test to compile.

HashEngineering (approved):
Excellent work.  Concerns are addressed.

PR #939, opened 2022-04-26
Custodial Swaps | Error Handling

hadia:
Handling Swap errors for cases

- The user should be informed if they enter an amount that is too low
- The user should be informed if they enter an amount that is too high
- The user should be notified if the amount is over their authorized amount (same error as given by the API)
- Network unavailable 

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):
Looks good

ClaudeHangui (approved):
Works for me!

PR #940, opened 2022-04-26
feat: transfer dash to/from coinbase

ClaudeHangui:
As a user we want to be able to transfer Dash to/from Coinbase from/to the Dash Wallet

- A custom view (i.e `ConvertView`) to handle swap direction change has already been provided (by Hadia) from `Convert Crypto` story/code; however it appears there has been some change in the design for this view. This design change has been submitted after the story has been done. As far as the `Transfer Dash` story is concerned, it has to be provided with the new design. I assume the design changes related to Crypto Conversion will be done in another story.
So in other to prevent any future conflict with the custom `ConvertView` class, we decided to create our custom view but using the same xml file

- Completing the transfer from Dash -> Coinbase requires authentication either by PIN, FaceID: the classes used to construct and therefore handle PIN/FaceID authentications are in the `wallet` module. So, in other to properly use those components, we hade to move the corresponding VM's (`SingleActionSharedViewModel` && `CheckPinSharedModel`) used in the authentication from `wallet -> common` to handle authentication states and define an interactor/interface to display the PIN/FaceId dialogs

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (changes requested):
You had the right idea about `PinInteractor`, however, the dialog itself can be extended to use callback or coroutine for returning the result. This way, there is no need to move `CheckPinSharedModel` around and we can get rid of it later.

I have already done this for CrowdNode, let me create a quick PR to maser which you can then integrate.

HashEngineering:
@ClaudeHangui The PR #941  has been merged to master.

ClaudeHangui:
@Syn-McJ I made the changes as per your suggestions; please check this out

wallet/src/de/schildbach/wallet/ui/UpholdTransferActivity.kt
```diff
@@ -152,8 +150,7 @@ class UpholdTransferActivity : InteractionAwareActivity() {
 
                 val fee = transaction.origin.fee.toPlainString()
                 val total = transaction.origin.amount.toPlainString()
-                val dialog = ConfirmTransactionDialog.createDialog(address, amountStr, amountFiat, fiatSymbol, fee, total, getString(R.string.uphold_transfer))
-                dialog.show(supportFragmentManager, "ConfirmTransactionDialog")
+                ConfirmTransactionDialog.createDialog(this@UpholdTransferActivity, address, amountStr, amountFiat, fiatSymbol, fee, total, getString(R.string.uphold_transfer))
```
Syn-McJ, line 153: @ClaudeHangui isn't a call to `dialog.show()` needed here?

wallet/src/de/schildbach/wallet/ui/send/ConfirmTransactionDialog.kt
```diff
@@ -47,21 +53,56 @@ class ConfirmTransactionDialog : BaseBottomSheetDialogFragment() {
...
-            bundle.putString(ARG_FIAT_SYMBOL, fiatSymbol)
-            bundle.putString(ARG_FEE, fee)
-            bundle.putString(ARG_TOTAL, total)
-            bundle.putString(ARG_PAYEE_NAME, payeeName)
-            bundle.putString(ARG_PAYEE_VERIFIED_BY, payeeVerifiedBy)
-            bundle.putString(ARG_BUTTON_TEXT, buttonText)
-            dialog.arguments = bundle
-            return dialog
+            val bundle = setBundle(address, amount, amountFiat, fiatSymbol, fee, total, payeeName, payeeVerifiedBy, buttonText)
+            showDialog(dialog, bundle, activity)
```
Syn-McJ, line 63: @ClaudeHangui it's not obvious that `createDialog` calls `show`. These should either be different methods like in `AdaptiveDialog`, or this method should be renamed

Syn-McJ (changes requested):
@ClaudeHangui changes with pin dialog look good, however I have a small change request for the `ConfirmTransactionDialog`, see the comments.

Syn-McJ (approved):
Looks great

HashEngineering (approved):

PR #942, opened 2022-04-26
fix: don't start blockchain service if the app is in the background

HashEngineering:

Changes are related to starting the Blockchain Service if the app is in the background.  This problem may be a result of a race condition if a Contact Request is received when the app is in the background as the Blockchain Service is shutting down.

When the Blockchain Service is shutting down, that should also stop the Job that is pulling Contact Request Updates.

Though QA has passed, the problem of trying to start the Blockchain service while in the background was rare, so we don't actually know if this fix works.  We have to wait for the Alpha Program feedback.

This PR is being merged into the Release 5.2 branch, though it also needs to be merged into the `dashpay` branch in some way.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
+++ b/wallet/src/de/schildbach/wallet/ui/dashpay/PlatformRepo.kt
@@ -15,6 +15,9 @@
  */
 package de.schildbach.wallet.ui.dashpay
 
+import android.app.ActivityManager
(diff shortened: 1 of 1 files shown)


Syn-McJ (approved):
Looks great

HashEngineering:
the branch isn't labeled correctly to activate CI

PR #945, opened 2022-05-01
[NMA-1033] (CrowdNode) Linking online account

Syn-McJ:
A user with an existing online account should be able to link his mobile wallet as an API address.

Apart from online linking I also:
- Fixed MAX button behavior. We should set `emptyWallet` when max is chosen (perhaps there will be a warning there before sending).
- Refactored CrowdNode module config so that it takes generic prefs instead of creating new pair of methods for each pref.
- Added cleaning datastore prefs and webivew cookies on wallet reset.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/LockScreenActivity.kt
```diff
@@ -151,6 +152,16 @@ open class LockScreenActivity : SecureActivity() {
         }
     }
 
+    protected open fun turnOffAutoLogout() {
+        autoLogout.stopTimer()
+    }
+
+    protected open fun turnOnAutoLogout() {
```
Syn-McJ, line 159: We have these methods in the `InteractionAwareActivity` for turning off autologout for liquid/uphold custom tabs. In crowdnode, it's needed for turning it off in the login webview.

Perhaps `LockScreenActivity` itself should derive from  `InteractionAwareActivity` so that we don't duplicate this code, however, the latest has some logic that finishes the activity on lock screen which we don't need here.

HashEngineering: Perhaps it should.

HashEngineering (approved):

PR #946, opened 2022-05-02
feat: transfer dash from wallet to coinbase

ClaudeHangui:
We want the user to be able to transfer dash from the wallet to Coinbase

The work related to this story has originally been started and almost completed in [this PR](https://github.com/dashevo/dash-wallet/pull/940)
The main improvement brought here was the implementation to get the blockchain fee used in the transaction

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/TransferDashFragment.kt
+++ b/integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/TransferDashFragment.kt
@@ -37,6 +37,7 @@ import org.bitcoinj.utils.MonetaryFormat
 import org.dash.wallet.common.Constants
 import org.dash.wallet.common.services.ConfirmTransactionService
 import org.dash.wallet.common.services.SecurityModel
+import org.dash.wallet.common.services.SendPaymentService
 import org.dash.wallet.common.ui.*
 import org.dash.wallet.common.ui.dialogs.AdaptiveDialog
 import org.dash.wallet.common.util.GenericUtils
@@ -67,8 +68,8 @@ class TransferDashFragment : Fragment(R.layout.transfer_dash_fragment) {
     private var loadingDialog: FancyAlertDialog? = null
     @Inject lateinit var securityModel: SecurityModel
     @Inject lateinit var confirmTransactionLauncher: ConfirmTransactionService
+    @Inject lateinit var transactionDetails: SendPaymentService
     private var dashValue: Coin = Coin.ZERO
-    private var fiatValue: Fiat? = null
     private val dashFormat = MonetaryFormat().withLocale(GenericUtils.getDeviceLocale())
         .noCode().minDecimals(2).optionalDecimals()
 
@@ -113,7 +114,6 @@ class TransferDashFragment : Fragment(R.layout.transfer_dash_fragment) {
         }
 
         enterAmountToTransferViewModel.onContinueTransferEvent.observe(viewLifecycleOwner){
-            fiatValue = it.first
             dashValue = it.second
             if (binding.transferView.walletToCoinbase){
                 val coinInput = it.second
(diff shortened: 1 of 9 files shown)


wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
```diff
@@ -33,39 +33,86 @@ import org.bouncycastle.crypto.params.KeyParameter
...
 import org.slf4j.LoggerFactory
+import java.security.GeneralSecurityException
 import javax.inject.Inject
 
 
 class SendCoinsTaskRunner @Inject constructor(
     private val walletApplication: WalletApplication
 ) : SendPaymentService {
     private val log = LoggerFactory.getLogger(SendCoinsTaskRunner::class.java)
+    val wallet = walletApplication.wallet ?: throw RuntimeException("this method can't be used before creating the wallet")
```
Syn-McJ, line 44: There is a chance that `SendCoinsTaskRunner` is injected and initiated before the wallet is created. Also, the wallet can change but this reference here will remain to the old one. It's better to keep this check in the methods, it's fine if it is duplicated.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/TransferDashFragment.kt
```diff
@@ -211,15 +211,17 @@ class TransferDashFragment : Fragment(R.layout.transfer_dash_fragment) {
...
-            val fiatSymbol = if (fiatValue != null) GenericUtils.currencySymbol(fiatValue?.currencyCode) else ""
-            val amountStr = dashValue.toPlainString()
-            val fee = VALUE_ZERO
-            val total = dashValue.toPlainString()
+            val fiatVal = enterAmountToTransferViewModel.getFiat(dashValue.toPlainString())
+            val amountFiat = dashFormat.format(fiatVal).toString()
+            val fiatSymbol = GenericUtils.currencySymbol(fiatVal.currencyCode)
+
             lifecycleScope.launch {
+                val details = transactionDetails.estimateNetworkFee(transferDashViewModel.dashAddress, dashValue)
```
Syn-McJ, line 219: I wonder if we should do this in CrowdNode to estimate fee for the topup. @HashEngineering you think it would make sense, or just leave it a constant?

Syn-McJ (approved):
Looks good, just a small note in `SendCoinsTaskRunner`

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/TransferDashFragment.kt
HashEngineering: I was wondering something similar, but regarding the `sendCoins` methods in the `SendCoinsTaskRunner` class.  

Regarding the Crowdnode and estimating fees, this method may be useful.  It may help prevent very rare edge case problems. though such problems may not be possible.

wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
```diff
@@ -55,17 +56,63 @@ class SendCoinsTaskRunner @Inject constructor(
...
+            val scryptIterationsTarget = walletApplication.scryptIterationsTarget()
+            val encryptionKey = deriveKey(wallet, password, scryptIterationsTarget)
+
+            sendRequest.aesKey = encryptionKey
+
+            wallet.completeTx(sendRequest)
+            if (checkDust(sendRequest)){
+                sendRequest = createSendRequest(address, amount, constrainInputsTo, emptyWallet)
+                wallet.completeTx(sendRequest)
+            }
```
HashEngineering, line 79: This check for dust with `checkDust` is very important because it helps to prevent an edge case where the transaction will be rejected by the network.

Unfortunately, recreating the exact same `sendRequest` will not resolve the issue.

From `PaymentProtocolViewModel`, we find that there is a difference in the sendRequests that are recreated related to `forceEnsureMinRequiredFee`.  First it is `false` and if this results in dust, then it is set to `true`
```
                var sendRequest = createSendRequest(wallet, false, paymentIntent, signInputs = false, forceEnsureMinRequiredFee = false)

                wallet.completeTx(sendRequest)
                if (checkDust(sendRequest)) {
                    sendRequest = createSendRequest(wallet, false, paymentIntent, signInputs = false, forceEnsureMinRequiredFee = true)
                    wallet.completeTx(sendRequest)
                }
```

This may be sufficient before the second call to `wallet.completeTx(sendRequest)`
```
        sendRequest.ensureMinRequiredFee = forceEnsureMinRequiredFee
        wallet.completeTx(sendRequest)
```

HashEngineering: For crowdnode when setting up an account, these three transactions were involved (only fees shown).

```
       fee  0.00001008 DASH/kB, 0.00000227 DASH for 225 bytes
       fee  0.00001008 DASH/kB, 0.00000227 DASH for 225 bytes
       fee  0.00001005 DASH/kB, 0.00000376 DASH for 374 bytes
 ```
 
 I was expecting the fees to be 0.00001000 DASH given that `ensureMinRequiredFee` is set to true rather than 0.00000227 DASH or 0.00000376 DASH.
 
 Perhaps somethings on the backend in DashJ have changed over the years and my understanding no longer applies.

HashEngineering: After further checking, the check for dust may not be required, but let us leave it for now.

Changes in DashJ in the past year or so have made fee calculation more robust. and less prone to errors.

HashEngineering (approved):
Great Work.

PR #947, opened 2022-05-04
[NMA-1263] Fix: syncing status during initialization

Syn-McJ:
During initialization of `BlockchainService`, our syncing status on the home page doesn't show, and `isSynced()` returns true. It should return false from initialization until it is confirmed that the blockchain is either syncing or up to date.

- Syncing issue is fixed like this: upon the start, the wallet sets the progress to 0 - `isSynced()` will return false. To avoid showing "syncing: 0%" in UI we just show "Syncing...". Then the blockchain either starts to synchronize and we switch to a percentage value, or `doneDownload()` is called, progress set to 100 and the syncing message is hidden.
- Our UI folder in the `wallet` module has a quite long list of files in it and its' hard to navigate around. I have moved some of them to dedicated subpaths: `main` - for MainActivity, view model, and Home related fragments; `rates` - for `ExchangeRates` fragments, activity and viewModel.
- I did some refactoring - primary in the HeaderBalanceFragment - to move more code in the MainViewModel. This is something we can also do with WalletActivity to simplify dashpay merge.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/res/values/strings-extra.xml
```diff
@@ -332,7 +332,6 @@
     <string name="welcome_screen_subtitle_3">Have complete control and customize your wallet based on your needs</string>
 
     <string name="sync_status_wait">Please wait for the sync to complete</string>
-    <string name="sync_status_syncing_title">Syncing %1$d</string>
```
Syn-McJ, line 335: Not used anymore

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -855,6 +874,12 @@ public Address currentReceiveAddress() {
         return wallet.currentReceiveAddress();
     }
 
+    @NonNull
+    @Override
+    public Flow<Coin> observeBalance() {
+        return new WalletBalanceObserver(wallet).observe();
```
Syn-McJ, line 880: Lifted from CrowdNode

wallet/src/de/schildbach/wallet/data/BlockchainStateDao.kt
```diff
@@ -30,7 +31,7 @@ import androidx.room.Query
 abstract class BlockchainStateDao {
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
-    abstract fun insert(blockchainState: BlockchainState)
+    protected abstract fun insert(blockchainState: BlockchainState)
```
Syn-McJ, line 34: Given that `save` has additional checks, it's better to hide this method so that no one uses it by mistake.

wallet/src/de/schildbach/wallet/ui/main/WalletActivity.java
```diff
@@ -830,17 +836,17 @@ public void onNewKeyChainEncrypted() {
     }
 
     private void updateSyncState() {
-        if (blockchainState == null) {
-            return;
-        }
+        Boolean isSyncFailed = viewModel.isBlockchainSyncFailed().getValue();
 
-        if (blockchainState != null && blockchainState.syncFailed()) {
+        if (isSyncFailed != null && isSyncFailed) {
```
Syn-McJ, line 841: The reason I decided to move `blockchainState` to the view model is that fragments/activities don't really need to know about it. They should only care about the state of UI - synced/failed, etc.

wallet/src/de/schildbach/wallet/ui/rates/ExchangeRatesViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.ViewModel
+import dagger.hilt.android.lifecycle.HiltViewModel
+import de.schildbach.wallet.rates.ExchangeRatesRepository
+import org.dash.wallet.common.data.ExchangeRate
+
+/**
+ * @author Samuel Barbosa
+ */
+class ExchangeRatesViewModel : ViewModel() {
```
Syn-McJ, line 29: This viewModel is pretty useless since it just maps livedata from one place to another. Leaving it be for now due to a lot of places using it.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountViewModel.kt
```diff
@@ -41,7 +38,7 @@ class EnterAmountViewModel @Inject constructor(
 ) : ViewModel() {
     private val _selectedCurrencyCode = MutableStateFlow(configuration.exchangeCurrencyCode)
     var selectedCurrencyCode: String
-        get() = _selectedCurrencyCode.value
+        get() = _selectedCurrencyCode.value!!
```
HashEngineering, line 41: I have a concern here about the exchange rate being null, which has been the case for Krip Phones with locale of Venezuela or perhaps when the currency of the device was set to VEF.  Our app uses VES instead.   

3 months ago or so, we made some changes to prevent crashes from from null currency codes and VEF.  Perhaps there is not an issue with `!!`.  The solutions that I made previously were not robust.  I think we can have QA test for this.

wallet/src/de/schildbach/wallet/ui/main/HeaderBalanceFragment.java
```diff
@@ -0,0 +1,176 @@
...
+import org.dash.wallet.common.ui.CurrencyTextView;
+import org.dash.wallet.common.util.GenericUtils;
+
+import dagger.hilt.android.AndroidEntryPoint;
+import de.schildbach.wallet.Constants;
+import de.schildbach.wallet_test.R;
+
+@AndroidEntryPoint
+public final class HeaderBalanceFragment extends Fragment {
+    private View showBalanceButton;
```
HashEngineering, line 44: This file will need to be "merged" with the java of `dashpay` manually when the time comes.

wallet/src/de/schildbach/wallet/ui/main/WalletActivity.java
```diff
@@ -1,25 +1,25 @@
...
- * Copyright 2020 Dash Core Group
+ * Copyright 2020 Dash Core Group.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
```
HashEngineering, line 7: I wonder how the license changed from GPL3 to Apache 2.0.  In 2013, before any of us were working on this project, this file was using GPL3.

Good catch.

https://github.com/dashevo/dash-wallet/blob/780fbad36d9d321d3103c973bc708497dac8accd/wallet/src/de/schildbach/wallet/ui/WalletActivity.java

HashEngineering (approved):

wallet/src/de/schildbach/wallet/ui/main/WalletTransactionsFragment.java
```diff
@@ -433,12 +440,19 @@ private void updateSyncState(BlockchainState blockchainState) {
...
-            int start = syncing.length() + 1;
-            int end = str.length();
-            str.setSpan(new android.text.style.StyleSpan(android.graphics.Typeface.BOLD), start, end,
-                    Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
-            syncingText.setText(str);
+
+            if (percentage == 0) {
+                syncing += "…";
+            } else {
+                SpannableStringBuilder str = new SpannableStringBuilder(syncing + " " + percentage + "%");
```
HashEngineering, line 447: I wonder if this will render correctly on RTL locales.

wallet/src/de/schildbach/wallet/ui/main/HeaderBalanceFragment.java
HashEngineering: Or maybe not.  It depends on how the move is tracked with git.  In this PR, it is showed as a delete and create, rather than a move.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountViewModel.kt
Syn-McJ: The problem is that `configuration.exchangeCurrencyCode` can return null, but it wasn't transparent for Kotlin code. So I have added a `@NotNull` annotation which then has propagated through here. But I have assumed that by this time we have a code already.

Depending on the location, I guess we can either provide a default value, or wait for changes in `exchangeCurrencyCode` the way I did it in the MainViewModel, and only act after some value is received.

Does it make sense to provide a default value here (I assume it's USD) if `configuration.exchangeCurrencyCode` is null?

wallet/src/de/schildbach/wallet/ui/main/HeaderBalanceFragment.java
Syn-McJ: I have thought about translating this file to Kotlin since it has become quite light after refactoring but decided it might be too many changes already for this PR.

I can handle the next merge if needed. This kind of refactoring that's done with `HeaderBalanceFragment` is what we can also do with `MainActivity` to significantly simplify merge in the future.

wallet/src/de/schildbach/wallet/ui/main/WalletActivity.java
Syn-McJ: That was actually an automatic change Android Studio made when I moved the file :)

wallet/src/de/schildbach/wallet/ui/main/WalletTransactionsFragment.java
Syn-McJ: RTL locales show it the same way - "Syncing 5%" with the percentage on the right, but I don't know if it's supposed to be like this. Should RTL language place the percentage on the left?

Seems like the purpose of this `SpannableStringBuilder` is to make the percentage bold. This should be possible to do in the .xml strings with `<b>` tag. Perhaps we can do that and let the translators decide the arrangement.

HashEngineering: I don't know how the `%` should be displayed on an RTL language.  We can figure this out later.

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountViewModel.kt
HashEngineering: it may make sense, though it would be better that the null was handled earlier inside Configuration.

wallet/src/de/schildbach/wallet/ui/main/WalletTransactionsFragment.java
```diff
@@ -443,16 +443,15 @@ private void updateSyncState(BlockchainState blockchainState) {
...
                 syncing += "…";
+                syncingText.setText(syncing);
             } else {
                 SpannableStringBuilder str = new SpannableStringBuilder(syncing + " " + percentage + "%");
                 int start = syncing.length() + 1;
                 int end = str.length();
                 str.setSpan(new android.text.style.StyleSpan(android.graphics.Typeface.BOLD), start, end,
                         Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
-                syncing = str.toString();
+                syncingText.setText(str);
```
Syn-McJ, line 453: Have to pass `SpannableStringBuilder` to the `setText`, otherwise bold won't work

PR #950, opened 2022-05-09
NMA-1262  Polishing

hadia:
UI changes to polish CrowdNode screens

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/res/layout/dialog_adaptive.xml
```diff
@@ -37,7 +37,8 @@
         android:layout_height="wrap_content"
         android:layout_marginTop="20dp"
         android:gravity="center"
-        android:textAlignment="gravity"
+        android:layout_gravity="center"
+        android:textAlignment="center"
```
Syn-McJ, line 41: This breaks the logic of aligning a message that is longer than 3 lines to the left (right in rtl locales). Is there a reason for this change?

Syn-McJ (commented):
@hadia I dodn't see the changes that are related to some ACs like

- New Account Screen
    - Remove “in”
   -  “To and from CrowdNode”

Both are still there: 
`You will lose access to your funds in within this wallet...`
`All transfers to and from CrowdNode from...`

common/src/main/res/layout/dialog_adaptive.xml
hadia: @Syn-McJ I tried to match new figma design 
image

Syn-McJ: You should be able to achieve that just by removing the icon though? Changing alignment logic isn't necessary.

PR #955, opened 2022-05-11
Fix: disable debug crashlytics & cleanup

Syn-McJ:
We don't need crashlytics reporting in debug mode since it tracks exceptions developers have during their work.
Checked on debug and release builds with a fake exception, only release shows up: https://console.firebase.google.com/u/1/project/dash-wallet-firebase/crashlytics/app/android:hashengineering.darkcoin.wallet_test/issues?time=last-twenty-four-hours&state=open&tag=all

Additionally, I have removed the .travis config file which isn't needed.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Excellent

PR #956, opened 2022-05-12
fix: initializing MainViewModel earlier in WalletActivity

Syn-McJ:
To avoid NRE in the `onCreate` code, the viewModel should be initialized early on.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Thank you for doing this fix.

PR #957, opened 2022-05-12
feat: update reset wallet flow

ClaudeHangui:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
```diff
@@ -44,6 +44,7 @@ class AdaptiveDialog(@LayoutRes private val layout: Int): DialogFragment() {
         private const val MESSAGE_ARG = "message"
         private const val POS_BUTTON_ARG = "positive_text"
         private const val NEG_BUTTON_ARG = "negative_text"
+        private const val EXTRA_MESSAGE_BUTTON_ARG = "extra_message_text"
```
Syn-McJ, line 47: I don't think we should be amending AdaptiveDialog for every possible case. This is a simple class for a simple purpose, if we start adding more stuff to it it'll bloat. 

If you require more complex behavior like additional buttons, you can create another dialog (make sure it's DialogFragment), or even try to derive from this one.

wallet/src/de/schildbach/wallet/ui/SecurityViewModel.kt
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2019 Dash Core Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
Syn-McJ, line 4: Our license is GPLv3

Syn-McJ (changes requested):
@ClaudeHangui could you please try to create a dialog that derives from AdaptiveDialog and extend it that way with the necessary logic.

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialog.kt
HashEngineering: I think this is a good point.  If we change the interface of `AdaptiveDialog`, then we should have QA test all places where `AdaptiveDialog` is used in the app, or at least all the ones affected by the change.  It may be difficult to identify all of those places.

If a different dialog class is used, then that will limit the places that are affected by the change.  We may find that the new class is useful in other places.  There are still many old style dialogs in our app that will need to be replaced eventually.

HashEngineering (changes requested):
Make the changes requested in the other review.  I have some disagreement with the text used, but that is a matter for the designer.

ClaudeHangui:
@HashEngineering could you please comment on the figma design as far as the texts are concerned ?

Syn-McJ (approved):
Looks great

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialogExt.kt
```diff
@@ -0,0 +1,98 @@
...
+
+import android.os.Bundle
+import android.view.View
+import android.widget.TextView
+import androidx.annotation.DrawableRes
+import androidx.annotation.LayoutRes
+import androidx.fragment.app.FragmentActivity
+import org.dash.wallet.common.R
+
+class AdaptiveDialogExt(@LayoutRes private val layout: Int) : AdaptiveDialog(layout) {
```
HashEngineering, line 28: `layout` should have a default value to avoid a crash from recreating this dialog

HashEngineering (changes requested):
there is one change to be made.

Also, this will be merged after 7.4.9 is completed on the master branch.

common/src/main/java/org/dash/wallet/common/ui/dialogs/AdaptiveDialogExt.kt
ClaudeHangui: @HashEngineering done!

PR #958, opened 2022-05-15
[NMA-1265] (CrowdNode) Verifying API address

Syn-McJ:
Before the user can deposit and withdraw using their API address, it needs to be validated and verified.

- Validation is done by CrowdNode, we only query the address status and check if it's done.
- Verification is done partially by the user - they need to transfer a small amount of DASH from the primary to the API address. We detect and re-send it to CrowdNode automatically.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -100,6 +100,10 @@ public class Configuration {
     public static final String PREFS_KEY_HAS_LOCATION_DIALOG_BEEN_SHOWN = "has_location_dialog_been_shown";
     public static final String PREFS_KEY_EXPLORE_DATABASE_NAME = "explore_database_name";
 
+    // CrowdNode
+    public static final String PREFS_KEY_CROWDNODE_ACCOUNT_ADDRESS = "crowdnode_account_address";
+    public static final String PREFS_KEY_CROWDNODE_PRIMARY_ADDRESS = "crowdnode_primary_address";
```
Syn-McJ, line 105: These are needed in the global configuration file for easy access from the `BlockhchainService`. I figure we might either keep using this pattern (mostly local configs for the feature modules with a couple of settings in the global one) or, if the `BlockchainService` is ever moved to kotlin, we can use injected local configs in it directly.

HashEngineering (approved):
Looks great.

PR #961, opened 2022-05-23
feat: update transaction details ui

ClaudeHangui:
- update transaction details UI
- provide new fonts
- 

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/res/values/styles.xml
+++ b/common/src/main/res/values/styles.xml
@@ -324,18 +324,6 @@
         <item name="android:textSize">14sp</item>
     </style>
 
-    <style name="InterRegular">
-        <item name="fontFamily">@font/inter_regular</item>
-    </style>
-
-    <style name="InterRegular.MG">
-        <item name="android:textColor">@color/light_gray_black</item>
-    </style>
-
-    <style name="InterRegular.MG.32">
-        <item name="android:textSize">32sp</item>
-    </style>
-
     <!--style name="My.Theme" parent="@android:style/Theme.Holo.Light.DarkActionBar">
         <item name="android:actionBarStyle">@style/My.Widget.ActionBar</item>
         <item name="android:actionModeStyle">@style/My.Widget.ActionMode</item>
@@ -530,14 +518,6 @@
         <item name="android:textColor">@drawable/blue_button_text</item>
         <item name="android:background">@drawable/blue_button_background</item>
         <item name="android:textSize">14sp</item>
-        <item name="android:textAllCaps">false</item>
-    </style>
-
-    <style name="DashButton.Blue.InterRegular" parent="Widget.AppCompat.Button.Borderless">
-        <item name="fontFamily">@font/inter_regular</item>
-        <item name="android:textColor">@drawable/blue_button_text</item>
-        <item name="android:background">@drawable/blue_button_background</item>
-        <item name="android:textSize">15sp</item>
         <item name="android:textAllCaps">false</item>
     </style>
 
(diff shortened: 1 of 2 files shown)


common/src/main/res/values/styles.xml
```diff
@@ -324,6 +324,18 @@
...
+        <item name="fontFamily">@font/inter_regular</item>
+    </style>
+
+    <style name="InterRegular.MG">
+        <item name="android:textColor">@color/light_gray_black</item>
+    </style>
+
+    <style name="InterRegular.MG.32">
+        <item name="android:textSize">32sp</item>
+    </style>
```
HashEngineering, line 337: Should we be using a different naming system for Styles that doesn't use the name of the font?

HashEngineering: Also, I am wondering if the requirement of a different font is a deliberate choice by the designer.

wallet/res/values/strings.xml
```diff
@@ -393,4 +393,6 @@
     <string name="activity_buy_and_sell_dash_filter">Filter</string>
     <string name="clear_filter">Clear Filter</string>
     <string name="select_currency">Select Currency</string>
+    <string name="reason">Reason</string>
+    <string name="resolve_txt">Double spend • This could be solved by</string>
```
HashEngineering, line 397: Not sure how this dot will get translated into other languages.

HashEngineering (changes requested):

common/src/main/res/values/styles.xml
Syn-McJ: This is correct, we should not be using styles with the font name in them. That's the whole point of the new typography styles - so that we can simply switch the font later and not have to do any other changes.

@ClaudeHangui I think you can continue using Montserrat for your story - we have a different one specifically for changing fonts throughout the entire app.

wallet/res/values/strings.xml
ClaudeHangui: this resource value shouldn't be here, let me remove it

wallet/src/de/schildbach/wallet/ui/TransactionDetailsDialogFragment.kt
```diff
@@ -44,45 +58,66 @@ class TransactionDetailsDialogFragment : DialogFragment() {
...
+        val animSet = AnimationSet(false)
+        animSet.apply {
+            addAnimation(containerAnimation)
+            addAnimation(contentAnimation)
+        }
+
+        transaction_details_dialog_container.postDelayed({
             transaction_details_dialog_content_container.visibility = View.VISIBLE
-            transaction_details_dialog_content_container.startAnimation(contentAnimation)
+            transaction_details_dialog_content_container.startAnimation(animSet)
```
Syn-McJ, line 120: I still see jerking in the animation. Most likely it's because `slide_in_bottom` doesn't actually start from the container being on the bottom, so it appears at `translation=0` for a moment. 
Also, `slide_in_bottom` already has fade-in animation itself. So you can fix this all like this (or use `container.animate()` extension method):
```suggestion
        val container = transaction_details_dialog_content_container
        val contentAnimation = AnimationUtils.loadAnimation(activity, R.anim.slide_in_bottom)
        container.translationY = container.measuredHeight.toFloat()
        container.visibility = View.VISIBLE
        container.startAnimation(contentAnimation)
```

Syn-McJ (changes requested):

Syn-McJ (approved):
Looks good

ClaudeHangui (commented):
Great, thanks !

Newman1515 (changes requested):
Submit the change

HashEngineering (approved):
The fonts and related changes are still included.

Syn-McJ:
@HashEngineering I've cleaned up the styles we don't need. The fonts will be used in the future, so we might want to keep them.

PR #962, opened 2022-06-01
[NMA-1251] (CrowdNode) Link a new Online account

Syn-McJ:
For the accounts newly created via blockchain API, we would like the ability to create an online account too.

Apart from the required CrowdNode functionality, I did some refactoring to start addressing a couple of issues with the `WalletApplication`:
- Our `WalletApplication` class is quite big and it is handling a lot of different things. It'd be nice to extract some of that responsibility into other classes.
- The `WalletApplication` class is getting passed around everywhere which gives a lot of places access to this "god-class" and potentially creates memory leaks.

Most of the users of the `WalletApplication` need it for accessing the `Wallet` object. I figure we better have separate storage for this object which then will be accessed by users and updated by the `WalletApplication`. It isn't done yet in this PR, but some work is started.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/CheckPinDialog.kt
```diff
@@ -104,23 +107,21 @@ open class CheckPinDialog(
     }
 
     private val binding by viewBinding(FragmentEnterPinBinding::bind)
+    protected open val sharedModel by activityViewModels<CheckPinSharedModel>()
```
Syn-McJ, line 110: `SharedViewModel` is unnecessary with callbacks and better to be got rid of later.

wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
```diff
@@ -30,8 +29,11 @@ class DecryptSeedWithPinDialog : CheckPinDialog() {
         @JvmStatic
         fun show(activity: AppCompatActivity, requestCode: Int = 0, pinOnly: Boolean = false) {
             val checkPinDialog = DecryptSeedWithPinDialog()
-            if (PinRetryController.getInstance().isLocked) {
-                checkPinDialog.showLockedAlert(activity)
```
Syn-McJ, line 34: Calling fragment methods from another object (a `companion object` in this case) is a bit dangerous. The fragment might not be attached and this might cause errors. In the past, we had an issue where in this method the `context` of the fragment was accessed instead of the passed context. I then had an issue with accessing the `viewModel` from that method.

I'm not sure if it's better to get strings beforehand and pass them to the method, or to move the dialog invocation out of there entirely. I've settled for the first option for now.

common/src/main/res/values/colors.xml
```diff
@@ -81,6 +81,11 @@
     <color name="text_highlight_blue_bg">#ebf6fd</color>
     <color name="ultra_light_gray">#f5f6f7</color>
     <color name="warning_yellow">#fff9ed</color>
+    <color name="input_error_background">#1AE85C4A</color>
+
+    <!-- Stokes -->
+    <color name="divider_color">@color/gray_50</color>
+    <color name="input_focused_stoke">#cce8fa</color>
```
HashEngineering, line 88: Should this be Stokes or Strokes?  I see that both words have some relation to art.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/ui/portal/TransferFragment.kt
```diff
@@ -62,7 +62,7 @@ class TransferFragment : Fragment(R.layout.fragment_transfer) {
     private var balanceAnimator: ObjectAnimator? = null
 
     @Inject
-    lateinit var securityModel: SecurityModel
+    lateinit var ISecurityFunctions: ISecurityFunctions
```
HashEngineering, line 65: Was `securityModel` supposed to be renamed as `ISecurityModel` or was it a Find/Release without being case sensitive?

common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
```diff
@@ -30,6 +30,7 @@ import org.dash.wallet.common.transactions.TransactionFilter
 import org.dash.wallet.common.transactions.TransactionWrapper
 
 interface WalletDataProvider {
+    val wallet: Wallet?
```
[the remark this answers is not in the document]
HashEngineering: The original idea behind `WalletDataProvider` was to give other modules some data without providing the Wallet itself.  Probably some of the reason for this was that third parties may be writing these integrations (which was done in the case of Liquid).  The `Wallet` object would be fully owned by the `wallet` module (which includes the means to decrypt the keys).

HashEngineering (changes requested):
There are a few questions.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/ui/portal/TransferFragment.kt
Syn-McJ: Yeah that is the case, thanks for noticing

common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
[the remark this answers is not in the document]
Syn-McJ: Yes, this is a correct idea and we should keep to it (don't use `wallet` object in the feature modules). Why I have put it in there for now:
- We have a lot of places that use the `WalletApplication` while they shouldn't. This is a god-object that does too much and provides a lot of access to a lot of functions. It's too tempting to call `getInstance()` on it and break all of the DI patterns we have added. Just now I had to deal with this problem while writing tests for the Explore database, even if injections were used there for some other things.
- Most places only need the `wallet`, not the entire `WalletApplication`.
- We need to abstract the access to the `wallet` from the `WalletApplication` somehow. I think the best option here would be to have some kind of "wallet storage" class that is getting updated by the `WalletApplication`. This class would only be available from the `wallet` module.
- Until this solution exists (I'm not 100% sure yet on what it should look like), we need to keep abstracting the access to `wallet` from the `WalletApplication`. I suppose we could create yet another interface for it, but I fear this could be even more confusing.

So for now I figure we can have the `wallet` in this interface, and once the access to it is moved from the `WalletApplication`, we will simply remove it from here and replace it with another injection which should be relatively easy to do.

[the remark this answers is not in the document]
Syn-McJ: I've added a comment warning from using it in the feature modules.

HashEngineering (approved):

PR #967, opened 2022-06-13
fix: autologout in Exchange integrations (NMA-1307)

HashEngineering:

After entering the Liquid or Uphold screens, the autologout timer was accelerated.

Each time `startTimer` was called, a new timer was created.  The fix is to prevent creating a new timer if one is active.

None
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):
Looks good

PR #968, opened 2022-06-13
fix: crowdnode feedback (NMA-1300)

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
+++ b/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
@@ -254,8 +254,10 @@ class CrowdNodeApiAggregator @Inject constructor(
 
             responseScope.launch {
                 try {
-                    val tx = blockchainApi.waitForWithdrawalResponse(requestValue)
-                    log.info("got withdrawal queue response: ${tx.txId}")
+                    val txResponse = blockchainApi.waitForWithdrawalResponse(requestValue)
+                    log.info("got withdrawal queue response: ${txResponse.txId}")
+                    val txWithdrawal = blockchainApi.waitForWithdrawalReceived()
+                    log.info("got withdrawal: ${txWithdrawal.txId}")
                 } catch (ex: Exception) {
                     handleError(ex, appContext.getString(R.string.crowdnode_withdraw_error))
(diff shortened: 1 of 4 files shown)


integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
```diff
@@ -258,6 +259,7 @@ class CrowdNodeApiAggregator @Inject constructor(
                 } catch (ex: Exception) {
                     handleError(ex, appContext.getString(R.string.crowdnode_withdraw_error))
                 }
+                refreshBalance(retries = 3)
```
HashEngineering, line 262: This doesn't catch the withdrawal transaction that has the amount specified by the user.

Syn-McJ: Right, we can track `CrowdNodeWithdrawalReceivedTx` here as well for a more robust solution. See DMs

Syn-McJ (changes requested):
Looks good, but some code for the balance after withdrawal is better to be added

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
```diff
@@ -293,8 +297,14 @@ class CrowdNodeApiAggregator @Inject constructor(
...
-                    // balance changed, no need to retry anymore
-                    break
+                    val minimumWithdrawal = CrowdNodeConstants.API_OFFSET + Coin.valueOf(ApiCode.MaxCode.code)
+                    if (!afterWithdrawal) {
+                        // balance changed, no need to retry anymore
+                        break;
+                    } else if (lastBalance - (currentBalance.data?.value?: 0L) >= minimumWithdrawal.value) {
+                        // balance changed, no need to retry anymore
+                        break
+                    }
```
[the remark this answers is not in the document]
Syn-McJ: Makes sense. Perhaps we should remove the waiting for 20016 in the future since it's not really used for anything.

Syn-McJ (approved):
Looks great

PR #969, opened 2022-06-14
[NMA-1303 && NMA-1302] (CrowdNode) Low balance warning and events

Syn-McJ:
If the user makes an outgoing transaction, we want to show a warning if the balance will drop lower than the amount necessary for withdrawal.
Also, we need events for CrowdNode.

There are 2 places through which transactions are going and where it is best to make this check: `SendCoinsTaskRunner` and `SendCoinsOfflineTask`.
In the first case, we make the check and throw an exception which then is up to the caller to handle and call again if the user clicked "Continue".

In the second case, the flow is a bit complicated with all the callbacks and overridden methods, so I implemented it in a robust but hacky way which is better to be refactored later. I imagine we'll have do it anyway when redesigning the Send screen.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -863,6 +888,7 @@ public void shutdownAndDeleteWallet() {
 
     public void finalizeWipe() {
         cancelScheduledStartBlockchainService();
+        WorkManager.getInstance(this.getApplicationContext()).cancelAllWork();
```
Syn-McJ, line 891: There was a bug when after the wipe the CrowdNode signup worker would still persist.
In the case of the Explore worker, we should cancel it too. Given that during the wipe all of the Explore settings are cleared, it can lead to undefined behavior of the sync job.

HashEngineering (approved):
Looks great.

PR #978, opened 2022-06-24
[NMA-1312] (CrowdNode) Fix for the invalid online state

Syn-McJ:
There is a problem with 7.5.0 (beta) -> 7.5.1 upgrade scenario: because of the changes in ApiStatuses, the old online status is restored into a wrong value.

Added a check for this specific case + some tests.
The base `bugfix-crowdnode-7.5.2` branch is checked out of 7.5.1 tag. We can release it as a hotfix with some other possible fixes.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/integrations/crowdnode/src/main/res/layout/fragment_portal.xml
+++ b/integrations/crowdnode/src/main/res/layout/fragment_portal.xml
@@ -113,7 +113,7 @@
                 android:id="@+id/deposit_btn"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:foreground="?attr/selectableItemBackground"
+                android:background="?attr/selectableItemBackground"
                 android:clickable="true"
                 android:focusable="true"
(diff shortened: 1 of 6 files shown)


integrations/crowdnode/src/main/res/layout/fragment_portal.xml
```diff
@@ -113,7 +113,7 @@
                 android:id="@+id/deposit_btn"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:foreground="?attr/selectableItemBackground"
+                android:background="?attr/selectableItemBackground"
```
Syn-McJ, line 116: There is also a small UI issue with this button because of the `foreground` property. It was used to make a selected button look good when there is the "deposit more.." badge present.

Now the "deposit more..." badge has alpha in its color and so we can change this to `background` and it'll look fine.

PR #982, opened 2022-06-30
feat: set transfer defaults (NMA-1270)

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/ui/TransactionMetadataViewModel.kt
+++ b/common/src/main/java/org/dash/wallet/common/ui/TransactionMetadataViewModel.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2020 Dash Core Group.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.dash.wallet.common.ui
+
+import androidx.lifecycle.*
+import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.launch
+import org.dash.wallet.common.services.TransactionMetadataProvider
+import org.dash.wallet.common.transactions.TaxCategory
+import javax.inject.Inject
+
+@HiltViewModel
+class TransactionMetadataViewModel @Inject constructor (private val transactionMetadataProvider: TransactionMetadataProvider) : ViewModel() {
+
+    private fun markAddressAsync(address: String, sendTo: Boolean, taxCategory: TaxCategory) {
+        viewModelScope.launch {
+            transactionMetadataProvider.markAddressWithTaxCategory(address, sendTo, taxCategory)
+        }
+    }
+    fun markAddressAsTransferOutAsync(address: String) {
+        markAddressAsync(address, true, TaxCategory.TransferOut)
+    }
+
+    fun markAddressAsTransferInAsync(address: String) {
+        markAddressAsync(address, false, TaxCategory.TransferIn)
+    }
+}
--- a/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
+++ b/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
@@ -36,7 +36,9 @@ import org.dash.wallet.common.data.Resource
 import org.dash.wallet.common.services.ISecurityFunctions
 import org.dash.wallet.common.services.LeftoverBalanceException
 import org.dash.wallet.common.services.NotificationService
+import org.dash.wallet.common.services.TransactionMetadataProvider
 import org.dash.wallet.common.services.analytics.AnalyticsService
+import org.dash.wallet.common.transactions.TaxCategory
 import org.dash.wallet.common.util.TickerFlow
 import org.dash.wallet.integrations.crowdnode.R
 import org.dash.wallet.integrations.crowdnode.model.*
@@ -90,6 +92,7 @@ class CrowdNodeApiAggregator @Inject constructor(
     private val config: CrowdNodeConfig,
     private val globalConfig: Configuration,
     private val securityFunctions: ISecurityFunctions,
+    private val transactionMetadataProvider: TransactionMetadataProvider,
     @ApplicationContext private val appContext: Context
 ): CrowdNodeApi {
     companion object {
@@ -491,7 +494,11 @@ class CrowdNodeApiAggregator @Inject constructor(
         fullSignUpSet?.let { set ->
             accountAddress = set.accountAddress
             requireNotNull(accountAddress) { "Restored signup tx set but address is null" }
-            configScope.launch { globalConfig.crowdNodeAccountAddress = accountAddress!!.toBase58() }
+            configScope.launch {
+                globalConfig.crowdNodeAccountAddress = accountAddress!!.toBase58()
+                transactionMetadataProvider.maybeMarkAddressWithTaxCategory(accountAddress!!.toBase58(), false, TaxCategory.TransferIn)
+                transactionMetadataProvider.maybeMarkAddressWithTaxCategory(accountAddress!!.toBase58(), true, TaxCategory.TransferOut)
+            }
 
             if (set.hasWelcomeToApiResponse) {
                 log.info("found finished sign up, account: ${set.accountAddress?.toBase58() ?: "null"}")
@@ -581,6 +588,8 @@ class CrowdNodeApiAggregator @Inject constructor(
                 accountAddress = address
                 globalConfig.crowdNodeAccountAddress = address.toBase58()
                 globalConfig.crowdNodePrimaryAddress = primary.toBase58()
+                transactionMetadataProvider.maybeMarkAddressWithTaxCategory(address.toBase58(), false, TaxCategory.TransferIn)
+                transactionMetadataProvider.maybeMarkAddressWithTaxCategory(address.toBase58(), true, TaxCategory.TransferOut)
                 changeOnlineStatus(OnlineAccountStatus.Validating)
             }
         }
--- a/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeBlockchainApi.kt
+++ b/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeBlockchainApi.kt
@@ -24,9 +24,11 @@ import org.bitcoinj.core.Transaction
 import org.dash.wallet.common.WalletDataProvider
 import org.dash.wallet.common.services.LeftoverBalanceException
 import org.dash.wallet.common.services.SendPaymentService
+import org.dash.wallet.common.services.TransactionMetadataProvider
 import org.dash.wallet.common.transactions.ByAddressCoinSelector
 import org.dash.wallet.common.transactions.ExactOutputsSelector
 import org.dash.wallet.common.transactions.LockedTransaction
+import org.dash.wallet.common.transactions.TaxCategory
 import org.dash.wallet.integrations.crowdnode.model.CrowdNodeException
 import org.dash.wallet.integrations.crowdnode.transactions.*
 import org.dash.wallet.integrations.crowdnode.utils.CrowdNodeConstants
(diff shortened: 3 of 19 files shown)


common/src/main/java/org/dash/wallet/common/ui/TransactionMetadataViewModel.kt
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+    fun markAddressAsTransferOutAsync(address: String) {
+        markAddressAsync(address, true, TaxCategory.TransferOut)
+    }
+
+    fun markAddressAsTransferInAsync(address: String) {
+        markAddressAsync(address, false, TaxCategory.TransferIn)
+    }
+}
```
HashEngineering, line 42: After making this class, I was wondering if it is better to simply Inject the `TransactionMetadataProvider` into the activities that need it and support the same function calls.  `TransactionMetadataProvider` would handle the co-routines.

This class is currently used by Uphold and Liquid Activities, which do not have much or any ViewModel support.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
```diff
@@ -491,7 +494,11 @@ class CrowdNodeApiAggregator @Inject constructor(
         fullSignUpSet?.let { set ->
             accountAddress = set.accountAddress
             requireNotNull(accountAddress) { "Restored signup tx set but address is null" }
-            configScope.launch { globalConfig.crowdNodeAccountAddress = accountAddress!!.toBase58() }
+            configScope.launch {
+                globalConfig.crowdNodeAccountAddress = accountAddress!!.toBase58()
+                transactionMetadataProvider.maybeMarkAddressWithTaxCategory(accountAddress!!.toBase58(), false, TaxCategory.TransferIn)
+                transactionMetadataProvider.maybeMarkAddressWithTaxCategory(accountAddress!!.toBase58(), true, TaxCategory.TransferOut)
+            }
```
HashEngineering, line 501: anytime that the `crowdNodeAccountAddress` is saved to the configuration, this address is saved to AddressMetadata table.

HashEngineering: 3 places were found where this happens, so each was marked.

integrations/liquid/src/main/java/org/dash/wallet/integration/liquid/ui/BuyDashWithCreditCardActivity.kt
```diff
@@ -191,7 +196,6 @@ class BuyDashWithCreditCardActivity : InteractionAwareActivity() {
 
 
         userAmount = intent.getStringExtra("Amount")
-        val walletDataProvider = application as WalletDataProvider
```
HashEngineering, line 194: using an injected walletDataProvider instead.

integrations/liquid/src/main/java/org/dash/wallet/integration/liquid/ui/BuyDashWithCreditCardActivity.kt
```diff
@@ -152,7 +157,7 @@ class BuyDashWithCreditCardActivity : InteractionAwareActivity() {
     }
 
     fun initViewModel() {
-        viewModel = ViewModelProvider(this)[ConnectivityViewModel::class.java]
+        viewModel = ViewModelProvider(this)[LiquidViewModel::class.java]
```
HashEngineering, line 160: don't need this change because LiquidViewModel wasn't used in this PR.

integrations/liquid/src/main/java/org/dash/wallet/integration/liquid/ui/BuyDashWithCreditCardActivity.kt
```diff
@@ -482,6 +486,7 @@ class BuyDashWithCreditCardActivity : InteractionAwareActivity() {
...
                                 "success" -> {
                                     log.info("liquid: buy dash transaction successful")
+                                    transactionMetadataViewModel.markAddressAsTransferInAsync(walletAddress)
```
HashEngineering, line 489: Here the transaction address is marked upon success, though what happens if the app crashes before "success" is sent to the widget.  Maybe we should mark the address earlier, after the fresh address is generated?

wallet/schemas/de.schildbach.wallet.AppDatabase/11.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 11,
-    "identityHash": "72542834dc65de042a7b899f254c7af5",
+    "identityHash": "d38e40798885331e7ba0b601088d0f65",
```
HashEngineering, line 5: this might look strange, but version 11 never went into production.

wallet/src/de/schildbach/wallet/data/AddressMetadataDao.kt
```diff
@@ -0,0 +1,53 @@
...
+    suspend fun exists(address: String, isInput: Boolean): Boolean
+
+    @Query("SELECT * FROM address_metadata WHERE address = :address AND isInput = 0")
+    suspend fun loadRecipient(address: String): AddressMetadata?
+
+    @Query("SELECT * FROM address_metadata WHERE address = :address AND isInput = 1")
+    suspend fun loadSender(address: String): AddressMetadata?
+
+    @Query("INSERT into address_metadata (address, isInput, taxCategory) VALUES (:address, :isInput, :taxCategory)")
+    suspend fun markAddress(address: String, isInput: Boolean, taxCategory: TaxCategory)
```
HashEngineering, line 46: this duplicates the `insert` function above and probably fails if there is already an entry.  Theoretically, the same address wouldn't be used twice.

HashEngineering: Some of these functions may not be used and could be removed when before the PR is closed out.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -16,30 +16,43 @@
...
     private val walletData: WalletDataProvider
 ) : TransactionMetadataProvider {
 
     companion object {
         val log = LoggerFactory.getLogger(WalletTransactionMetadataProvider::class.java)
     }
 
+    private val syncScope = CoroutineScope(
+        Executors.newCachedThreadPool().asCoroutineDispatcher()
+    )
```
HashEngineering, line 54: During the blockchain sync, if there are many transactions this scope, could this result in creating countless threads?  Would it be better to use a fixed pool size or some other scope?

The idea was taken from the CrowdNode module.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -102,20 +116,185 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
+                            ScriptPattern.isP2PKH(output.scriptPubKey) ->
+                                Address.fromPubKeyHash(
+                                    walletData.networkParameters,
+                                    ScriptPattern.extractHashFromP2PKH(output.scriptPubKey)
+                                )
+                            ScriptPattern.isP2SH(output.scriptPubKey) ->
+                                Address.fromScriptHash(
+                                    walletData.networkParameters,
+                                    ScriptPattern.extractHashFromP2SH(output.scriptPubKey)
+                                )
```
HashEngineering, line 196: this probably would never be triggered because our wallets don't support P2SH addresses internally. 

Our wallet can send to P2SH addresses, which is why we can check for this below.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -102,20 +116,185 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
+    @OptIn(ExperimentalCoroutinesApi::class)
     override fun observeTransactionMetadata(txId: Sha256Hash): Flow<TransactionMetadata?> {
-        return transactionMetadataDao.observe(txId)
+        return transactionMetadataDao.observe(txId).map { transactionMetadata ->
+            // if there is no user specified tax category, then look at address_metadata
+            if (transactionMetadata != null && transactionMetadata.taxCategory == null) {
+                transactionMetadata.taxCategory = getDefaultTaxCategory(txId)
+            }
+            transactionMetadata
+        }
```
HashEngineering, line 291: as in the other functions in this class, before returning TransactionMetadata, we need to check the address table if no tax category is specified.

Currently, transactionMetadataDao is only storing user-specified tax categories.  If the user has not chosen a tax category, then a default value is taken from the addressMetadataDao table if the transaction has addresses that line up with that table.

I did wonder if it would be better to save this default in the transaction metadata table.

wallet/src/de/schildbach/wallet/ui/UpholdTransferActivity.kt
```diff
@@ -155,6 +158,7 @@ class UpholdTransferActivity : InteractionAwareActivity() {
             }
 
             override fun onTransfer() {
+                transactionMetadataViewModel.markAddressAsTransferInAsync(receiveAddress.toBase58())
```
HashEngineering, line 161: maybe it is better, as stated with the Liquid code, that this should be done after freshAddress is called.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -102,20 +116,185 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
+    }
+
+    override fun clear() {
+        syncScope.launch {
+            transactionMetadataDao.clear()
+            addressMetadataDao.clear()
```
Syn-McJ, line 297: Here "run and forget" function seems appropriate.

wallet/src/de/schildbach/wallet/ui/UpholdTransferActivity.kt
Syn-McJ: Is it possible that the created address isn't used?

common/src/main/java/org/dash/wallet/common/ui/TransactionMetadataViewModel.kt
Syn-McJ: I have just commented on this :) Yes, this class is unnecessary.

integrations/liquid/src/main/java/org/dash/wallet/integration/liquid/ui/BuyDashWithCreditCardActivity.kt
Syn-McJ: In this case it might make sense

Syn-McJ:
@HashEngineering some comments left, but nothing too critical.
The test needs an additional `mock`, failing CI now.

wallet/src/de/schildbach/wallet/ui/UpholdTransferActivity.kt
HashEngineering: Yes.  If the user does not finish the transfer or if the transfer fails due to an error from Uphold.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeBlockchainApi.kt
```diff
@@ -65,6 +68,8 @@ open class CrowdNodeBlockchainApi @Inject constructor(
             throw CrowdNodeException("SignUp request returned an error")
         }
 
+        transactionMetadataProvider.markAddressWithTaxCategory(accountAddress.toBase58(), true, TaxCategory.TransferOut)
+        transactionMetadataProvider.markAddressWithTaxCategory(accountAddress.toBase58(), false, TaxCategory.TransferIn)
```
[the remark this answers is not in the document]
HashEngineering: Great idea.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -99,23 +113,222 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
+
+    override fun syncTransactionBlocking(tx: Transaction) {
+        runBlocking {
+            syncTransaction(tx)
+        }
+    }
+
+    override suspend fun syncTransaction(tx: Transaction) {
+        log.info("sync transaction metadata: ${tx.txId}")
+        val metadata = transactionMetadataDao.load(tx.txId)
```
HashEngineering, line 135: Now there are two of these functions.  The first can be called from BlockchainServiceImpl in a dedicated handler.

The second is called by the first, but may at a later point in development be called by a worker that will gather historical transaction metadata (exchange rates).

Not sure if Blocking at the end of the name is the best nomenclature, perhaps it could be "Sync" though there is already sync in the name.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -102,20 +116,185 @@ class WalletTransactionMetadataProvider @Inject constructor(
         TODO("Not yet implemented")
     }
 
+    // should this be running in a separate thread each time, or the same as the caller?
```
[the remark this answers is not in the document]
HashEngineering: I have made a change following this:

> Since this one is run by the BlockchainService which can't handle suspend methods, we can expose a sync method in a runBlocking wrapper instead. Then the BlockchainService can run it in a dedicated handler or however else it wishes.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
Syn-McJ: yeah Blocking is fine

Syn-McJ (approved):
Looks great

Syn-McJ (approved):

PR #987, opened 2022-07-05
[NMA-1317] fix for detecting CrowdNode transactions

Syn-McJ:
Changing detection logic for CrowdNode responses to fix some sign-up issues.
We don't know if we're gonna go ahead with this solution yet.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

### DIFF
--- a/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
+++ b/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
@@ -29,6 +29,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
 import org.bitcoinj.core.Address
 import org.bitcoinj.core.Coin
+import org.bitcoinj.core.Transaction
 import org.dash.wallet.common.Configuration
 import org.dash.wallet.common.Constants
 import org.dash.wallet.common.WalletDataProvider
@@ -40,6 +41,8 @@ import org.dash.wallet.common.services.analytics.AnalyticsService
 import org.dash.wallet.common.util.TickerFlow
 import org.dash.wallet.integrations.crowdnode.R
 import org.dash.wallet.integrations.crowdnode.model.*
+import org.dash.wallet.integrations.crowdnode.transactions.CrowdNodeAcceptTermsResponse
+import org.dash.wallet.integrations.crowdnode.transactions.CrowdNodeWelcomeToApiResponse
 import org.dash.wallet.integrations.crowdnode.utils.CrowdNodeConfig
 import org.dash.wallet.integrations.crowdnode.utils.CrowdNodeConstants
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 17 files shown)


integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeBlockchainApi.kt
```diff
@@ -75,8 +76,9 @@ open class CrowdNodeBlockchainApi @Inject constructor(
         val acceptTx = paymentService.sendCoins(crowdNodeAddress, requestValue, selector)
         log.info("acceptTx id: ${acceptTx.txId}")
         val errorResponse = CrowdNodeErrorResponse(params, requestValue)
-        val tx = walletDataProvider.observeTransactions(
+        val tx = walletData.observeTransactions(
             CrowdNodeWelcomeToApiResponse(params),
+            PossibleWelcomeResponse(walletData.transactionBag, accountAddress),
```
Syn-McJ, line 81: It's easier to add these as separate filters.
If/when we implement the better solution, `possible` filters will go away from here, but probably will have to stay in other places since some users might have valid accounts that are only detected with these filters.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/transactions/FullCrowdNodeSignUpTxSet.kt
```diff
@@ -32,20 +32,54 @@ open class FullCrowdNodeSignUpTxSet(
...
-        get() = matchedFilters.any { it is CrowdNodeWelcomeToApiResponse }
+        get() = matchedFilters.any {
+            it is CrowdNodeWelcomeToApiResponse ||
+                    (it is PossibleWelcomeResponse && didSignUpFromAddress(it.toAddress))
+        }
 
     open val accountAddress: Address?
-        get() = (matchedFilters.firstOrNull { it is CrowdNodeSignUpTx } as? CrowdNodeSignUpTx)?.fromAddresses?.first()
+        get() {
+            matchedFilters.firstOrNull { it is CrowdNodeWelcomeToApiResponse }?.let {
```
Syn-McJ, line 57: Added a better account detection logic that priorities responses with full info about connected transactions.

integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeApi.kt
```diff
@@ -125,7 +128,6 @@ class CrowdNodeApiAggregator @Inject constructor(
     override var showNotificationOnResult = false
 
     init {
-        restoreStatus()
```
Syn-McJ, line 128: Not that we call this method from `StakingActivity` anyway, it can be made `suspend` to account for fetching `GetFunds` and removed from the ctor.

wallet/res/layout/activity_staking.xml
```diff
@@ -23,6 +23,12 @@
     android:background="@color/background_primary"
     tools:context="de.schildbach.wallet.ui.staking.StakingActivity">
 
+    <ProgressBar
+        android:id="@+id/progress_bar"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center" />
```
Syn-McJ, line 30: Shows up on the `StakingActivity` before navigation is set in case we need to fetch `GetFunds` first.

HashEngineering (approved):
Excellent

PR #990, opened 2022-07-08
fix: reverting transaction total amount + fee changes

Syn-McJ:
During the development of the new Transaction Details screen, the logic for displaying the total amount was changed to show the amount only.
We want to revert that to show the amount + fee as it is now in production.
CrowdNode total amount calculation will also include fees.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/transactions/TransactionResultViewBinder.kt
+++ b/wallet/src/de/schildbach/wallet/ui/transactions/TransactionResultViewBinder.kt
@@ -161,16 +161,10 @@ class TransactionResultViewBinder(
 
         dashAmount.setFormat(dashFormat)
         //For displaying purposes only
-        val amount = if (isSent && isFeeAvailable(tx.fee)) {
-            value.plus(tx.fee)
+        if (value.isNegative) {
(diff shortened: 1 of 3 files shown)


wallet/src/de/schildbach/wallet/ui/transactions/TransactionResultViewBinder.kt
```diff
@@ -221,7 +215,7 @@ class TransactionResultViewBinder(
                 transactionTitle.setTextColor(ContextCompat.getColor(ctx, R.color.dash_blue))
                 transactionTitle.text = ctx.getText(R.string.transaction_details_amount_sent)
                 transactionAmountSignal.text = "-"
-                transactionAmountSignal.isVisible = !TransactionUtils.isEntirelySelf(tx, wallet)
+                transactionAmountSignal.isVisible = true
```
Syn-McJ, line 218: The fact that we want to include the fee into the total amount kinda clashes with our idea of removing the signal from internal transactions (it'll look as if we had received the fee amount).
For now, let's just keep it the way it is in production currently.

HashEngineering (approved):

PR #993, opened 2022-07-09
feat: add tx metadata support (NMA-1321)

HashEngineering:

Add minimal tx metadata support
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/androidTest/de/schildbach/wallet/database/DatabaseMigrationTest.kt
+++ b/wallet/androidTest/de/schildbach/wallet/database/DatabaseMigrationTest.kt
@@ -24,8 +24,14 @@ import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import de.schildbach.wallet.AppDatabase
-import de.schildbach.wallet.data.AppDatabaseMigrations.Companion.migration8To10
+import de.schildbach.wallet.data.AppDatabaseMigrations
 import kotlinx.coroutines.runBlocking
+import org.bitcoinj.core.Coin
+import org.bitcoinj.core.Sha256Hash
+import org.dash.wallet.common.data.ServiceName
+import org.dash.wallet.common.data.TaxCategory
+import org.dash.wallet.common.data.TransactionMetadata
+import org.dash.wallet.common.transactions.TransactionCategory
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -38,9 +44,23 @@ open class DatabaseMigrationTest {
         private const val TEST_DB_NAME = "test_database"
         private const val BLOCKCHAIN_HEIGHT = 587680
         private const val EXCHANGE_RATE = "31438.8212"
+
+        private val txMetadata = TransactionMetadata(
+            Sha256Hash.ZERO_HASH,
+            System.currentTimeMillis() / 1000,
+            Coin.CENT,
+            TransactionCategory.Sent,
+            TaxCategory.TransferOut,
+            "USD",
+            "45.99",
+            "Pizza for Bob's party",
+            ""
+        )
+        private val address = "yNo1YJcNBoveEHWB7eYmxFZBVEAYQo46Yb"
+        private val service = ServiceName.CrowdNode
     }
 
-    private val migrations = arrayOf(migration8To10)
+    private val migrations = arrayOf(AppDatabaseMigrations.migration11To12)
 
     @Rule
     @JvmField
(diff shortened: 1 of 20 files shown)


wallet/androidTest/de/schildbach/wallet/database/DatabaseMigrationTest.kt
```diff
@@ -54,7 +74,37 @@ open class DatabaseMigrationTest {
     @Throws(IOException::class)
     fun migrateAll() {
         // Create db and fill with data
-        testHelper.createDatabase(TEST_DB_NAME, 8).apply {
+        testHelper.createDatabase(TEST_DB_NAME, 1).apply {
```
HashEngineering, line 77: Similar to the other PR and perhaps unnecessary, this test shows that starting from version 1, that no data is migrated to version 12.

wallet/src/de/schildbach/wallet/AppDatabase.java
```diff
@@ -36,12 +41,14 @@
...
         BlockchainState.class,
+        TransactionMetadata.class,
+        AddressMetadata.class,
         BlockchainIdentityData.class,
         DashPayProfile.class,
         DashPayContactRequest.class,
         UserAlert.class,
         Invitation.class
-}, version = 16)
+}, version = 12)
```
HashEngineering, line 51: This looks strange.  DashPay was on 16, while master was on 11. 

In our next release of DashPay, there will be a complete data wipe on Platfrom, so none of the data will need to be saved, though it may cause the app to crash.

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -960,10 +964,10 @@ public void finalizeWipe() {
             @Override
             public void run() {
                 PlatformRepo.getInstance().clearDatabase(true);
-                ProcessPhoenix.triggerRebirth(WalletApplication.this);
```
HashEngineering, line 963: This should have been removed in a previous merge that resolved the Reset Wallet crash.  Though the location of this line was different than `master` at the time of the merge, so it would have been missed easily.

wallet/src/de/schildbach/wallet/AppDatabase.java
Syn-McJ: This will surely cause an upgraded app to crash at the start, so DashPay users will have to reinstall it. Is that fine?
Maybe we can set the version to 17, it shouldn't be a problem imo.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -874,7 +886,7 @@ public void onCreate() {
 
         try {
             bootStrapStream = getAssets().open(Constants.Files.MNLIST_BOOTSTRAP_FILENAME);
-            SimplifiedMasternodeListManager.setBootStrapStream(bootStrapStream);
+            SimplifiedMasternodeListManager.setBootStrapStream(bootStrapStream, null, SimplifiedMasternodeListManager.QUORUM_ROTATION_FORMAT_VERSION);
```
Syn-McJ, line 889: This seems to cause CI to fail.

Syn-McJ (commented):
Looks good. CI fails though because of `BlockchainService` changes, please check

wallet/src/de/schildbach/wallet/AppDatabase.java
HashEngineering: That is a good question.  When we release a new DashPay app, we are not expecting the old data to be usable.  We will tell users to uninstall, reinstall, but some will not.  If it crashes, the user may give up on the app instead of uninstalling and reinstalling.

Although Version 16, won't have the tables from version 11.

Dash Wallet 10->11 Add metadata tables - after the next master merge into DashPay.
DashPay 11->16, add dashpay tables, but leave out metadata tables
DashPay 16->17 - Add metadata tables.

For the purposes of DashPay on testnet, this is probably fine, but when considering an upgrade on mainnet, where we upgrade from DashWallet to DashPay, it seems like there might be room for losing metadata tables.  If we go with v17 now, I should put notes about changing it later?

Syn-McJ: When we upgrade the Wallet to DashPay, we will need to write migrations for adding DashPay tables. It will be a migration from whatever version the wallet has to whatever version DashPay has at that time.
We'll need these migrations anyway if we plan to do any changes in the Wallet database.
I'm sure the upgrade process will be tested thoufully.

HashEngineering: Ok, then I will return this version to 17.  When we do the upgrade from Dash Wallet to DashPay, then the need for the DashPay Testnet app will probably be reduced significantly and we won't need to support it as much.  If something is broken on the testnet side with older versions, then that is probably fine.

Syn-McJ (approved):
Looks good

PR #998, opened 2022-07-20
fix(coinbase): prevent transfer out if still syncing (NMA-1330)

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/TransferDashFragment.kt
```diff
@@ -172,6 +172,19 @@ class TransferDashFragment : Fragment(R.layout.transfer_dash_fragment) {
             )
         }
 
+        enterAmountToTransferViewModel.dashWalletNotSyncedCallback.observe(viewLifecycleOwner){
+            val dashAccountEmptyError = CoinbaseGenericErrorUIModel(
```
Syn-McJ, line 176: It might be better to use an `AdaptiveDialog` here. Some advantages:
- Auto dismissal on the lock screen.
- Clear flow: immediately obvious what happens when the user chooses positive/negative action.
- Easier to add additional code for the positive action (like analytics).
- Universal design that can be easily changed for all dialogs.

HashEngineering: Agreed.  Though new designs have to dialog when syncing.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/EnterAmountToTransferFragment.kt
```diff
@@ -228,4 +227,11 @@ class EnterAmountToTransferFragment : Fragment(R.layout.enter_amount_to_transfer
             }
         }
     }
+
+    fun showKeyboardAndButton(isVisible: Boolean) {
+        // avoid IllegalStateException
+        if (isAdded) {
+            viewModel.keyboardStateCallback.value = isVisible
+        }
+    }
```
HashEngineering, line 236: While the wallet is syncing, this does not work to hide the keyboard and button if the user leaves the Transfer Dash screen and returns to it.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/EnterAmountToTransferFragment.kt
```diff
@@ -228,4 +227,10 @@ class EnterAmountToTransferFragment : Fragment(R.layout.enter_amount_to_transfer
             }
         }
     }
+
+    fun showKeyboardAndButton(isVisible: Boolean) {
+        if (isAdded) { // avoid IllegalStateException
+            binding.keyboardContainer.isVisible = isVisible
```
HashEngineering, line 233: For some reason this doesn't work after leaving this screen and returning to it.

```
         viewModel.keyboardStateCallback.value = isVisible
         ```

Syn-McJ (approved):
Looks good

Syn-McJ (approved):

PR #1001, opened 2022-07-25
NMA-1335  Screen loading improvements

hadia:
Hide the loading dialog with balance loading 

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ:
@hadia this seems to only cover the dialog on Coinbase portal, but when we open "Buy Dash" or "Convert Crypto" there are also loading dialogs that might not be necessary. Can we get rid of those as well?

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseServicesViewModel.kt
```diff
@@ -69,8 +69,16 @@ class CoinbaseServicesViewModel @Inject constructor(
...
+    val latestUserBalance: LiveData<String>
+        get() = _latestUserBalance
+
     private fun getUserAccountInfo() = viewModelScope.launch(Dispatchers.Main) {
-        _showLoading.value = true
+        if(config.lastCoinbaseBalance.isNullOrEmpty()) {
+            _showLoading.value = true
+        }else{
+            _latestUserBalance.value = config.lastCoinbaseBalance
+        }
```
HashEngineering, line 81: So we will show the Loading dialog the first time the user views this Fragment?

HashEngineering (commented):
have a question for now.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseServicesViewModel.kt
hadia: in case for any reason the user did not get the balance on Service screen

HashEngineering (approved):

Syn-McJ (approved):
Looks good

HashEngineering (approved):

PR #1007, opened 2022-08-05
feat(coinbase): add transfer defaults

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/services/TransactionMetadataProvider.kt
+++ b/common/src/main/java/org/dash/wallet/common/services/TransactionMetadataProvider.kt
@@ -47,30 +47,55 @@ interface TransactionMetadataProvider {
     suspend fun getAllTransactionMetadata(): List<TransactionMetadata>
 
     // Address methods
+    /**
+     * mark an address with a tax category.  This will replace existing data
+     *
+     * @param address the address to mark
+     * @param isInput the address is an input in a transaction
+     * @param taxCategory the tax category
+     * @param service the name of the service associated with this address
+     */
     suspend fun markAddressWithTaxCategory(
         address: String,
-        sendTo: Boolean,
+        isInput: Boolean,
         taxCategory: TaxCategory,
         service: String
     )
 
+    /**
+     * mark an address with a tax category if it hasn't been marked
+     *
+     * @param address the address to mark
+     * @param isInput the address is an input in a transaction
+     * @param taxCategory the tax category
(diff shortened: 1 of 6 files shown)


wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -207,55 +207,57 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
-                        }
-                    }
-                }
-            } else {
-                // incoming transaction, checkout puts
-                for (output in outputs) {
+            // outgoing transaction, check inputs
+            for (input in inputs) {
+                if (input.connectedOutput != null) {
+                    val output = input.connectedOutput!!
```
HashEngineering, line 213: The main change here is removing the check to see if the balance is negative.

For coinable we mark the destination address as TransferOut and the transaction value is negative.  The original code was only checking inputs for addresses in the database.  However, for this case, we should be checking for the outputs in the database.

So this modified code will first check inputs, then outputs and it will return on the first match.

common/src/main/java/org/dash/wallet/common/services/TransactionMetadataProvider.kt
```diff
@@ -47,30 +47,55 @@ interface TransactionMetadataProvider {
...
 
     // Address methods
+    /**
+     * mark an address with a tax category.  This will replace existing data
+     *
+     * @param address the address to mark
+     * @param isInput the address is an input in a transaction
+     * @param taxCategory the tax category
+     * @param service the name of the service associated with this address
+     */
```
HashEngineering, line 57: added this documentation to explain the API.

After maybe two months, I had forgotten some of the basics and it was unclear how things were working or not working.

common/src/main/java/org/dash/wallet/common/services/TransactionMetadataProvider.kt
```diff
@@ -47,30 +47,55 @@ interface TransactionMetadataProvider {
...
     suspend fun markAddressWithTaxCategory(
         address: String,
-        sendTo: Boolean,
+        isInput: Boolean,
```
HashEngineering, line 60: this was using the old `sendTo` nomenclature, but really it should be `isInput` meaning: is the address used in an input to a transaction.

Syn-McJ (approved):
Looks good

PR #1009, opened 2022-08-07
Nma 1337

hadia:

- NMA-1337 User should be able to enter amount in their default fiat currency 
- Fix issue to Min amount  value  message warning to show default fiat currency
- NMA-1333 don't receive dash in the wallet when Convert Crypto( USDC -> Dash wallet)
- The ConvertCrypto and transfer to coinbase confirmation is showing the wrong direction message 
- The retry button that is on the failed transfer state is supposed to attempt to try to transfer the Dash to the DW

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/ConvertViewViewModel.kt
```diff
@@ -142,8 +142,13 @@ class ConvertViewViewModel @Inject constructor(
         }
         this._selectedCryptoCurrencyAccount.value = account
 
+        //To check if the user has different fiat than usd the min is 2 usd
+        val minFaitValue = "2".toBigDecimal() / account.currencyToUSDExchangeRate.toBigDecimal()
```
Syn-McJ, line 146: What if the min fiat value changes?

Syn-McJ (approved):
Looks good but I have some concerns about the hardcoded "min fiat" value. Doesn't seem like a good solution.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/ConvertViewViewModel.kt
hadia: this is an assumption from our side I mean value is 2 USD

HashEngineering: we can address this in the next coinbase related story.

HashEngineering (approved):

PR #1013, opened 2022-08-11
[NMA-1353] Coinbase loading dialog improvements

Syn-McJ:
Improvements for the loading dialog showing time on the Convert screen.
This PR includes #1009 to avoid conflicts.

Coinbase code seems to request the `/accounts` endpoint each time it needs something from it - I think there is an opportunity for caching and reusing the result. It will still be updated when a user opens buy/sell or the Coinbase portal and at some other times, but for deeper navigations, we might as well use the data we already have.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/repository/CoinBaseRepository.kt
```diff
@@ -34,9 +35,12 @@ class CoinBaseRepository @Inject constructor(
     private val commitBuyOrderMapper: CommitBuyOrderMapper,
     private val coinbaseAddressMapper: CoinbaseAddressMapper
 ) : CoinBaseRepositoryInt {
-    override suspend fun getUserAccount() = safeApiCall {
+    private var userAccountInfo: List<CoinBaseUserAccountData> = listOf()
```
Syn-McJ, line 38: All methods in this class call `/accounts` endpoint every time they need something which results in long processing times and loading dialogs. It's better to cache the result. `getUserAccount` will still update it periodically.
We might also want to add more sophisticated caching logic.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseConvertCryptoFragment.kt
```diff
@@ -183,47 +182,45 @@ class CoinbaseConvertCryptoFragment : Fragment(R.layout.fragment_coinbase_conver
...
             convertViewModel.setOnSwapDashFromToCryptoClicked(it)
         }
 
-
         convertViewModel.selectedLocalExchangeRate.observe(viewLifecycleOwner) {
             binding.convertView.exchangeRate = ExchangeRate(Coin.COIN, it.fiat)
             setConvertViewInput()
         }
 
-        viewModel.userAccountsWithBalance.observe(
```
Syn-McJ, line 199: Passing the execution flow around by calling/observing events is convoluted and unnecessary. Just call a suspend method from the fragment and wait for the result.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseConvertCryptoFragment.kt
```diff
@@ -261,6 +258,33 @@ class CoinbaseConvertCryptoFragment : Fragment(R.layout.fragment_coinbase_conver
...
+                when (uiState) {
+                    is BaseIdForFaitDataUIState.Success -> {
+                        uiState.baseIdForFaitDataList.let { list ->
+                            viewModel.setBaseIdForFaitModelCoinBase(list)
+                        }
+                    }
+
+                    is BaseIdForFaitDataUIState.LoadingState ->{
+                        if (uiState.isLoading) {
+                            showProgress(R.string.loading)
```
Syn-McJ, line 274: This is from #1009. This code adds more loading dialogs while we're trying to get rid of them. Is it really necessary here? @hadia

wallet/src/de/schildbach/wallet/ui/BuyAndSellViewModel.kt
```diff
@@ -193,7 +193,7 @@ class BuyAndSellViewModel
                     _coinbaseBalance.value = response.value?.balance?.amount
                 }
                 is ResponseResource.Failure -> {
-                    _coinbaseBalance.value =  if (config.lastCoinbaseBalance.isNullOrEmpty()) {
+                    _coinbaseBalance.value = if (!config.lastCoinbaseBalance.isNullOrEmpty()) {
```
Syn-McJ, line 196: This is an error. If `lastCoinbaseBalance` is not empty, we want to return it and not 0.

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseConvertCryptoFragment.kt
hadia: @Syn-McJ  we need to keep status if the user choose feature before all data related loaded

Syn-McJ: Yes, but this code sets `_baseIdForFaitModelCoinBase` which is only needed when the swap trade is initiated, not before. So technically, we don't need a loading dialog for this until the user actually starts the conversion process, right?

hadia (approved):
👍

integrations/coinbase-integration/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/convert_currency/model/BaseIdForFaitDataUIState.kt
```diff
@@ -0,0 +1,10 @@
+package org.dash.wallet.integration.coinbase_integration.ui.convert_currency.model
+
+import org.dash.wallet.common.ui.payment_method_picker.PaymentMethod
+import org.dash.wallet.integration.coinbase_integration.model.BaseIdForUSDData
+
+sealed class BaseIdForFaitDataUIState {
```
HashEngineering, line 6: perhaps, `Fait` should be `Fiat` -- we can fix this in a different PR if that is the case.

HashEngineering (approved):

HashEngineering:
This PR pulls in code from #1009 which has not been approved, it will be merged after #1009

PR #1014, opened 2022-08-11
[NMA-1356] Return to StakingActivity from VerifySeedActivity

Syn-McJ:
If the `VerifySeedActivity` opened from CrowdNode, we want to go back to CrowdNode and not the home screen when it's closed.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Excellent.

PR #1020, opened 2022-08-16
[NMA-1362] (Explore) Do not download DB from GCloud if the prepackaged one is newer

Syn-McJ:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreSyncWorker.kt
```diff
@@ -54,41 +54,52 @@ class ExploreSyncWorker @AssistedInject constructor(
...
                     AppExploreDatabase.getAppDatabase(
                         appContext,
                         config,
                         exploreRepository
                     )
-                }
-
-                localDataTimestamp = exploreRepository.localTimestamp
-                log.info("local data timestamp: $localDataTimestamp (${Date(localDataTimestamp)})")
+                } else {
```
Syn-McJ, line 79: If the prepackaged database is newer than the local one, it will be unpacked and the GCloud database won't be checked until the next launch. 
This is fine - we cannot rely on the localDataTimestamp check here until the next launch anyway since it is updated asynchronously after DB is unpacked and opened.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/ExploreRepository.kt
```diff
@@ -45,14 +44,14 @@ interface ExploreRepository {
     fun getTimestamp(file: File): Long
     fun getUpdateFile(): File
     suspend fun download()
-    fun deleteOldDB(dbFile: File)
+    fun markDbForDeletion(dbFile: File)
```
Syn-McJ, line 47: Does not actually immediately delete the file.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/ExploreRepository.kt
```diff
@@ -68,8 +67,6 @@ class GCExploreDatabase @Inject constructor(
         private val log = LoggerFactory.getLogger(GCExploreDatabase::class.java)
     }
 
-    private var contextRef: WeakReference<Context> = WeakReference(context)
```
Syn-McJ, line 71: This is getting forcefully unpacked anyway. Might as well use the `context` directly.

HashEngineering (approved):
Looks like this will work.

PR #1025, opened 2022-08-21
[NMA-1355] Remove Liquid integration

Syn-McJ:
- Liquid code is removed apart from the bits that check the login status.
- `integrations/coinbase-integration` is renamed to `integrations/coinbase`. The first variant seems redundant.
- Fixed the progress dialog crash on entering Buy & Sell. The problem here is that Uphold balance is returned before the activity is created, so the dialog isn't attached when dismissed. Since we don't actually need to show a progress dialog for the balance loading, I have this code removed.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/buy_sell/BuyAndSellIntegrationsActivity.kt
```diff
@@ -210,9 +187,7 @@ class BuyAndSellIntegrationsActivity : LockScreenActivity() {
         viewModel.upholdBalanceLiveData.observe(this) {
             if (it != null) {
                 when (it.status) {
-                    Status.LOADING -> {
-                        loadingDialog?.show(this)
```
Syn-McJ, line 214: We don't need to show the loading dialog when the balance is being fetched - just update UI when it's received.

wallet/src/de/schildbach/wallet/ui/buy_sell/BuyAndSellIntegrationsActivity.kt
```diff
@@ -223,7 +198,6 @@ class BuyAndSellIntegrationsActivity : LockScreenActivity() {
                                 balance
                             )
                         }
-                        loadingDialog?.dismiss()
```
Syn-McJ, line 226: This causes a crash since `Success` is reached before the activity is created. Removing this progress dialog solves the problem. Other cases of dismissing a progress dialog are called in a LiveData observer which should respect the lifecycle automatically.

wallet/src/de/schildbach/wallet/ui/coinbase/CoinBaseWebClientActivity.kt
```diff
@@ -26,23 +26,22 @@ import androidx.appcompat.widget.Toolbar
 import de.schildbach.wallet_test.BuildConfig
 import de.schildbach.wallet_test.R
 import org.dash.wallet.common.InteractionAwareActivity
-import org.dash.wallet.integration.liquid.databinding.ActivityLoginWebviewBinding
```
Syn-McJ, line 29: Feature/integration modules shouldn't depend on each other.

HashEngineering (approved):
Looks great.  This will reduce build times, perhaps offsetting the addition of coinbase.

PR #1027, opened 2022-08-24
[NMA-1060] Target SDK 31

Syn-McJ:
Changes necessary to target SDK 31.
Additionally, the gradle plugin is upgraded to version 7.2.2 and some related migrations were run:
Screenshot 2022-08-22 164932

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


fastlane/Fastfile
```diff
@@ -20,7 +20,7 @@ platform :android do
   lane :test do |options|
     flavor=options[:flavor].capitalize
 	type=options[:type].capitalize
-    gradle(task: "test")
+    gradle(task: "test#{flavor}#{type}UnitTest")
```
Syn-McJ, line 23: Seems like there was an error and the flavor/type weren't actually specified for the tests, which resulted in all tests for all flavors and types being run every time.

I think my original idea was to only run flavor/type that was requested for the build itself. This should reduce the amount of time necessary to make a manual distribution build.

HashEngineering (changes requested):
Looks good except that PR #1025, which was merged earlier today moved the `coinbase-integration` files to `coinbase` and that results in a conflict on several files.

Syn-McJ:
@HashEngineering fixed

HashEngineering (approved):

PR #1030, opened 2022-08-25
fix: setting desugaring library version to 1.1.5 to avoid crashes

Syn-McJ:
There are some crashes related to `java.time` library, we need a correct version of desugaring library to avoid them.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Just came across the bug this fixes, better approve and merge it.

PR #1032, opened 2022-08-27
[NMA-1373] (Crowdnode) Restore linked online account

Syn-McJ:
When a wallet is restored that had previously an online account linked to it, we want to restore the online account as well.
There is at least a confirmation transaction present for a linked account, so we can use that to detect it.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks good.

PR #1039, opened 2022-09-13
[NMA-1377] Dashpay 7.6.1 merge

Syn-McJ:
Coinbase, new font, and replacing old typography/button styles.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/invite/InviteHandler.kt
```diff
@@ -157,40 +134,52 @@ class InviteHandler(val activity: FragmentActivity, private val analytics: Analy
...
+        inviteAlreadyClaimedDialog.onFancyAlertButtonsClickListener = object :
+            FancyAlertDialog.FancyAlertButtonsClickListener {
+            override fun onPositiveButtonClick() {
+                handleDialogButtonClick(activity)
+            }
+
+            override fun onNegativeButtonClick() {
+                handleDialogButtonClick(activity)
+            }
+        }
```
Syn-McJ, line 171: Fancy viewModel is removed in master. Simple dialogs are replaced with the `AdaptiveDialog` here, the rest use the click listener for now - we can transition them from `FancyAlertDialog` later.

wallet/src/de/schildbach/wallet/ui/main/MainViewModel.kt
```diff
@@ -134,6 +142,10 @@ class MainViewModel @Inject constructor(
     val isNetworkUnavailable: LiveData<Boolean>
         get() = _isNetworkUnavailable
 
+    private val _stakingAPY = MutableLiveData<Double>()
+    val stakingAPY: LiveData<Double>
+        get() = _stakingAPY
```
Syn-McJ, line 147: Explore fragment is hosted in the `MainActivity` in DashPay, so staking API is moved here. It probably makes more sense too, since CrowdNode APY doesn't relate to Explore. This will be implemented in the master as well after the bottom nav bar is backported.

There is a risk for the `MainViewModel` to become a god class that handles everything; we can deal with it later after enough logic is moved here from fragments/activities and it becomes clear how to break it up. Right now it's still not very large.

HashEngineering:
This has a failing test, but can be fixed in another PR, perhaps even #1042

PR #1040, opened 2022-09-17
[NMA-1388] BottomNavigationView in Wallet

Syn-McJ:
We want to bring `BottomNavigationView` to the Wallet so that it's aligned with the DashPay navigation structure for ease of future merges.

However, some improvements can be made to DashPay solution - specifically, we can start using the Navigation Component instead of fragment transactions. This will require some work to merge back into DashPay, but not a huge amount.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -841,43 +821,4 @@ class ExploreViewModel @Inject constructor(
...
-        viewModelScope.launch(Dispatchers.IO) {
-            _stakingAPY.postValue(0.85 * blockchainStateProvider.getMasternodeAPY())
-        }
-    }
-
-    fun getLastStakingAPY() {
-        viewModelScope.launch(Dispatchers.IO) {
-            _stakingAPY.postValue(0.85 * blockchainStateProvider.getLastMasternodeAPY())
-        }
-    }
```
Syn-McJ, line 882: Moved to the `MainViewModel` since `ExploreFragment` is hosted in the `WalletActivity` now.

robo-script-7.1-and-above.json
```diff
@@ -2376,15 +2367,6 @@
...
 
-      },
-      {
-        "className": "de.schildbach.wallet.ui.widget.GlobalFooterView",
-        "recyclerViewChildPosition": -1,
-        "adapterViewChildPosition": -1,
-        "groupViewChildPosition": 0,
-        "resourceId": "",
-        "contentDescription": ""
-
```
Syn-McJ, line 2387: Is this file auto-generated?

wallet/res/values/values.xml
```diff
@@ -22,7 +22,7 @@
 		<item>@string/preferences_precision_labels_0_6</item>
 	</string-array>
 	<string-array name="preferences_block_explorer_values">
-		<item>http://insight.dash.org/insight/</item>
+		<item>https://insight.dash.org/insight/</item>
```
Syn-McJ, line 25: @knst noticed that we open `http` link instead of `https` here.

wallet/src/de/schildbach/wallet/ui/explore/ExploreActivity.kt
```diff
@@ -1,98 +0,0 @@
...
-import kotlinx.coroutines.launch
-import org.dash.wallet.common.ui.dialogs.AdaptiveDialog
-import org.dash.wallet.features.exploredash.ui.ExploreViewModel
-import org.dash.wallet.features.exploredash.ui.NavigationRequest
-import javax.inject.Inject
-
-@AndroidEntryPoint
-@FlowPreview
-@ExperimentalCoroutinesApi
-class ExploreActivity : BaseMenuActivity() {
```
Syn-McJ, line 42: Not needed anymore. Explore will be hosted in the `WalletActivity` and its nav graph included in the main nav graph.

The same can be done for other feature modules in the future; this is a good first step towards a single activity architecture.

robo-script-7.1-and-above.json
HashEngineering: This file was created by Android Studio as a Robo script after recording steps to run in Firebase TestLab.  It worked well for version v7.1 and a few releases after that.  

It is no longer usable and we can delete it.

common/src/main/java/org/dash/wallet/common/util/NavigationExtensions.kt
```diff
@@ -47,4 +48,13 @@ fun DialogFragment.dialogSafeNavigate(directions: NavDirections) {
             navController.navigate(directions)
         }
     }
+}
+
+enum class DeepLinkDestination(val deepLink: Uri) {
+    ReceiveDash(Uri.parse("android-app://hashengineering.darkcoin.wallet/payments/0")),
+    SendDash(Uri.parse("android-app://hashengineering.darkcoin.wallet/payments/1"))
+}
```
HashEngineering, line 56: Should `hashengineering.darkcoin.wallet` be replaced by a variable that is unique to the current app package id since we have a different id for testnet and dashpay and our devnet flavors?

robo-script-7.1-and-above.json
HashEngineering: This file can be deleted.

HashEngineering (changes requested):
I have two items noted above to look at.

common/src/main/java/org/dash/wallet/common/util/NavigationExtensions.kt
Syn-McJ: Not necessary, because the deep link is declared here and it stays the same across flavors: https://github.com/dashevo/dash-wallet/pull/1040/files#diff-9843f8fc1e09aaca34667187e3b60c8deb5858c04f4fd213d77e32b398ddcd92R45

So is confirmed by QA. We can, however, move it to a constant to avoid potential problems with refactoring later.

HashEngineering (approved):
LGTM

PR #1041, opened 2022-09-19
[NMA-1392] UI crashes

Syn-McJ:
A bunch of UI-related crashes + some refactoring.
These crashes aren't related to the 8.0 upgrade and appeared earlier.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/res/values/styles.xml
```diff
@@ -245,8 +245,14 @@
         <item name="strokeColor">@color/gray_40</item>
     </style>
 
-    <style name="ExampleBackground" parent="GrayBackground">
-        <item name="cornerRadius">20dp</item>
+    <style name="TaxCategoryExampleBackground">
+        <item name="backgroundColor">@color/tx_category_example_background</item>
+        <item name="strokeColor">@color/tx_category_example_background</item>
```
Syn-McJ, line 250: I made some UI fixes to make the tax explainer UI closer to the design, plus added support for the dark theme.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/Constants.kt
```diff
@@ -16,23 +16,24 @@
...
-const val USD_CURRENCY = "USD"
-const val ERROR_ID_INVALID_REQUEST = "invalid_request"
-const val ERROR_ID_2FA_REQUIRED = "two_factor_required"
-const val ERROR_MSG_INVALID_REQUEST = "That code was invalid"
-const val MIN_SWAP_DASH_AMOUNT_VALUE = "0.01909299"
-const val VALUE_ZERO = "0"
-const val DEFAULT_CURRENCY_USD = "USD"
-const val MIN_USD_COINBASE_AMOUNT = "2"
\ No newline at end of file
+object CoinbaseConstants {
```
Syn-McJ, line 19: It's better to group constants to an object. Keeps it more structured in case we want to reference these in the wallet module, plus no need to import each one separately.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseBuyDashFragment.kt
```diff
@@ -144,26 +145,25 @@ class CoinbaseBuyDashFragment : Fragment(R.layout.fragment_coinbase_buy_dash) {
...
 
     private fun setupPaymentMethodPayment() {
-        viewModel.activePaymentMethods.observe(viewLifecycleOwner){
-            binding.paymentMethodPicker.paymentMethods = it
-        }
-
-        arguments?.let {
-            viewModel.getActivePaymentMethods(CoinbaseBuyDashFragmentArgs.fromBundle(it).paymentMethods)
+        viewModel.activePaymentMethods.observe(viewLifecycleOwner) {
+            if (it.isNotEmpty()) {
```
Syn-McJ, line 155: We should not get to the point of `activePaymentMethods` being empty due to changes in `CoinbaseServicesFragment`, so this is just in case.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/ui/CoinbaseServicesFragment.kt
```diff
@@ -75,40 +69,30 @@ class CoinbaseServicesFragment : Fragment(R.layout.fragment_coinbase_services) {
...
-                                    dismissProgress()
-                            }
+            sharedViewModel.paymentMethodsUiState.observe(viewLifecycleOwner) { uiState ->
+                // New value received
+                when (uiState) {
+                    is PaymentMethodsUiState.Success -> {
+                        val paymentMethodsArray = uiState.paymentMethodsList.toTypedArray()
+
+                        if (paymentMethodsArray.isEmpty()) {
+                            showNoPaymentMethodsError()
```
Syn-McJ, line 79: If `paymentMethodsList` is empty, this is functionally the same as an error case later.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseActivityViewModel.kt
```diff
@@ -40,11 +40,13 @@ class CoinbaseActivityViewModel @Inject constructor(
...
-    private val _paymentMethodsUiState = MutableStateFlow<PaymentMethodsUiState>(PaymentMethodsUiState.LoadingState(true))
-    val paymentMethodsUiState: StateFlow<PaymentMethodsUiState> = _paymentMethodsUiState
+    private val _paymentMethodsUiState = MutableLiveData<PaymentMethodsUiState>(PaymentMethodsUiState.LoadingState(true))
+    val paymentMethodsUiState: LiveData<PaymentMethodsUiState>
+        get() = _paymentMethodsUiState
 
-    private val _baseIdForFaitModelCoinBase= MutableStateFlow<BaseIdForFaitDataUIState>(BaseIdForFaitDataUIState.LoadingState(true))
-    val baseIdForFaitModelCoinBase:StateFlow<BaseIdForFaitDataUIState> = _baseIdForFaitModelCoinBase
+    private val _baseIdForFaitModelCoinBase = MutableLiveData<BaseIdForFaitDataUIState>(BaseIdForFaitDataUIState.LoadingState(true))
+    val baseIdForFaitModelCoinBase: LiveData<BaseIdForFaitDataUIState>
```
Syn-McJ, line 48: We should not mix approaches (`LiveData` vs `StateFlow`) for communicating between ViewModels and fragments in a single module. It's fine to use `StateFlow` everywhere in a feature module if one prefers it, but it should be consistent.

Personally, I found that flows have some problems with respecting view lifecycles, so I stick to LiveData for now.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/EnterAmountToTransferViewModel.kt
```diff
@@ -195,16 +195,18 @@ class EnterAmountToTransferViewModel @Inject constructor(
...
+            } ?: CoinbaseConstants.VALUE_ZERO
         }
 
     private fun applyCoinbaseExchangeRate(amount: String): String {
         return coinbaseExchangeRate?.let { uiModel ->
-            val cleanedValue =
-                amount.toBigDecimal() / uiModel.currencyToDashExchangeRate.toBigDecimal()
+            val cleanedValue = amount
+                .replace(',', '.') // TODO: the amount sometimes comes here with a comma as decimal separator.
+                // TODO: it's better to identify the root of this and replace in there to prevent this problem from appearing anywhere else.
```
Syn-McJ, line 206: It's not clear to me where this value is coming from, but it sometimes comes with a comma as a decimal separator.
Eventually, we should fix this at the point of receiving such value.

common/src/main/java/org/dash/wallet/common/ui/NetworkUnavailableFragment.kt
```diff
@@ -1,60 +0,0 @@
...
-import android.os.Bundle
-import android.view.View
-import androidx.core.view.isVisible
-import androidx.fragment.app.Fragment
-import androidx.lifecycle.ViewModelProvider
-import org.dash.wallet.common.R
-import org.dash.wallet.common.databinding.NetworkUnavailableFragmentBinding
-
-class NetworkUnavailableFragment(val messageId: Int = 0, val buttonTextId: Int = 0) :
-    Fragment(R.layout.network_unavailable_fragment) {
```
HashEngineering, line 29: This must have only been used in the liquid integration.

PR #1043, opened 2022-09-21
fix: improve sync speeds for wallet restoration/rescan

HashEngineering:
A listener in WalletTransactionObserver seemed to be the problem after doing analysis and many, many tests

https://github.com/dashevo/dashj/pull/178
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
+++ b/common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
@@ -51,6 +51,8 @@ interface WalletDataProvider {
 
     fun observeTransactions(vararg filters: TransactionFilter): Flow<Transaction>
 
+    fun observeTransactionsWithConfidence(vararg filters: TransactionFilter): Flow<Transaction>
+
     fun getTransactions(vararg filters: TransactionFilter): Collection<Transaction>
 
     fun wrapAllTransactions(vararg wrappers: TransactionWrapper): Collection<TransactionWrapper>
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/dash/wallet/common/WalletDataProvider.kt
```diff
@@ -51,6 +51,8 @@ interface WalletDataProvider {
 
     fun observeTransactions(vararg filters: TransactionFilter): Flow<Transaction>
 
+    fun observeTransactionsWithConfidence(vararg filters: TransactionFilter): Flow<Transaction>
+
```
HashEngineering, line 55: I didn't like having to add another method, but it was more simple than adding  Boolean to `observeTransactions`

Perhaps this would have worked:
`fun observeTransactions(observeTxConf: Boolean = false, vararg filters: TransactionFilter): Flow<Transaction>`

Doing this required changing every method call, mostly in the crowdnode module:
`fun observeTransactions(observeTxConf: Boolean, vararg filters: TransactionFilter): Flow<Transaction>`

I wanted to minimize the number of changes associated with turning ON the transaction confidence listener.

Syn-McJ: This will work fine. 
We can probably also change `vararg` to a simple list later - there isn't really any need to be fancy here - and then add a default boolean argument.

Syn-McJ (approved):
Looks great

PR #1044, opened 2022-09-22
[NMA-1378] About screen redesign + Explore sync info fixes

Syn-McJ:
About screen needs a redesign. Explore fixes include a bugfix for the `finalizeUpdate` method, plus some changed logic for displaying the last sync time.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreDatabase.kt
```diff
@@ -0,0 +1,190 @@
...
+
+        @VisibleForTesting
+        suspend fun preloadAndOpen(
+            dbBuilder: Builder<ExploreDatabase>,
+            repository: ExploreRepository,
+            dbUpdateFile: File
+        ): ExploreDatabase {
+            require(dbUpdateFile.exists()) { "dbUpdateFile doesn't exist"}
+
+            return suspendCancellableCoroutine { coroutine ->
```
Syn-McJ, line 127: The worker needs to wait for the preload operation to ensure that `localDatabaseTimestamp` is updated; to achieve it this method is turned into a suspend method.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DataSyncStatusService.kt
```diff
@@ -19,11 +19,8 @@ package org.dash.wallet.features.exploredash.repository
 
 import kotlinx.coroutines.flow.Flow
 import org.dash.wallet.common.data.Resource
-import java.lang.Exception
 
 interface DataSyncStatusService {
-    suspend fun setSyncProgress(progress: Double)
-    suspend fun setSyncError(exception: Exception)
```
Syn-McJ, line 26: Interface users don't actually need access to these methods since they aren't supposed to change the state. The implementation can be injected into the worker which needs full access.

common/src/main/java/org/dash/wallet/common/services/SystemActionsService.kt
```diff
@@ -0,0 +1,27 @@
...
+package org.dash.wallet.common.services
+
+interface SystemActionsService {
+    fun copyText(text: String, label: String, showToast: Boolean = true)
+    fun shareText(text: String)
+    fun reviewApp()
+    fun openAppSettings()
+    fun openNotificationChannelSettings(channel: String)
+    fun openNotificationSettings()
+}
```
Syn-McJ, line 27: Such an interface could be useful in the case we ever use multiplatform: both android and ios can call on it from shared code, while implementations will be specific to the platform.

wallet/res/values/strings-extra.xml
```diff
@@ -284,16 +284,21 @@
     <string name="tools_show_xpub">Extended Public Key</string>
 
     <string name="about_help_us_improve">Help us improve your experience</string>
-    <string name="about_review_and_rate">Review &amp; Rate the app</string>
+    <string name="about_review_and_rate">Review and rate the app</string>
```
HashEngineering, line 287: this change will make it easier for translators.  The `&` causes some trouble and we have to manually fix some of the text that includes this symbol.

HashEngineering (approved):
Looks great and has some good refactoring.

Syn-McJ:
@HashEngineering thanks, fixed

HashEngineering (approved):

PR #1045, opened 2022-09-23
[NMA-1383] "You will receive" remains after reopen Convert Crypto

Syn-McJ:
We need to clear entered dash/fiat amount from the shared viewModel after navigating back from the Convert Crypto fragment.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
looks great

PR #1047, opened 2022-09-29
[NMA-1380] Integrations Overview screen

Syn-McJ:
For Coinbase and Uphold, we want to add/update the info screens that provide some overview on these features.

- Added UI for Coinbase overview and updated UI for the Uphold
- Moved `BuySellIntegrationsActivity` to `BuySellIntegrationsFragment` for simpler navigation structure.
- Because Coinbase authentication is moved from the `IntegrationOverviewFragment`, some refactoring is done in the `BuySellIntegrationsFragment` and the ViewModel.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/BaseAdapter.kt
```diff
@@ -1,36 +0,0 @@
...
-import androidx.recyclerview.widget.RecyclerView
-
-abstract class BaseAdapter<T : Any,>(
-    diffCallback: DiffUtil.ItemCallback<T> = object : DiffUtil.ItemCallback<T>() {
-        override fun areItemsTheSame(oldItem: T, newItem: T) =
-            oldItem == newItem
-
-        @SuppressLint("DiffUtilEquals")
-        override fun areContentsTheSame(oldItem: T, newItem: T) =
-            oldItem == newItem
```
Syn-McJ, line 18: Generally, `areItemsTheSame` and `areContentsTheSame` should not have the same comparing logic - it makes the whole DiffUtil and the adapter useless since any change to the item contents causes the whole list to redraw.

`areItemsTheSame` should compare items identifiers (unique type, transaction hashes, etc) to check if it's the same item (Coinbase vs Uphold), while `areContentsTheSame` should compare fields of the items to know when to redraw a specific row.

wallet/src/de/schildbach/wallet/adapter/BuyAndSellDashServicesAdapter.kt
```diff
@@ -17,82 +17,128 @@
...
+    ): BuyAndSellDashServicesViewHolder {
+        val inflater = LayoutInflater.from(parent.context)
+        val binding = ItemServiceListBinding.inflate(inflater, parent, false)
 
+        return BuyAndSellDashServicesViewHolder(binding)
     }
 
-    fun updateIconState(hasInternet: Boolean) {
-        isOnline = hasInternet
-        notifyDataSetChanged()
```
Syn-McJ, line 47: The items already have a field `serviceStatus` that is assigned, among other things, based on network availability. We can use it for graying out icons as well - it will allow getting rid of `notifyDataSetChanged` which is bad for performance.

integrations/uphold/src/main/java/org/dash/wallet/integration/uphold/ui/UpholdSplashActivity.java
```diff
@@ -45,6 +45,7 @@
 import kotlin.Unit;
 
 @AndroidEntryPoint
+// TODO: move this into IntegrationOverviewFragment
```
Syn-McJ, line 48: I wanted to use the same fragment for both Coinbase and Uphold overview since they use the same layout, however, since UpholdSplashActivity is called by a callback to finish authentication it's hard to do it this way right now - needs some more refactoring. Leaving it be for now.

HashEngineering (approved):
This has new text to translate.

PR #1042, opened 2022-09-19
fix(dashpay): fix key issue on create identity, restore identity

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

PR #1048, opened 2022-10-03
feat: support dashpay sync txmetadata

HashEngineering:
Before this PR there was a single table called `transaction_metadata` that stored tx metadata.

With this PR, we are adding two more tables
1. transaction_metadata_cache - this table contains changes that need to be synced with platform.
2. transaction_metadata_platform - this table is synced with Platform

in theory `transaction_metadata = latest(transaction_metadata_platform + transaction_metadata_cache)`.  
In other words, the current state of the metadata that is displayed to the user is equal to the data stored on platform plus the data not yet pushed to platform.  

**Example:**
1. For transaction A, there is no memo
2. User sets a memo "Alice Pizza Party" for transaction A.  
3. `transaction_metadata` is updated with this change.  
4. An entry is also added to `transaction_metadata_cache` which indicates that the memo on transaction A has changed.
5. After a period of time (currently 15 seconds) or before Reset Wallet, all records (including transaction A) in `transaction_metadata_cache` are combined into a single list and published to Platform. `transaction_metadata_cache` is emptied.
6. Shortly after 5), the app detects that there is a new information published on Platform and this will be added to the `transaction_metadata_platform` table.

**Restoration of a Wallet**
1. Sync headers
2. Sync masternode lists
3. Sync Platform data (identity, username, contacts, profiles and tx_metadata). Tx Metadata is stored in the `transaction_metadata_platform` table.  No transactions have been found.
4. Sync blockchain and transactions.
5. As each transaction is found, we search in `transaction_metadata_platform` for the latest data on a transaction. Then an entry with that information is added to `transaction_metadata`.  For transaction A, it will find "Alice Pizza Party" as the memo.
6. After blockchain sync is complete, `transaction_metadata` should be the same as it was before the Reset Wallet action. 

**Refactoring**
* `PlatformRepo` has been split into `PlatformRepo` and `PlatformSyncService` and `PlatformBroadcastService`
* `PlatformRepo` can still be further refactored to separate out identity/username creation, etc in other stories.

#1042 
https://github.com/dashevo/android-dpp/pull/35
https://github.com/dashevo/android-dashpay/pull/67

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/data/TaxCategory.kt
```diff
@@ -43,6 +43,10 @@ enum class TaxCategory(val value: Int) {
             return values()[value];
         }
 
+        fun fromValue(name: String): TaxCategory? {
+            return values().find { it.name.lowercase() == name.lowercase() }
+        }
```
HashEngineering, line 48: This allows TaxCategories to be stored as strings on Platform.  

We may want to change to numbers, which could save some fees.

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -995,15 +1008,8 @@ public void finalizeWipe() {
...
-        Executors.newSingleThreadExecutor().execute(new Runnable() {
-            @Override
-            public void run() {
-                PlatformRepo.getInstance().clearDatabase(true);
-            }
-        });
-
         // clear data on wallet reset
-        transactionMetadataProvider.clear();
+        WalletApplicationExt.INSTANCE.clearDatabases(this, true);
```
HashEngineering, line 1012: Here we are removing several different run and forget methods and replacing it with a single run and forget method.

wallet/src/de/schildbach/wallet/WalletApplicationExt.kt
```diff
@@ -0,0 +1,48 @@
...
+        val scope = CoroutineScope(Executors.newSingleThreadExecutor().asCoroutineDispatcher())
+
+        scope.launch(Dispatchers.IO) {
+            platformSyncService.clearDatabases()
+            if (isWalletWipe) {
+                transactionMetadataProvider.clear()
+            }
+            platformRepo.clearDatabase(isWalletWipe)
+        }
+    }
```
HashEngineering, line 47: I wanted to use `runBlocking` instead but the database clear functions cannot run on the UI thread.  Is there a different way to do this so it is not a run and forget call?

wallet/src/de/schildbach/wallet/data/TransactionMetadataDocument.kt
```diff
@@ -0,0 +1,20 @@
...
+    var txId: Sha256Hash,
+    var sentTimestamp: Long? = null,
+    var taxCategory: TaxCategory? = null,
+    var currencyCode: String? = null,
+    var rate: Double? = null,
+    var memo: String? = null,
+    var service: String? = null
+) {
+
+}
```
HashEngineering, line 20: On platform, `TransactionMetadataDocument` contain an encrypted list of transaction metadata.  Here the metadata will be stored with one transaction change per Platform document `id` per record.

wallet/src/de/schildbach/wallet/data/TransactionMetadataDocumentDao.kt
```diff
@@ -0,0 +1,106 @@
...
+        SELECT sentTimestamp 
+        FROM transaction_metadata_platform 
+        WHERE timestamp = (
+            SELECT MAX(timestamp) 
+            FROM transaction_metadata_platform 
+            WHERE txId = :txId
+                AND sentTimestamp is NOT NULL
+        )
+        """)
+    suspend fun getSentTimestamp(txId: Sha256Hash): Long?
```
HashEngineering, line 46: This query scheme is present on all of these get functions.

We want to obtain the most recent value of some data for a particular transaction.  This appears to have one query embedded in another.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -239,13 +244,17 @@ public void onThrottledWalletChanged() {
         public void onCoinsReceived(final Wallet wallet, final Transaction tx, final Coin prevBalance,
                 final Coin newBalance) {
 
-            final int bestChainHeight = blockChain.getBestChainHeight();
-            final boolean replaying = bestChainHeight < config.getBestChainHeightEver();
+            //final int bestChainHeight = blockChain.getBestChainHeight();
+            final boolean replaying = blockchainState.getReplaying();
```
HashEngineering, line 248: This fix prevents old transactions from being assigned current exchange rates.  

For some reason `bestChainHeight < config.getBestChainHeightEver();` was false when restoring a wallet.  Instead, we will check the most recently saved `blockchainState` to see if it is replaying.

I wonder though, if the app crashes, will replaying be true or false when the app resumes the blockchain sync?  There may still be some holes in this fix.  Need to investigate further why the original code wasn't working correctly.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -239,13 +244,17 @@ public void onThrottledWalletChanged() {
...
 
             long now = new Date().getTime();
             long blockChainHeadTime = blockChain.getChainHead().getHeader().getTime().getTime();
             boolean insideTxExchangeRateTimeThreshold = (now - blockChainHeadTime) < TX_EXCHANGE_RATE_TIME_THRESHOLD_MS;
 
+            log.info("onCoinsReceived: {}; rate: {}; replaying: {}; inside: {}, confid: {}; will update {}",
+                    tx.getTxId(), tx.getExchangeRate(), replaying, insideTxExchangeRateTimeThreshold, tx.getConfidence().getConfidenceType(),
+                    tx.getExchangeRate() == null && ((!replaying || insideTxExchangeRateTimeThreshold) || tx.getConfidence().getConfidenceType() == ConfidenceType.PENDING));
+
```
HashEngineering, line 257: For now we will leave in these extra log statements until the sync code appears stable.

wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.java
```diff
@@ -1132,13 +1144,7 @@ public void onDestroy() {
...
             }
             //Clear the blockchain identity
-            executor.execute(new Runnable() {
-                @Override
-                public void run() {
-                    // This code is not executed during a wipe, only a blockchain reset
-                    PlatformRepo.getInstance().clearDatabase(false);
-                }
-            });
+            WalletApplicationExt.INSTANCE.clearDatabases(application, false);
```
HashEngineering, line 1147: we can use the same cleanup code as we did in `WalletApplication`

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -53,82 +59,150 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
+        timestamp: Long,
+        isSyncingPlatform: Boolean
+    ) {
+        updateAndInsertIfNotExist(txId, isSyncingPlatform) {
+            transactionMetadataDao.updateSentTime(txId, timestamp)
+            if (!isSyncingPlatform) {
+                transactionMetadataChangeCacheDao.insertSentTime(txId, timestamp)
+            }
+        }
+    }
```
HashEngineering, line 166: `isSyncingPlatform` is true when this method is called from `PlatformSyncService` -- this will prevent an infinite loop of sending new data to platform which is then read, possibly triggering those same changes to be sent again.

wallet/src/de/schildbach/wallet/WalletApplicationExt.kt
Syn-McJ: I think you can do `runBlocking(Dispatchers.IO)` if that's what you mean. 
But my idea was to move the run-and-forget call in here which is what you did - I don't think we need to change it to a blocking call since yes, it can block UI accidentally.

Syn-McJ (approved):
Looks great. Nice to see the `PlatformRepo` being refactored.

build.gradle
```diff
@@ -1,14 +1,14 @@
...
-        coroutines_version = '1.5.2'
+        coroutines_version = '1.6.4'
         ok_http_version = '4.9.1'
         dashjVersion = '0.18.2-SNAPSHOT'
         hiltVersion = '2.38.1'
         hiltWorkVersion = '1.0.0'
         workRuntimeVersion='2.6.0'
         firebaseVersion = '28.4.2'
-        roomVersion = '2.3.0'
+        roomVersion = '2.4.0-alpha03'
```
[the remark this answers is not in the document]
HashEngineering: I tried to use `2.4.3` and it required some other changes which broke the build process -- that can be attempted in the next PR.  At least the `dashpay` branch will now build on a Mac M1 system.

PR #1053, opened 2022-10-22
[NMA-1419] Tx statuses bugfix

Syn-McJ:
To make sure that transaction statuses on the home page are refreshed correctly and the sync time remains fast, we need to make use of the `walletChanged` event with some throttling.

- `WalletTransactionObserver` is combined with `WalletChangedObserver` into a single class (we can move other observers in there in the future).
- Observers are run in `USER_THREAD` instead of `SAME_THREAD`.
- `ThrottlingWalletChangeListener` is not used anymore since throttling is up to observers using `debounce`.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/transactions/WalletObserver.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.bitcoinj.wallet.listeners.WalletCoinsSentEventListener
+import org.dash.wallet.common.transactions.filters.TransactionFilter
+
+class WalletObserver(private val wallet: Wallet) {
+    fun observeWalletChanged(): Flow<Unit> = callbackFlow {
+        val walletChangeListener = WalletChangeEventListener {
+            trySend(Unit)
+        }
+
+        wallet.addChangeEventListener(Threading.USER_THREAD, walletChangeListener)
```
Syn-McJ, line 40: `USER_THREAD` doesn't make much of a difference in the case of a simple `trySend` since it's quite fast. However, it does make a difference below in the `observeTransactions` method since the transaction filter is running on that thread.

Additionally, I figure if there are any errors caused by `trySend` or the filters, we don't want to crash the thread where the sync is running.

wallet/src/de/schildbach/wallet/transactions/WalletMostRecentTransactionsObserver.kt
```diff
@@ -28,7 +27,6 @@ import org.bitcoinj.core.Transaction
 import org.bitcoinj.utils.Threading
 import org.bitcoinj.wallet.Wallet
 
-@ExperimentalCoroutinesApi
 class WalletMostRecentTransactionsObserver(private val wallet: Wallet) {
     fun observe(): Flow<Transaction> = callbackFlow {
```
HashEngineering, line 31: Perhaps at a later time, we could merge the functions of this class into `WalletObserver`.

HashEngineering (approved):
This looks great.

HashEngineering (approved):
LGTM

PR #1061, opened 2022-11-21
[NMA-1413] (Coinbase) Show different error dialog in case of unusable payment method

Syn-McJ:
If there is a payment method but buying from it isn't allowed, we want to show a different error dialog. Currently, "No payment methods" error is shown.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks great.  This will require a translation update after being merged to master.

PR #1065, opened 2022-11-24
feat(coinbase): cache base id values [NMA-1402]

Syn-McJ:
Base ids of currencies are required for making swaps with them. The request to get base ids returns 100kb of data and often times out, so we want to cache them.

- Added caching for Coinbase network layer in general.
- Added custom cache for the `assets/prices` endpoint.
- Added datastore config for Coinbase and began to move some prefs in there.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/Configuration.java
+++ b/common/src/main/java/org/dash/wallet/common/Configuration.java
@@ -91,7 +91,6 @@ public class Configuration {
 
     public static final String PREFS_KEY_LAST_COINBASE_ACCESS_TOKEN = "last_coinbase_access_token";
     public static final String PREFS_KEY_LAST_COINBASE_REFRESH_TOKEN = "last_coinbase_refresh_token";
-    public static final String PREFS_KEY_LAST_COINBASE_BALANCE = "last_coinbase_balance";
     public static final String PREFS_KEY_COINBASE_USER_ACCOUNT_ID = "coinbase_account_id";
     public static final String PREFS_KEY_COINBASE_AUTH_INFO_SHOWN = "coinbase_auth_info_shown";
     public static final String PREFS_KEY_COINBASE_USER_WITHDRAWAL_LIMIT = "withdrawal_limit";
@@ -535,6 +534,7 @@ public class Configuration {
     }
 
     // Coinbase
+    // TODO: put new preferences in the CoinbaseConfig and migrate these.
 
     public void setLastCoinBaseAccessToken(String token) {
         prefs.edit().putString(PREFS_KEY_LAST_COINBASE_ACCESS_TOKEN, token).apply();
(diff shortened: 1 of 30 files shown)


integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/network/RemoteDataSource.kt
```diff
@@ -16,55 +16,65 @@
...
         return OkHttpClient.Builder()
             .addInterceptor(HeadersInterceptor(userPreferences))
+            .addInterceptor(CustomCacheInterceptor(context, config))
             .connectTimeout(20.seconds.toJavaDuration())
             .callTimeout(20.seconds.toJavaDuration())
             .readTimeout(20.seconds.toJavaDuration())
+            .cache(Cache(
+                directory = CoinbaseConstants.getCacheDir(context),
+                maxSize = 10L * 1024L * 1024L // 10 MB
+            ))
```
Syn-McJ, line 77: The built-in cache functionality won't help us with the base_ids (for that we'll use `CustomCacheInterceptor`), but Coinbase APIs provide caching values like `wtag` which okHttp can take advantage of. This should improve networking for the Coinbase feature in general.

common/src/main/java/org/dash/wallet/common/Configuration.java
```diff
@@ -553,14 +553,6 @@ public String getLastCoinbaseRefreshToken() {
...
     }
 
-    public void setLastCoinbaseBalance(String balance) {
-        prefs.edit().putString(PREFS_KEY_LAST_COINBASE_BALANCE, balance).apply();
-    }
-
-    public String getLastCoinbaseBalance() {
-        return prefs.getString(PREFS_KEY_LAST_COINBASE_BALANCE, null);
-    }
-
```
Syn-McJ, line 563: I figured we should start moving feature-specific prefs from here into dedicated configs. The last balance seems like a good one to start with since it's not critical if something goes wrong with it.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/utils/CoinbaseConfig.kt
```diff
@@ -0,0 +1,80 @@
...
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+import org.bitcoinj.core.Coin
+import java.io.IOException
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class CoinbaseConfig @Inject constructor(private val context: Context) {
```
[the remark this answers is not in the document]
Syn-McJ: Yeah, I figure we can have a base class or provide an instance of the same class with different storage names. We'll see.

PR #1068, opened 2022-12-06
feat(dashdirect): add login function with temporary UI

hadia:

Add sign in dialog for dashdirect Api
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -500,7 +517,22 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
+            }
+        }
+
+        viewModel.dashDirectSignIn.observe(viewLifecycleOwner) {
+            if (it) {
+                // TODO open Buy card UI
+                Toast.makeText(requireContext(), "Open Buy Card", Toast.LENGTH_SHORT).show()
+                viewModel.logEvent(AnalyticsConstants.Explore.MERCHANT_DETAILS_BUY_GIFT_CARD)
+            }
+        }
```
Syn-McJ, line 535: The pattern of calling a function and observing result in `LiveData` is not great. It makes the execution flow more complicated and hard to track.

Instead, simply call a suspend function in a `lifecycleScope` and process the result.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -201,6 +206,10 @@ class ExploreViewModel @Inject constructor(
     val screenState: LiveData<ScreenState>
         get() = _screenState
 
+    private val _dashDirectSignIn = MutableLiveData<Boolean>()
+    val dashDirectSignIn: LiveData<Boolean>
+        get() = _dashDirectSignIn
```
Syn-McJ, line 211: You don't need this. See below.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -390,6 +400,52 @@ class ExploreViewModel @Inject constructor(
...
 
+    fun isUserSignInDashDirect() = repository.isUserSignIn()
+
+    fun signInToDashDirect(email: String, password: String) {
+        if (repository.isUserSignIn()) {
+            _dashDirectSignIn.value = true
+            return
+        }
+
+        viewModelScope.launch {
```
Syn-McJ, line 411: Just make `signInToDashDirect` suspend and return the result directly. See below.

Syn-McJ (changes requested):
Some changes needed.

Syn-McJ (approved):
This is better. Some more points to improve in the future:
- `runBlocking` usage is discourage. It's better to have the method that deals with settings as `suspend`. Reading from prefs is a IO operation and should be treated in an async manner.
- Returning `ResponseResource` from the viewModel method is unnecessary. Just return the result or throw an error, then `try/catch` in the caller.

HashEngineering:
The base branch is called `feature/DashDirect` which doesn't follow our `feature-*` nameing scheme which is required to activate CI.

HashEngineering:
new branch created named `feature-dashdirect`

HashEngineering (approved):

PR #1071, opened 2022-12-13
feat: Send screen improvements [NMA-1039]

Syn-McJ:
We want to resolve the issue of `Insufficient funds` error appearing after the user clicked the Max button and then erased the last digit or two.

- Max button will display the amount equal to `wallet balance - fee`
- Per @HashEngineering suggestion, if there is an error after erasing the last digit, we will recalculate the amount to show the max non-error amount that can be sent.
- Events added for the show/hide balance button.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/payments/MaxOutputAmountCoinSelector.kt
```diff
@@ -0,0 +1,45 @@
...
+        private const val TX_OUTPUT_SIZE = 34 // estimated size for a typical transaction output
+        private const val TX_INPUT_SIZE = 148 // estimated size for a typical compact pubkey transaction input
+    }
+
+    override fun select(target: Coin, candidates: MutableList<TransactionOutput>): CoinSelection {
+        val value = Coin.valueOf(candidates.sumOf { it.value.value })
+        val inputCount = candidates.size.toLong()
+        val outputCount = 1L
+        // Formula is lifted from DashSync for ios: https://github.com/dashpay/dashsync-iOS/blob/master/DashSync/shared/Models/Wallet/DSAccount.m#L1680-L1711
+        // Android has an extra byte per input
```
Syn-McJ, line 38: Not sure where the extra byte comes from, but it seems consistent - the fee on Android is larger by 1 for each input.
Otherwise, the formula works great.

HashEngineering (commented):
Excellent

wallet/src/de/schildbach/wallet/payments/MaxOutputAmountCoinSelector.kt
HashEngineering: Yes, I noticed that the fees on Android seemed to be 1 duff more, but didn't think that it was due to the size of the input.

It is probably due to the signature being 1 byte larger.

PR #1075, opened 2023-01-01
fix: SweepWallet back button and OffsetDialog crush [NMA-1444]

Syn-McJ:
Fixes for some regression bugs.

- Handle the back button manually for the `SweepWalletActivity`. Moved it to Kotlin while at it.
- Check if the `OffsetDialogFragment` is added before proceeding with offset logic.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/dialogs/OffsetDialogFragment.kt
```diff
@@ -46,6 +46,10 @@ open class OffsetDialogFragment : BottomSheetDialogFragment() {
         super.onViewCreated(view, savedInstanceState)
 
         dialog?.setOnShowListener { dialog ->
+            if (!this@OffsetDialogFragment.isAdded) {
+                return@setOnShowListener
+            }
```
Syn-McJ, line 51: My guess is that this listener is sometimes called after the dialog is dismissed. Some simulations with a delay support this theory.

HashEngineering (approved):
LGTM

PR #1079, opened 2023-01-10
 Purchase gift card UI

hadia:
Create Ui For Purchase gift card
 
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountFragment.kt
```diff
@@ -77,6 +86,13 @@ class EnterAmountFragment: Fragment(R.layout.fragment_enter_amount) {
         val args = requireArguments()
         binding.maxButtonWrapper.isVisible = args.getBoolean(ARG_MAX_BUTTON_VISIBLE)
         binding.amountView.showCurrencySelector = args.getBoolean(ARG_SHOW_CURRENCY_SELECTOR_BUTTON)
+        binding.amountView.hideResultContainer = args.getBoolean(ARG_HIDE_AMOUNT_RESULT_CONTAINER)
+        binding.currencyOptions.isVisible = args.getBoolean(ARG_CURRENCY_Code)
```
Syn-McJ, line 90: This seems to be an error. It should be `ARG_CURRENCY_OPTIONS_PICKER_VISIBLE` here. This causes the currency option picker to be hidden in other Enter Amount screens.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -206,6 +208,7 @@ class ExploreViewModel @Inject constructor(
     val screenState: LiveData<ScreenState>
         get() = _screenState
 
+    var purchaseGiftCardData: Pair<Pair<Coin, Fiat>, Merchant>? = null
```
Syn-McJ, line 211: This seems a bit convoluted. Let's create a dedicated data class to store this.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/SearchFragment.kt
```diff
@@ -84,7 +83,7 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
     @Inject
     lateinit var analyticsService: AnalyticsService
     private val binding by viewBinding(FragmentSearchBinding::bind)
-    private val viewModel: ExploreViewModel by activityViewModels()
+    private val viewModel: ExploreViewModel by navGraphViewModels(R.id.explore_dash) { defaultViewModelProviderFactory }
```
Syn-McJ, line 86: This also has to be changed everywhere where this view model is used, specifically in `ExploreMapFragment` and `FiltersDialog`. Right now, the location and filter functionality in Explore is broken.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dialogs/PurchaseGiftCardConfirmDialog.kt
```diff
@@ -0,0 +1,126 @@
...
+                merchant?.merchantId?.let {
+                    paymentValue?.let { amountValue ->
+                        exploreViewModel.purchaseGiftCard(
+                            merchantId = it,
+                            giftCardAmount = amountValue.first.toPlainString().toDouble(),
+                            currency = Constants.DASH_CURRENCY,
+                            deviceID = UUID.randomUUID().toString()
+                        )
+                    }
+                }
```
Syn-McJ, line 109: This code doesn't need to be in the fragment. You're getting two values from the viewModel and calling a method in the same viewModel.
Just move this whole thing to the viewModel.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
```diff
@@ -401,12 +402,13 @@ class ExploreViewModel @Inject constructor(
...
 
-    suspend fun purchaseGiftCard(deviceID: String, currency: String, giftCardAmount: Double, merchantId: Int) =
+    suspend fun purchaseGiftCard(deviceID: String, currency: String, giftCardAmount: Double, merchantId: Long) =
         repository.purchaseGiftCard(
             deviceID = deviceID,
             giftCardAmount = giftCardAmount,
             currency = currency,
-            merchantId = merchantId
+            merchantId = merchantId,
+            userEmail = ""
```
Syn-McJ, line 411: Can't this be moved into `PurchaseGiftCardViewModel` since we have it?

features/exploredash/src/main/res/layout/fragment_purchase_gift_card.xml
```diff
@@ -0,0 +1,80 @@
...
+        app:layout_constraintHorizontal_bias="0.0"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/title_bar" />
+
+    <androidx.constraintlayout.widget.Guideline
+        android:id="@+id/enter_amount_guideline"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constraintGuide_percent="0.18" />
```
Syn-McJ, line 29: I think 18% isn't enough. On tall screens it doesn't look good - it's too close to the header. The amount should be somewhere in the center between the header and the keyboard.

Syn-McJ (changes requested):
Some fixes are needed.

features/exploredash/src/main/res/layout/fragment_purchase_gift_card.xml
hadia: Fixed

common/src/main/java/org/dash/wallet/common/ui/enter_amount/EnterAmountFragment.kt
hadia: Fixed

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ExploreViewModel.kt
hadia: Refactored code in Putsches viewmodel

HashEngineering: Good Point, we have some things like this in DashJ too.

Syn-McJ (approved):
Looks good now

HashEngineering (approved):
LGTM

PR #1082, opened 2023-01-22
Feature/login via otp

hadia:
Add  signup and  login to dash direct 
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/res/layout/dialog_dash_direct_login_info.xml
```diff
@@ -0,0 +1,97 @@
...
+
+    <Button
+        android:id="@+id/dialog_extra_message"
+        style="@style/Button.Tertiary.Small.Blue"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="10dp"
+        android:layout_marginBottom="25dp"
+        app:layout_constraintTop_toTopOf="@id/title"
+        tools:text="Back up recovery phase" />
```
Syn-McJ, line 76: I don't see an icon on this button like in the design. Since you're using a custom layout, you can modify it as necessary and not use the "extra message" pattern.

features/exploredash/src/main/res/layout/dialog_dash_direct_login_info.xml
```diff
@@ -0,0 +1,97 @@
...
+        android:layout_gravity="end"
+        android:layout_marginTop="20dp"
+        android:paddingHorizontal="0dp"
+        android:src="@drawable/ic_close"
+        android:text="@string/explore_apply_filter" />
+
+    <ImageView
+        android:id="@+id/dialog_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="50dp"
```
Syn-McJ, line 38: Per the design, this should be 60dp

features/exploredash/src/main/res/layout/dialog_dash_direct_login_info.xml
```diff
@@ -0,0 +1,97 @@
...
+
+    <Button
+        android:id="@+id/dialog_negative_button"
+        style="@style/Button.Tertiary.Blue"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginHorizontal="20dp"
+        android:layout_marginTop="5dp"
+        android:layout_marginBottom="15dp"
+        tools:text="Cancel" />
```
Syn-McJ, line 95: Per the design, these should be `Button.Primary.Large.Blue` and `Button.Tertiary.Large.Blue`. You can see the size here:
Screenshot 2023-01-23 at 10 21 53

features/exploredash/src/main/res/layout/dialog_dash_direct_login_info.xml
```diff
@@ -0,0 +1,97 @@
...
+        app:layout_constraintTop_toTopOf="@id/title"
+        tools:text="@string/create_an_account_at_dash_direct_or_log_in_to_the_existing_one" />
+
+    <Button
+        android:id="@+id/dialog_extra_message"
+        style="@style/Button.Tertiary.Small.Blue"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="10dp"
+        android:layout_marginBottom="25dp"
```
Syn-McJ, line 74: Per the design, these should be bigger. 45 is the total margin between this label and the button, so you can have 40dp here and leave 5dp on the button.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DashDirectRepository.kt
```diff
@@ -34,22 +34,67 @@ class DashDirectRepository @Inject constructor(
...
+    override fun getDashDirectEmail(): String? {
+        return runBlocking { config.getPreference(DashDirectConfig.PREFS_KEY_DASH_DIRECT_EMAIL) }
+    }
+
+    override fun logout() {
+        runBlocking {
+            config.setPreference(DashDirectConfig.PREFS_KEY_LAST_DASH_DIRECT_ACCESS_TOKEN, "")
+            config.setPreference(DashDirectConfig.PREFS_KEY_DASH_DIRECT_EMAIL, "")
+        }
+    }
```
Syn-McJ, line 97: As I've mentioned before, the use of `runBlocking` is discouraged. Reading the preferences is an asynchronious I/O operation and should be treated as such. On slow devices, these calls can cause ANRs.

Please remove `runBlocking`, make the methods `suspend`, and call them from a coroutine context.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/DashDirectUserAuthFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        binding.input.doOnTextChanged { text, _, _, _ ->
+
+            binding.inputWrapper.isErrorEnabled = false
+
+            if(currentDirectUserAuthType!=DashDirectUserAuthType.OTP )
+            binding.continueButton.isEnabled =isEmail(text)
+            else
+                binding.continueButton.isEnabled = !text.isNullOrEmpty()
```
Syn-McJ, line 114: This code snippet is a mess. There are missing and extra spaces and indentations do not match, all of which is making it hard to read. 

Please reformat it manually or use some formatting tool. There are many other places in the code that have similar issues.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/DashDirectUserAuthFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+            ViewCompat.setOnApplyWindowInsetsListener(decor) { _, insets ->
+               val showingKeyboard = insets.isVisible(WindowInsetsCompat.Type.ime())
+
+                if (showingKeyboard) {
+                    val systemBarsIMEInsets = insets.getInsets(WindowInsetsCompat.Type.ime())
+                    binding.root.setPadding(0, 0, 0, systemBarsIMEInsets.bottom)
+
+                }
+                insets
+            }
```
Syn-McJ, line 126: This doesn't work well right now - the button stays raised if I hide the keyboard.

There is a better way to handle this - set `android:windowSoftInputMode="adjustResize"` for the `WalletActivity`. Unless there are side effects of changing that field, we should use it instead. Did you try it?

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/DashDirectUserAuthFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+@ExperimentalCoroutinesApi
+@AndroidEntryPoint
+class DashDirectUserAuthFragment : Fragment(R.layout.fragment_dash_direct_user_auth) {
+    companion object {
+        private val log = LoggerFactory.getLogger(DashDirectUserAuthFragment::class.java)
+    }
+
+    private var currentDirectUserAuthType: DashDirectUserAuthType?=null
+    private val binding by viewBinding(FragmentDashDirectUserAuthBinding::bind)
+    private val exploreViewModel: ExploreViewModel by navGraphViewModels(R.id.explore_dash) { defaultViewModelProviderFactory }
```
Syn-McJ, line 61: I thought we have a dash direct viewModel? Let's not put dash direct logic into the `ExploreViewModel` if we can avoid it.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/utils/DashDirectConfig.kt
```diff
@@ -32,6 +32,7 @@ import javax.inject.Singleton
 class DashDirectConfig @Inject constructor(private val context: Context) {
     companion object {
         val PREFS_KEY_LAST_DASH_DIRECT_ACCESS_TOKEN = stringPreferencesKey("last_dash_direct_access_token")
+        val PREFS_KEY_DASH_DIRECT_EMAIL = stringPreferencesKey("dash_direct_email")
```
Syn-McJ, line 35: @hadia this will be stored as plain text in the settings. Personal user info needs to be stored securely. For example:
https://developer.android.com/topic/security/data
https://proandroiddev.com/securing-androids-datastore-ad56958ca6ee

Syn-McJ (changes requested):
Some problems need to be fixed.

features/exploredash/src/main/res/layout/dialog_dash_direct_login_info.xml
hadia: nice catch Fixed

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/utils/DashDirectConfig.kt
hadia: nice catch ,I forget about that part

hadia: create EncryptionProvider

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/DashDirectUserAuthFragment.kt
hadia: I need that part to check and map user email in the search fragment

Syn-McJ (approved):
Looks good now. @HashEngineering could you please check the encryption too to confirm that everything is ok

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/data/model/merchent/GetDataMerchantIdResponse.kt
```diff
@@ -0,0 +1,507 @@
...
+            val savingsPercentage: Int? = 0,
+            @SerializedName("SystemName")
+            val systemName: String? = "",
+            @SerializedName("TermsAndConditions")
+            val termsAndConditions: String? = "",
+            @SerializedName("Website")
+            val website: String? = ""
+        )
+    }
+}
```
Syn-McJ, line 507: Do we need all of these or just a couple of classes/fields? Having the entirety of the response in here might be confusing for someone dealing with this code later.

hadia: I add this api for @HashEngineering  to use in his ticket

Syn-McJ: I understand that, but there are a lot of fields here and we might only need several. In his PR @HashEngineering has a much smaller amount of fields: https://github.com/dashpay/dash-wallet/blob/16560031887d19f1d85020875601aff6261efcae/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/data/model/merchants/GetMerchantByIdResponse.kt

HashEngineering: In this PR, we can leave all of the fields as they are.  In my PR, I will remove what is not needed and make a few other necessary changes to the data types on the required fields.

HashEngineering (changes requested):
The security aspects look good.

`EncryptionProvider` is a kotlin translated copy of `ModernEncryptionProvider` -- I think this is a good start because no changes were made to the current wallet encryption files or class.

Later we should write a test for this class and compare to `openssl enc -aes-256-gcm`.  Also at that time, some of the common encryption classes between `common` and `wallet` could be refactored into a single set of classes.

`security` is misspelled in one of the packages.

HashEngineering (approved):
Looks good to me.

PR #1083, opened 2023-01-24
feat: add dashdirect gift card info to Purchase UI

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/build.gradle
```diff
@@ -15,6 +15,11 @@ android {
         targetSdkVersion 31
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles "consumer-rules.pro"
+        javaCompileOptions {
+            annotationProcessorOptions {
+                arguments += ["room.schemaLocation": "$projectDir/schemas".toString()]
+            }
+        }
```
HashEngineering, line 22: `room.schemaLocation` - it seems that this is a global setting.  If this doesn't match `wallet` then these schemas will be placed in a new folder compared to where they were before.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreDatabase.kt
```diff
@@ -45,7 +46,7 @@ import kotlin.coroutines.resumeWithException
     MerchantFTS::class,
     Atm::class,
     AtmFTS::class
-], version = 1)
+], version = 2, exportSchema = true)
```
HashEngineering, line 49: exportSchema is used mainly to properly configure the DB generated by the Explore Sync Tool.  We need the `identity_hash`

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreDatabase.kt
```diff
@@ -83,7 +93,8 @@ abstract class ExploreDatabase : RoomDatabase() {
 
             log.info("Open database {}", config.exploreDatabaseName)
             return dbBuilder
-                .fallbackToDestructiveMigration()
+                //.fallbackToDestructiveMigration()
```
HashEngineering, line 96: I will need to remove the comment.  By removing `.fallbackToDestructiveMigration()` we will get more useful error reporting during development instead of an empty database.

Going forward, using `fallbackToDestructiveMigration()` will be unnecessary.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/network/service/DashDirectServicesApi.kt
```diff
@@ -30,4 +32,9 @@ interface DashDirectServicesApi {
         @Header(DashDirectConstants.EMAIL) email: String,
         @Body purchaseGiftCardRequest: PurchaseGiftCardRequest
     ): PurchaseGiftCardResponse?
+    @POST("GetMerchantById")
+    suspend fun getMerchantById(
+        //@Header(DashDirectConstants.API_KEY) appKey: String,
+        @Body getMerchantByIdRequest: GetMerchantByIdRequest
+    ): GetMerchantByIdResponse?
```
HashEngineering, line 39: This is a guess what the parameters for this API should be.  The Authentication token appears to be included in the headers.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DashDirectRepository.kt
```diff
@@ -93,6 +95,9 @@ class DashDirectRepository @Inject constructor(
         config.clearAll()
     }
 
+    suspend fun getUserAppKey() =
+        config.getPreference(DashDirectConfig.PREFS_KEY_LAST_DASH_DIRECT_ACCESS_TOKEN)!!
+
```
HashEngineering, line 100: not sure if this method is needed.  In this PR it is not used.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ItemDetails.kt
```diff
@@ -208,7 +208,7 @@ class ItemDetails(context: Context, attrs: AttributeSet): LinearLayout(context,
                 payBtn.text = context.getText(R.string.explore_pay_with_dash)
                 payBtn.setOnClickListener { onSendDashClicked?.invoke(true) }
             } else {
-                payBtn.isVisible = !merchant.deeplink.isNullOrBlank()
+                payBtn.isVisible = true//!merchant.deeplink.isNullOrBlank()
```
HashEngineering, line 211: in the new explore db, the `deeplink` field is empty, so we shouldn't use it to hide/show the Pay/Purchase button.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreDatabase.kt
```diff
@@ -62,6 +63,15 @@ abstract class ExploreDatabase : RoomDatabase() {
...
 
+        @JvmStatic
+        val migration1To2 = object : Migration(1, 2) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.execSQL("ALTER TABLE merchant ADD COLUMN minCardPurchase REAL DEFAULT 0.0")
+                database.execSQL("ALTER TABLE merchant ADD COLUMN maxCardPurchase REAL DEFAULT 0.0")
+                database.execSQL("ALTER TABLE merchant ADD COLUMN savingsPercentage REAL DEFAULT 0.0")
+            }
+        }
+
```
HashEngineering, line 74: this will handle the database upgrade, though in theory it may not be used.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/data/model/merchent/GetDataMerchantIdResponse.kt
```diff
@@ -489,13 +489,13 @@ data class GetDataMerchantIdResponse(
...
-            val maximumCardPurchase: Int? = 0,
+            val maximumCardPurchase: Double? = 0.0,
             @SerializedName("MinimumCardPurchase")
-            val minimumCardPurchase: Int? = 0,
+            val minimumCardPurchase: Double? = 0.0,
             @SerializedName("PaymentInstructions")
             val paymentInstructions: String? = "",
             @SerializedName("SavingsPercentage")
-            val savingsPercentage: Int? = 0,
+            val savingsPercentage: Double? = 0.0,
```
HashEngineering, line 498: This file is not currently being used -- this is the class that contains all fields returned, though we are only using 3 of them.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/ExploreRepository.kt
```diff
@@ -77,7 +77,7 @@ class GCExploreDatabase @Inject constructor(
             return if (isTestDB) {
                 "explore/explore-testnet.db"
             } else {
-                "explore/explore.db"
+                "explore/explore-devnet.db" // this is temporary, should have means of selecting dev db
```
HashEngineering, line 80: for now we will use the `devnet` db during development.

However, when we move to production, perhaps we should call it `explore-2.db`

Then the Cloud Storage will have:
1. `explore.db` - older Dash Wallet apps will see this v1 file and sync with it.  
2. `explore-2.db` - newer apps will see this v2 file and sync with it.

This way old apps don't download v2 files which will be considered invalid.  Not sure if a v2 db will end up clearing out all merchants with a destructive fallback or if the app will try to sync with the updated db many, many times.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/PurchaseGiftCardFragment.kt
```diff
@@ -55,7 +58,7 @@ class PurchaseGiftCardFragment : Fragment(R.layout.fragment_purchase_gift_card)
     private val exploreViewModel: ExploreViewModel by navGraphViewModels(R.id.explore_dash) { defaultViewModelProviderFactory }
     private val enterAmountViewModel by activityViewModels<EnterAmountViewModel>()
 
-    var selectedMerchent: Merchant? = null
+    var selectedMerchant: Merchant? = null
```
HashEngineering, line 61: There is a `merchant` data item in the Fragment, but there is also one in the ViewModel that is assigned when the user presses "Next"

Shouldn't the merchant data be in ViewModel only?

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ItemDetails.kt
HashEngineering: There may be another field that should be checked instead

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/PurchaseGiftCardFragment.kt
```diff
@@ -84,61 +87,124 @@ class PurchaseGiftCardFragment : Fragment(R.layout.fragment_purchase_gift_card)
...
+                                merchant.savingsPercentage = it.savingsPercentage
+                                merchant.minCardPurchase = it.minimumCardPurchase
+                                merchant.maxCardPurchase = it.maximumCardPurchase
+
+                                setCardPurchaseLimits()
+                                setDiscountHint()
+                            }
+                        }
+                    }
+                }
```
HashEngineering, line 117: Here we check to see if there is updated merchant info (only 3 fields)

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/PurchaseGiftCardViewModel.kt
```diff
@@ -65,6 +65,11 @@ class PurchaseGiftCardViewModel @Inject constructor(
     var purchaseGiftCardDataMerchant: Merchant? = null
     var purchaseGiftCardDataPaymentValue: Pair<Coin, Fiat>? = null
 
+    var minCardPurchaseCoin: Coin = Coin.ZERO
+    var minCardPurchaseFiat: Fiat = Fiat.valueOf(Constants.USD_CURRENCY, 0)
+    var maxCardPurchaseCoin: Coin = Coin.ZERO
+    var maxCardPurchaseFiat: Fiat = Fiat.valueOf(Constants.USD_CURRENCY, 0)
```
HashEngineering, line 71: We need to keep track of the purchase limits in Fiat and DASH.  The user sees the Fiat values, but internally the `EnterAmountViewModel` needs the limits in DASH (`Coin`).

There is probably a better way to handle this.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/PurchaseGiftCardViewModel.kt
```diff
@@ -94,16 +99,48 @@ class PurchaseGiftCardViewModel @Inject constructor(
...
-                    userEmail = email
-                )
-            }
+    suspend fun getMerchantById(merchantId: Long): ResponseResource<GetMerchantByIdResponse?>? {
+        repository.getDashDirectEmail()?.let { email ->
+            return repository.getMerchantById(
+                merchantId = merchantId,
+                includeLocations = false,
+                userEmail = email
+            )
```
HashEngineering, line 108: currently `purchaseGiftCardDataMerchant` is only valid when the user presses `Next`, but we need to query this information at the beginning.

However, if we didn't have a merchant field in the fragment and the view model, but only one in the viewModel -- I could make changes to simplify code related to the selected merchant.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.java
```diff
@@ -186,12 +194,20 @@ public static String formatFiatWithoutComma(String fiatValue){
 
     public static String fiatToString(Fiat fiat) {
         MonetaryFormat format = Constants.INSTANCE.getSEND_PAYMENT_LOCAL_FORMAT().noCode();
+        return fiatToString(fiat, format);
+    }
+
+    public static String fiatToStringRoundUp(Fiat fiat) {
+        MonetaryFormat format = Constants.INSTANCE.getSEND_PAYMENT_LOCAL_FORMAT().noCode().roundingMode(RoundingMode.UP);
+        return fiatToString(fiat, format);
+    }
```
HashEngineering, line 203: Currently, this is used for discounted values only and the purpose is to round up to ensure that the calculations agree.  In both cases below, the discount was 2.75% when the user entered $50.  The enter amount screen was showing the discounted value of 48.62 while the payment confirmation dialog shows 48.63.

Displayed |  | internal (long) | Description
-- | -- | -- | --
48.62 |   | 4862499917 | calculated from DASH value and converted to fiat using an exchange rate
48.63 |   | 4862500000 | calculated from an exact fiat value of 50.00

If we round up, then both will match, which means that the user will see 48.63 both times in the flow.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/PurchaseGiftCardViewModel.kt
Syn-McJ: I seem to recall discussing with @hadia that `EnterAmountFragment` and `EnterAmountViewModel` are not suitable here because of this exact reason. We should've handled this differently - by building this fragment from the ground out of necessary components so that it is similar to `EnterAmountFragment` but with its own logic.

We had somewhat complicated our lives here, but as long as it works it's fine I guess.

wallet/src/de/schildbach/wallet/Constants.java
```diff
@@ -74,7 +74,7 @@ public final class Constants {
                 IS_PROD_BUILD = true;
                 FILENAME_NETWORK_SUFFIX = "";
                 WALLET_NAME_CURRENCY_CODE = "dash";
-                org.dash.wallet.common.util.Constants.INSTANCE.setEXPLORE_GC_FILE_PATH("explore/explore.db");
+                org.dash.wallet.common.util.Constants.INSTANCE.setEXPLORE_GC_FILE_PATH("explore/explore-devnet.db");
```
Syn-McJ, line 77: The same concern here as before. Can we not use testnet builds to test this database and not modify production values?

Syn-McJ (approved):
@HashEngineering looks good but I have some questions regarding the explore database namings. Could you please clarify before merging.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/ExploreRepository.kt
HashEngineering: Since this is for preloaded databases, then we don't need to rename it.

wallet/src/de/schildbach/wallet/Constants.java
HashEngineering: This is where it gets tricky.  DashDirect doesn't support testnet, only mainnet.  Purchases couldn't happen, but other aspects of the new feature could be tested.

Additionally, the production and testnet databases are version 1.  

`explore-devnet.db` is version 2.  I was thinking about renaming this to `explore-v2.db` which also needs to be the name created by **Explore Sync Tool**.  Version 2 not only has extra fields, but it also has different data since it comes from the devapi server where `merchantId`'s differ and the deeplinks are currently missing.  So we can't use testnet builds, but need to make devapi builds (it is named devnet, but it has nothing to do with any devnets).

Syn-McJ (approved):

hadia (approved):
Nice work @HashEngineering

PR #1085, opened 2023-02-04
fix(crowdnode): prevent Android 12 crash and close online linking screen

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/AndroidActionsService.kt
+++ b/wallet/src/de/schildbach/wallet/service/AndroidActionsService.kt
@@ -17,6 +17,7 @@
 
 package de.schildbach.wallet.service
 
+import android.app.Activity
 import android.content.*
 import android.net.Uri
 import android.os.Build
@@ -49,6 +50,7 @@ class AndroidActionsService @Inject constructor(
         }
 
         val shareIntent = Intent.createChooser(sendIntent, null)
+        shareIntent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
         context.startActivity(shareIntent)
     }
 
--- a/wallet/src/de/schildbach/wallet/ui/staking/StakingActivity.kt
+++ b/wallet/src/de/schildbach/wallet/ui/staking/StakingActivity.kt
@@ -97,14 +97,14 @@ class StakingActivity : LockScreenActivity() {
         when (status) {
             OnlineAccountStatus.None -> { }
             OnlineAccountStatus.Linking, OnlineAccountStatus.SigningUp -> super.turnOffAutoLogout()
-            OnlineAccountStatus.Validating -> super.turnOffAutoLogout()
-            else -> {
+            OnlineAccountStatus.Validating -> {
                 val isWebView = navController.currentDestination?.id == R.id.crowdNodeWebViewFragment
                 if (isWebView) {
                     navController.navigate(WebViewFragmentDirections.webViewToPortal())
                 }
                 super.turnOnAutoLogout()
             }
+            else -> super.turnOnAutoLogout()
         }
     }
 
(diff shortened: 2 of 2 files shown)


wallet/src/de/schildbach/wallet/service/AndroidActionsService.kt
```diff
@@ -49,6 +50,7 @@ class AndroidActionsService @Inject constructor(
         }
 
         val shareIntent = Intent.createChooser(sendIntent, null)
+        shareIntent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
```
HashEngineering, line 53: This fix was tested on Android 12.

wallet/src/de/schildbach/wallet/ui/staking/StakingActivity.kt
```diff
@@ -96,7 +97,14 @@ class StakingActivity : LockScreenActivity() {
...
             OnlineAccountStatus.Linking, OnlineAccountStatus.SigningUp -> super.turnOffAutoLogout()
-            else -> super.turnOnAutoLogout()
+            OnlineAccountStatus.Validating -> super.turnOffAutoLogout()
+            else -> {
+                val isWebView = navController.currentDestination?.id == R.id.crowdNodeWebViewFragment
+                if (isWebView) {
+                    navController.navigate(WebViewFragmentDirections.webViewToPortal())
+                }
+                super.turnOnAutoLogout()
+            }
```
HashEngineering, line 107: Not sure if this works.  Code was adapted from this change:

https://github.com/dashpay/dash-wallet/commit/a7f5073e5a4fb4a5c62f00037c5dbcd09ead284d#diff-1a8247bf710f1f2c58895df41fea7d3b06c50ca531744bda31260ee654671398L96-L101

Syn-McJ: This is not going to work. We need to move from the `WebView` to the portal when the status is changed from `linking` to `validating`, so the logic should be in the `OnlineAccountStatus.Validating` case.

This will also avoid issues with another place where this WebView is used which is creating the new account. That flow doesn't have `.linking` or `.validating` cases.

Syn-McJ, line 107:
```suggestion
            OnlineAccountStatus.Validating -> {
            val isWebView = navController.currentDestination?.id == R.id.crowdNodeWebViewFragment
                if (isWebView) {
                    navController.navigate(WebViewFragmentDirections.webViewToPortal())
                }
                super.turnOnAutoLogout()
            }
            else -> super.turnOnAutoLogout()
```

Syn-McJ (changes requested):
@HashEngineering the autodismiss issue isn't fixed, needs a small change.

wallet/src/de/schildbach/wallet/ui/staking/StakingActivity.kt
```diff
@@ -96,6 +97,13 @@ class StakingActivity : LockScreenActivity() {
         when (status) {
             OnlineAccountStatus.None -> { }
             OnlineAccountStatus.Linking, OnlineAccountStatus.SigningUp -> super.turnOffAutoLogout()
+            OnlineAccountStatus.Validating -> {
+                val isWebView = navController.currentDestination?.id == R.id.crowdNodeWebViewFragment
+                if (isWebView) {
+                    navController.navigate(WebViewFragmentDirections.webViewToPortal())
+                }
+                super.turnOnAutoLogout()
+            }
```
HashEngineering, line 106: @Syn-McJ This is your suggestion and it works for me.

Syn-McJ: Yes, this should work

Syn-McJ (approved):

HashEngineering:
Thank you for the approval.  QA has also approved.

PR #1086, opened 2023-02-13
feat(dashpay): use coinjoin to create username

HashEngineering:

Create a username by first mixing the funds used for it.

https://github.com/dashpay/dashj/pull/196

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/CoinJoinService.kt
+++ b/wallet/src/de/schildbach/wallet/service/CoinJoinService.kt
@@ -1,18 +1,15 @@
 package de.schildbach.wallet.service
 
-import android.content.Intent
 import de.schildbach.wallet.Constants
 import de.schildbach.wallet.WalletApplication
-import de.schildbach.wallet.data.BlockchainStateDao
-import de.schildbach.wallet.ui.dashpay.CreateIdentityService
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.asCoroutineDispatcher
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import org.bitcoinj.coinjoin.CoinJoinClientManager
 import org.bitcoinj.coinjoin.CoinJoinClientOptions
 import org.bitcoinj.coinjoin.PoolMessage
@@ -26,28 +23,32 @@ import org.bitcoinj.coinjoin.utils.CoinJoinManager
 import org.bitcoinj.coinjoin.utils.ProTxToOutpoint
 import org.bitcoinj.core.AbstractBlockChain
 import org.bitcoinj.core.Coin
+import org.bitcoinj.core.Context
 import org.bitcoinj.utils.Threading
 import org.bitcoinj.wallet.Wallet
 import org.bitcoinj.wallet.WalletEx
 import org.dash.wallet.common.WalletDataProvider
+import org.dash.wallet.common.data.NetworkStatus
+import org.dash.wallet.common.services.BlockchainStateProvider
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import java.util.concurrent.Executors
-import java.util.concurrent.TimeUnit
 import javax.inject.Inject
-import kotlin.coroutines.resumeWithException
-import kotlin.coroutines.suspendCoroutine
 
 interface CoinJoinService {
     fun needsToMix(amount: Coin): Boolean
-    fun configureMixing(amount: Coin, requestKeyParameter: RequestKeyParameter, requestDecryptedKey: RequestDecryptedKey)
-    suspend fun prepareAndStartMixing(): Boolean
-    fun prepareAndStartMixingAsync()
+    fun configureMixing(
+        amount: Coin,
+        requestKeyParameter: RequestKeyParameter,
+        requestDecryptedKey: RequestDecryptedKey
+    )
+
+    suspend fun prepareAndStartMixing()
     fun startMixing(): Boolean
     fun stopMixing()
-    fun setBlockchain(blockChain: AbstractBlockChain)
     fun getMixingStatus(): MixingStatus
-    // TODO: add getMixingProgress() : Flow<progress data>
+    suspend fun waitForMixing()
+    suspend fun waitForMixingWithException()
 }
 
 enum class MixingStatus {
@@ -59,9 +60,9 @@ enum class MixingStatus {
 }
 
 class CoinJoinMixingService @Inject constructor(
-     val walletDataProvider: WalletDataProvider,
-     val walletApplication: WalletApplication,
-     val blockchainStateDao: BlockchainStateDao
+    val walletDataProvider: WalletDataProvider,
+    val walletApplication: WalletApplication,
+    val blockchainStateProvider: BlockchainStateProvider
 ) : CoinJoinService {
 
     companion object {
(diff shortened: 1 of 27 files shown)


wallet/src/de/schildbach/wallet/service/CoinJoinService.kt
```diff
@@ -0,0 +1,323 @@
...
+                stopMixing()
+            }
+            mixingStatus == MixingStatus.PAUSED && this.networkStatus == NetworkStatus.NOT_AVAILABLE && networkStatus == NetworkStatus.CONNECTED -> {
+                startMixing()
+            }
+            mixingStatus == MixingStatus.ERROR -> setMixingComplete()
+        }
+        this.networkStatus = networkStatus
+        this.mixingStatus = mixingStatus
+    }
```
HashEngineering, line 144: This is the function that will start and stop mixing based on various inputs such as what do we want to do (mix) and whether are we connected to the Dash Network.

HashEngineering: Not fully tested yet.

wallet/src/de/schildbach/wallet/service/CoinJoinService.kt
```diff
@@ -0,0 +1,323 @@
...
+        requestDecryptedKey: RequestDecryptedKey
+    )
+
+    suspend fun prepareAndStartMixing()
+    fun startMixing(): Boolean
+    fun stopMixing()
+    fun getMixingStatus(): MixingStatus
+    suspend fun waitForMixing()
+    suspend fun waitForMixingWithException()
+}
```
HashEngineering, line 52: Some of these methods will be removed as the main is to have the implementation manage starting/stopping.

Some of these methods are leftover from previous iterations.

wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt
```diff
@@ -325,11 +337,29 @@ class CreateIdentityService : LifecycleService() {
...
+            platformRepo.updateIdentityCreationState(blockchainIdentityData, CreationState.MIXING_FUNDS)
+            coinJoinService.configureMixing(
+                Constants.DASH_PAY_FEE,
+                { encryptionKey },
+                { it.decrypt(encryptionKey) })
+
+            coinJoinService.prepareAndStartMixing()
 
+            coinJoinService.waitForMixing();
+        }
```
HashEngineering, line 362: waitForMixing() is a blocking call.  I suppose that is fine because the calling method is also a `suspend` method that is waiting on my other functions to return control before moving on to the next step.

Syn-McJ: It could still block: `suspend` method doesn't guarantee that the code will be executed in another thread - if it's called from the main thread, it'll be executed in the main thread until the context is switched.

I see that the `serviceScope` in which a lot of suspend methods are launched is created using `Dispatchers.Main` - any reason for this? Perhaps `.Unconfined` is a better choice.

HashEngineering: There is no reason for using `Dispatchers.Main` unless it is the default if we don't specify another.

HashEngineering: I am pretty sure I don't want to use `Dispatchers.Main` because I was getting some exceptions from Room DB related calls.

HashEngineering: Is there a way to specify a dispatcher in this part of the code?
```
        blockchainStateProvider.observeNetworkStatus()
            .filterNot { it == NetworkStatus.UNKNOWN }
            .onEach { status ->
                updateNetworkStatus(status)
            }
            .launchIn(coroutineScope)
 ```

Comment (Syn-McJ) [reply wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt:362]: @HashEngineering my point is that right now, the scope created with `.Main`:
`private var serviceScope = CoroutineScope(serviceJob + Dispatchers.Main)`

That means it's executed in the main thread. Perhaps `.IO` or `.Unconfined` which is the default is better here.

Comment (HashEngineering) [reply wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt:362]: Thanks.  This is what I needed to know to understand.

Comment (HashEngineering) [reply wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt:362]: Changing to `.IO` results in these exceptions.  This gives us an opportunity to move away from these `AsyncTask` methods such as `DecryptSeedTask` and use coroutines instead.

```
[DefaultDispatcher-worker-1] [UPGRADING_WALLET(error)]
java.lang.NullPointerException: Attempt to read from field 'android.os.MessageQueue android.os.Looper.mQueue' on a null object reference
	at android.os.Handler.<init>(Handler.java:257) ~[na:0.0]
	at android.os.Handler.<init>(Handler.java:162) ~[na:0.0]
	at de.schildbach.wallet.payments.DecryptSeedTask.<init>(DecryptSeedTask.java:39) ~[na:0.0]
	at de.schildbach.wallet.ui.dashpay.CreateIdentityService$decryptSeed$2$1.<init>(CreateIdentityService.kt:795) ~[na:0.0]
	at de.schildbach.wallet.ui.dashpay.CreateIdentityService.decryptSeed(CreateIdentityService.kt:795) ~[na:0.0]
	at de.schildbach.wallet.ui.dashpay.CreateIdentityService.createIdentity(CreateIdentityService.kt:339) ~[na:0.0]
	at de.schildbach.wallet.ui.dashpay.CreateIdentityService.access$createIdentity(CreateIdentityService.kt:58) ~[na:0.0]
	at de.schildbach.wallet.ui.dashpay.CreateIdentityService$createIdentity$1.invokeSuspend(Unknown Source:16) ~[na:0.0]
```

Comment (Syn-McJ) [reply wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt:362]: Could be a good idea. `DecryptSeedTask` seems to be quite simple class whose purpose is to run `decrypt` in the background. We can just launch it in a `withContext` instead. This method that wraps it in a coroutine is also unnecessary:
```
/**
* Wraps callbacks of DecryptSeedTask as Coroutine
*/
private suspend fun decryptSeed(handler: Handler, wallet: Wallet, encryptionKey: KeyParameter): DeterministicSeed 
```

Instead, we can just run it like this:
```
withContext(Dispatchers.Default) {
   val decryptedSeed = seed.decrypt(keyCrypter, null, keyParameter)
}
```

For `DeriveKeyTask`, there is an analog function in the `SecurityFunctions` that can also be run in a context.

File: wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt [hunk 4]
```diff
@@ -317,19 +305,35 @@ class CreateIdentityService : LifecycleService() {
         platformRepo.resetIdentityCreationStateError(blockchainIdentityData)
 
         val wallet = walletApplication.wallet!!
-        val password = securityGuard.retrievePassword()
-
 
-        val encryptionKey = deriveKey(backgroundHandler, wallet, password)
+        val encryptionKey = platformRepo.getWalletEncryptionKey() ?: throw IllegalStateException("cannot obtain wallet encryption key")
```
HashEngineering, line 309: > Could be a good idea. `DecryptSeedTask` seems to be quite simple class whose purpose is to run `decrypt` in the background. We can just launch it in a `withContext` instead. This method that wraps it in a coroutine is also unnecessary:
> 
> ```
> /**
> * Wraps callbacks of DecryptSeedTask as Coroutine
> */
> private suspend fun decryptSeed(handler: Handler, wallet: Wallet, encryptionKey: KeyParameter): DeterministicSeed 
> ```
> 
> Instead, we can just run it like this:
> 
> ```
> withContext(Dispatchers.Default) {
>    val decryptedSeed = seed.decrypt(keyCrypter, null, keyParameter)
> }
> ```
> 
> For `DeriveKeyTask`, there is an analog function in the `SecurityFunctions` that can also be run in a context.

I didn't follow these recommendations yet.  There was already a `PlatformRepo.getWalletEncryptionKey()` that I decided to use in `CreateIdentityService`.  And I added a similar method to get the Seed.

Ideally, though, PlatformRepo should be handling Platform functions and not wallet or security functions.  Since this PR has bene tested already, I propose that this suggestion above be part of some refactoring on the next DashPay / CoinJoin related story.

wallet/src/de/schildbach/wallet/ui/dashpay/CreateIdentityService.kt
```diff
@@ -317,19 +305,35 @@ class CreateIdentityService : LifecycleService() {
...
-            val seed = decryptSeed(backgroundHandler, wallet, encryptionKey)
-            platformRepo.addWalletAuthenticationKeysAsync(seed, encryptionKey)
+            val seed = platformRepo.getWalletSeed() ?: throw IllegalStateException("cannot obtain wallet seed")
+            platformRepo.addWalletKeyChainsAsync(seed, encryptionKey)
         }
 
         val blockchainIdentity = platformRepo.initBlockchainIdentity(blockchainIdentityData, wallet)
+        // look for the credit funding tx in case there was an error in the next step previously
+        for (tx in wallet.creditFundingTransactions) {
+            tx as CreditFundingTransaction
```
Syn-McJ, line 320: This is a weird expression. Perhaps it should be a check `if (tx is CreditFundingTransaction)`

Syn-McJ (approved):
Looks great

PR #1088, opened 2023-02-21
fix: crowdnode transaction grouping

Syn-McJ:
There is a bug that leads to internal transactions with an amount that matches the CrowdNode API amount to be considered a possible CrowdNode response.

- Only check for possible CrowdNode responses if it's an incoming transaction.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (commented):
LGTM

PR #1092, opened 2023-03-02
fix: clearing Explore prefs on wallet reset

Syn-McJ:
When the wallet is reset, Explore prefs are not cleared. This causes the data to be empty in some cases because the saved timestamp is higher than the preloaded db or remote timestamp.

- Clear Explore prefs when the wallet is reset.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
LGTM

PR #1101, opened 2023-03-24
feat(dashdirect): polling API after a puchase

Syn-McJ:
- Instead of waiting 2 seconds after gift card purchase and showing a retry dialog in case info isn't yet available, we will poll the API until we get the info, error or the user dismisses the dialog.
- BIP70 flow in the `SendCoinsTaskRunner` is refactored to make it less convoluted.
- Methods to access `PackageInfo`-related information moved from `WalletApplication` into a separate class.
- Some unused code is removed
- Some unused resources are removed

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dashdirect/DashDirectViewModel.kt
```diff
@@ -105,26 +104,31 @@ class DashDirectViewModel @Inject constructor(
...
-                    giftCardMerchant.savingsPercentage ?: DashDirectConstants.DEFAULT_DISCOUNT
-                val discountedValue = getDiscountedAmount(amountValue, savingsPercentage)
                 val response = repository.purchaseGiftCard(
                     merchantId = it,
-                    giftCardAmount = discountedValue.toBigDecimal().toDouble(),
-                    currency = Constants.DASH_CURRENCY,
-                    deviceID = UUID.randomUUID().toString(),
+                    amountUSD = amountValue.toBigDecimal().toDouble(),
+                    paymentCurrency = Constants.DASH_CURRENCY,
+                    deviceID = UUID.randomUUID().toString(), // TODO
```
Syn-McJ, line 122: I'll fix this in the next story.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dashdirect/DashDirectViewModel.kt
```diff
@@ -207,49 +191,35 @@ class DashDirectViewModel @Inject constructor(
...
-    fun saveGiftCard(
-        txId: Sha256Hash,
-        data: GetGiftCardResponse.Data
-    ) {
-        logOnPurchaseEvents()
-        val giftCardId = "${ServiceName.DashDirect.lowercase()}+${data.id ?: -1}" // e.g. dashdirect+1234
+    fun saveGiftCardDummy(txId: Sha256Hash, orderId: String, paymentId: String) {
         val giftCard = GiftCard(
-            id = giftCardId,
+            id = UUID.randomUUID().toString(),
```
Syn-McJ, line 196: Because gift card ID isn't available at the time we save the dummy record, I figure it's better to have a UUID here and separate fields for the service.

wallet/res/xml/preference_headers.xml
```diff
@@ -1,11 +0,0 @@
...
-<preference-headers xmlns:android="http://schemas.android.com/apk/res/android" >
-
-    <header
-        android:fragment="de.schildbach.wallet.ui.preference.DiagnosticsFragment"
-        android:title="@string/preferences_category_diagnostics" />
-    <header
-        android:fragment="de.schildbach.wallet.ui.preference.AboutFragment"
-        android:title="@string/about_title" />
-
-</preference-headers>
```
Syn-McJ, line 11: Doesn't look like these fragments are used now anywhere.

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -782,39 +779,6 @@ public void broadcastTransaction(final Transaction tx) {
...
-    public static String httpUserAgent(final String versionName) {
-        final VersionMessage versionMessage = new VersionMessage(Constants.NETWORK_PARAMETERS, 0);
-        versionMessage.appendToSubVer(Constants.USER_AGENT, versionName, null);
-        return versionMessage.subVer;
-    }
-
-    public String httpUserAgent() {
-        return httpUserAgent(packageInfo().versionName);
-    }
-
```
Syn-McJ, line 817: In the ongoing effort to remove dependencies on the `WalletApplication`, I moved this code into a separate service class.

wallet/src/de/schildbach/wallet/payments/RequestPaymentRequestTask.java
```diff
@@ -130,102 +122,6 @@ protected void error(Exception x, final int messageResId, final Object... messag
         }
     }
 
-    public final static class BluetoothRequestTask extends RequestPaymentRequestTask {
```
Syn-McJ, line 133: Doesn't seem to be used.

wallet/src/de/schildbach/wallet/offline/HttpDirectPaymentTask.kt
```diff
@@ -1,117 +0,0 @@
...
-import okhttp3.RequestBody
-import okio.BufferedSink
-import org.bitcoin.protocols.payments.Protos.Payment
-import org.bitcoin.protocols.payments.Protos.PaymentACK
-import org.bitcoinj.protocols.payments.PaymentProtocol
-import org.dash.wallet.common.util.Constants
-import org.slf4j.LoggerFactory
-import java.io.IOException
-
-abstract class HttpDirectPaymentTask(
```
Syn-McJ, line 36: These classes are just Kotlin copies of the java classes and as such not very useful. When we rewrite code to Kotlin we should think about how to make use of coroutines and simplify execution flow for the caller.

wallet/src/de/schildbach/wallet/payments/parsers/PaymentIntentParser.kt
```diff
@@ -0,0 +1,214 @@
...
+import java.io.InputStream
+import java.util.*
+import java.util.regex.Pattern
+
+class PaymentIntentParserException(
+    innerException: Exception,
+    val localizedMessage: ResourceString
+) : Exception(innerException)
+
+object PaymentIntentParser {
```
Syn-McJ, line 50: `InputParser` is hard to use from Kotlin and is responsible for too many things. Most callers know which parsing functionality they need and do not require all of the callbacks `InputParser` has. 

I think it's better to have several dedicated parsing classes instead.

wallet/res/xml/preference_headers.xml
HashEngineering: This is from the version 6 settings screens.

wallet/src/de/schildbach/wallet/payments/RequestPaymentRequestTask.java
HashEngineering: Bluetooth features were part of the app in prior versions, though they were probably not used by many.  Like NFC it was an alternate way to send payments or payment requests to another user's device.

wallet/src/de/schildbach/wallet/ui/preference/ResolveDnsTask.java
```diff
@@ -1,66 +0,0 @@
...
- * @author Andreas Schildbach
- */
-public abstract class ResolveDnsTask {
-    private final Handler backgroundHandler;
-    private final Handler callbackHandler;
-
-    public ResolveDnsTask(final Handler backgroundHandler) {
-        this.backgroundHandler = backgroundHandler;
-        this.callbackHandler = new Handler(Looper.myLooper());
-    }
```
HashEngineering, line 36: This may have been used for the Trusted Peer feature which didn't make from version 6 to version 7.

It is a feature that I would like to bring back, but we can always go back in time to get the code and probably rewrite most of it.

HashEngineering (approved):
Looks great.

wallet/src/de/schildbach/wallet/ui/preference/ResolveDnsTask.java
Syn-McJ: I agree. If there is a story for bringing it back, we can mention this file in there.

PR #1102, opened 2023-03-24
Feature core 19 masternode keys

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyChainAdapter.kt
+++ b/wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyChainAdapter.kt
@@ -44,10 +44,14 @@ class MasternodeKeyChainAdapter(
     }
 
     override fun getItemCount(): Int {
-        return keyChain.currentIndex + 1
+        return keyChain.issuedKeyCount + 1
     }
 
     override fun getItemId(position: Int): Long {
         return keyChain.getKey(position).hashCode().toLong()
+    }
+
+    fun addKey(key: IKey) {
+        notifyItemInserted(itemCount + 1)
     }
 }
--- a/wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyChainAdapter.kt
+++ b/wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyChainAdapter.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2023 Dash Core Group.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package de.schildbach.wallet.ui.more
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import de.schildbach.wallet_test.databinding.MasternodeKeyRowBinding
+import org.bitcoinj.crypto.IDeterministicKey
+import org.bitcoinj.crypto.IKey
+import org.bitcoinj.wallet.AuthenticationKeyChain
+import org.bitcoinj.wallet.authentication.AuthenticationKeyUsage
+
+class MasternodeKeyChainAdapter(
+    val keyChain: AuthenticationKeyChain,
+    val keyUsage: Map<IKey, AuthenticationKeyUsage>,
+    private val clickListener: (String) -> Unit
+) : RecyclerView.Adapter<MasternodeKeyViewHolder>() {
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): MasternodeKeyViewHolder {
+        val inflater = LayoutInflater.from(parent.context)
+        val binding = MasternodeKeyRowBinding.inflate(inflater, parent, false)
+        return MasternodeKeyViewHolder(binding)
+    }
+
+    override fun onBindViewHolder(holder: MasternodeKeyViewHolder, position: Int) {
+        val key = keyChain.getKey(position)
+        holder.bind(key, keyUsage.get(key), clickListener)
+    }
+
+    override fun getItemCount(): Int {
+        return keyChain.currentIndex + 1
+    }
+
+    override fun getItemId(position: Int): Long {
+        return keyChain.getKey(position).hashCode().toLong()
+    }
+}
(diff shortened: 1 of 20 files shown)


wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyViewHolder.kt
```diff
@@ -0,0 +1,77 @@
...
+            keypairUsage.setText(
+                when (usage?.status) {
+                    AuthenticationKeyStatus.CURRENT -> R.string.masternode_key_used
+                    AuthenticationKeyStatus.REVOKED -> R.string.masternode_key_revoked
+                    AuthenticationKeyStatus.PREVIOUS -> R.string.masternode_key_not_used
+                    AuthenticationKeyStatus.UNKNOWN,
+                    AuthenticationKeyStatus.NEVER,
+                    null,
+                    -> R.string.masternode_key_not_used
+                },
```
HashEngineering, line 50: The ktlint extension was used to format this file.

the `when` looks strange when covering several items.

wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyViewHolder.kt
```diff
@@ -0,0 +1,77 @@
...
+            )
+            // set all of the key serializations
+            address.text = Address.fromKey(Constants.NETWORK_PARAMETERS, masternodeKey).toBase58()
+            keyId.text = Utils.HEX.encode(masternodeKey.pubKeyHash)
+            publicKey.text = Utils.HEX.encode(masternodeKey.pubKey)
+            privateKeyHex.text = "encrypted" // Utils.HEX.encode(masternodeKey.privKeyBytes)
+            privateKeyWif.text = "encrypted" // masternodeKey.getPrivateKeyAsWiF(Constants.NETWORK_PARAMETERS)
```
HashEngineering, line 57: Haven't figured out how to decrypt the private key yet.

HashEngineering: Or rather where in the code  it should be decrypted.

wallet/src/de/schildbach/wallet/ui/more/MasternodeKeysFragment.kt
```diff
@@ -0,0 +1,97 @@
...
+        binding.keyTypeList.setHasFixedSize(true)
+        if (viewModel.hasMasternodeKeys()) {
+            loadKeyTypes()
+        } else {
+            CheckPinDialog() { password ->
+                lifecycleScope.launch {
+                    viewModel.addKeyChains(password!!)
+                    loadKeyTypes()
+                }
+            }.showNow(childFragmentManager, "check-pin")
```
HashEngineering, line 80: this may be temporary, but we do need to add the authentication  (masternode) keychains to the Wallet somewhere.  We can and should also add them when a wallet is created.  But for existing wallets, we need to choose a place and time to add them: either here or the first time the app starts after being upgraded to the version that supports the Masternode Keys feature.

In the past, we asked for the user's pin before "upgrading" the wallet to support masternode keychains.  In DashPay, we don't do that when creating a username which must also "upgrade" the Wallet to support identity keychains.

wallet/src/de/schildbach/wallet/ui/more/MasternodeKeyChainAdapter.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.bitcoinj.crypto.IDeterministicKey
+import org.bitcoinj.crypto.IKey
+import org.bitcoinj.wallet.AuthenticationKeyChain
+import org.bitcoinj.wallet.authentication.AuthenticationKeyUsage
+
+class MasternodeKeyChainAdapter(
+    val keyChain: AuthenticationKeyChain,
+    val keyUsage: Map<IKey, AuthenticationKeyUsage>,
+    private val clickListener: (String) -> Unit
+) : RecyclerView.Adapter<MasternodeKeyViewHolder>() {
```
HashEngineering, line 33: Currently, instead of holding a list of items to display, this adapter is holding a keychain class, which holds a list of keys.  Those keys are all encrypted, so the private keys cannot be displayed.

More work needs to be done on this class to make it meet the requirements of the User Story.  The way DashJ KeyChains keep track of things doesn't line up with the Story requirements, yet.

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -324,6 +330,18 @@ public void setWallet(Wallet newWallet) {
...
+                    wallet.getParams(),
+                    wallet.getKeyChainSeed(),
+                    EnumSet.of(
+                            AuthenticationKeyChain.KeyChainType.MASTERNODE_OWNER,
+                            AuthenticationKeyChain.KeyChainType.MASTERNODE_VOTING,
+                            AuthenticationKeyChain.KeyChainType.MASTERNODE_OPERATOR,
+                            AuthenticationKeyChain.KeyChainType.MASTERNODE_PLATFORM_OPERATOR
+                    )
+            );
+        }
```
HashEngineering, line 344: We should separate wallet creation into its own class.  We have many classes that create wallet objects based on what the user is doing.  And there are many static helper methods.  Currently there is wallet creation code sprinkled about in many places and probably inconsistently.

wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -1101,4 +1119,12 @@ public void checkSendingConditions(
...
         );
     }
+
+    public void clearExtensions() {
+        authenticationGroupExtension = new AuthenticationGroupExtension(Constants.NETWORK_PARAMETERS);
+    }
+
+    public WalletExtension[] getWalletExtensions() {
+        return new WalletExtension[] {authenticationGroupExtension};
+    }
```
HashEngineering, line 1129: this would also go into a wallet creation/management class.

wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeyChainFragment.kt
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package de.schildbach.wallet.ui.more.masternode_keys
+
```
HashEngineering, line 18: All masternode key classes moved to `masternode_key` package.  

Ktlint (the Android Studio Extension) doesn't like underscore in the names!

wallet/src/de/schildbach/wallet/util/WalletUtils.java
```diff
@@ -117,11 +118,16 @@ public static Editable formatHash(@Nullable final String prefix, final String ad
     }
 
     public static Wallet restoreWalletFromProtobufOrBase58(final InputStream is,
-            final NetworkParameters expectedNetworkParameters) throws IOException {
+                                                           final NetworkParameters expectedNetworkParameters) throws IOException {
+        return restoreWalletFromProtobufOrBase58(is, expectedNetworkParameters, null);
+    }
```
HashEngineering, line 123: In a future story, we could refactor this class and its static methods into a Wallet Creation/Management Class that gets injected into the right places.

Syn-McJ (approved):
Looks great!

PR #1104, opened 2023-03-28
feat(dashdirect): no network indicator

Syn-McJ:
We want to show an indicator on the Purchase screen when there is no network available. This is similar to what we already have in Coinbase.

- The no network indicator is added.
- Explore Devnet and some min/max/discount changes are reverted - we won't have them in the database.
- New DataStore config for Explore prefs is created and old Explore prefs are migrated to this config.
- Some refactoring for the various DataStore configs (crowdnode/explore/dashdirect).
- Some `EncryptionProvider` refactoring to combine the one we're using in the `wallet` and `explore` modules.
- Some UI fixes.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/data/BaseConfig.kt
```diff
@@ -0,0 +1,142 @@
...
+                emit(emptyPreferences())
+            } else {
+                throw exception
+            }
+        }
+
+    init {
+        walletDataProvider.attachOnWalletWipedListener {
+            @OptIn(DelicateCoroutinesApi::class)
+            GlobalScope.launch { clearAll() }
```
Syn-McJ, line 72: Unfortunately, DataStore prefs remain in memory even if we delete the file. So if the config is initialized already, we need to listen to the wipe event and clear it manually.

common/src/main/java/org/dash/wallet/common/di/NetworkStateProviderModule.kt
```diff
@@ -1,44 +0,0 @@
...
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-package org.dash.wallet.common.di
```
Syn-McJ, line 18: Generally, we should only keep the interfaces in `common` and provide injections in the `wallet`.

common/src/main/java/org/dash/wallet/common/livedata/NetworkState.kt
```diff
@@ -1,124 +0,0 @@
...
-        }
-    }
-}
-
-/**
- * @author Kebab Krabby
- * https://stackoverflow.com/questions/36421930/connectivitymanager-connectivity-action-deprecated
- */
-@Deprecated ("Use the NetworkState service", ReplaceWith("NetworkState"))
-class ConnectionLiveData(val context: Context) : LiveData<Boolean>() {
```
Syn-McJ, line 76: I don't think we need dedicated live data, instead, we can do 
`networkState.isConnected.asLiveData()`

common/src/main/java/org/dash/wallet/common/ui/ConnectivityViewModel.kt
```diff
@@ -1,44 +0,0 @@
...
-import androidx.lifecycle.viewModelScope
-import dagger.hilt.android.lifecycle.HiltViewModel
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.launch
-import org.dash.wallet.common.livedata.NetworkStateInt
-import javax.inject.Inject
-
-@HiltViewModel
-open class ConnectivityViewModel @Inject constructor(private val networkStateProvider: NetworkStateInt)
```
Syn-McJ, line 31: I don't think we need this class. It handles some boilerplate that we can just get rid of instead.

common/src/main/java/org/dash/wallet/common/util/security/EncryptionProvider.kt
```diff
@@ -1,83 +1,33 @@
...
-        if (encryptionIv == null) {
-            cipher.init(Cipher.ENCRYPT_MODE, secretKey)
-            saveIv(cipher.iv)
-        } else {
-            val spec = GCMParameterSpec(128, encryptionIv)
-            cipher.init(Cipher.ENCRYPT_MODE, secretKey, spec)
-        }
-        return cipher.doFinal(textToEncrypt.toByteArray(StandardCharsets.UTF_8))
-    }
+interface EncryptionProvider {
```
Syn-McJ, line 24: Keeping the interface only in `common` and providing the implementation in the `wallet`.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DashDirectRepository.kt
```diff
@@ -62,6 +63,7 @@ class DashDirectRepository @Inject constructor(
                 createUser(email)
             }
             config.setSecuredData(DashDirectConfig.PREFS_KEY_DASH_DIRECT_EMAIL, email)
+            config.set(DashDirectConfig.PREFS_DEVICE_UUID, UUID.randomUUID().toString())
```
Syn-McJ, line 66: Generating a device UUID when the user signed up instead of every purchase.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DashDirectRepository.kt
```diff
@@ -92,35 +95,33 @@ class DashDirectRepository @Inject constructor(
                 }
             }
             it?.data?.accessToken?.let { token ->
-                config.setPreference(DashDirectConfig.PREFS_KEY_LAST_DASH_DIRECT_ACCESS_TOKEN, token)
+                config.setSecuredData(DashDirectConfig.PREFS_KEY_ACCESS_TOKEN, token)
```
Syn-McJ, line 98: I think the access token should also be stored securely.

wallet/src/de/schildbach/wallet/Constants.java
```diff
@@ -74,7 +74,7 @@ public final class Constants {
                 IS_PROD_BUILD = true;
                 FILENAME_NETWORK_SUFFIX = "";
                 WALLET_NAME_CURRENCY_CODE = "dash";
-                org.dash.wallet.common.util.Constants.INSTANCE.setEXPLORE_GC_FILE_PATH("explore/explore-devnet.db");
+                org.dash.wallet.common.util.Constants.INSTANCE.setEXPLORE_GC_FILE_PATH("explore/explore.db");
```
HashEngineering, line 77: This reverts to the 1.0 database schema.

wallet/src/de/schildbach/wallet/security/ModernEncryptionProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+import java.nio.charset.StandardCharsets
+import java.security.GeneralSecurityException
+import java.security.KeyStore
+import java.security.KeyStoreException
+import javax.crypto.Cipher
+import javax.crypto.KeyGenerator
+import javax.crypto.SecretKey
+import javax.crypto.spec.GCMParameterSpec
+
+class ModernEncryptionProvider(
```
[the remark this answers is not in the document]
HashEngineering: I wanted to have some unit tests for this class.  This class is a very critical part of the wallet as it controls encrypting and decrypting of the wallet keys.  I will make a separate ticket for this.

While this class is identical in functionality, it makes me feel better if we had some tests that verify that the functionality is identical beyond what QA does.  I am not sure if we can easily use OpenSSL to create such tests or if we need to use the java implementation on the master branch.

HashEngineering (approved):

wallet/src/de/schildbach/wallet/security/ModernEncryptionProvider.kt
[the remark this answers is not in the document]
Syn-McJ: Yes, good idea. We can keep the java class around for tests.

Some Explore tests are failing in this PR, it seems like it will take time to fix them. We can do it in the same story (or separate, doesn't matter).

PR #1106, opened 2023-04-02
feat: redesign passphrase backup [NMA-1484]

Syn-McJ:
We want to remove an extra info screen from the "backup passphrase" flow and add a "success" screen.
Some redesign is also required.

- Redesigned Info screen and added Success screen.
- Refactored `VerifyActivity` to use a navigation controller.
- Moved `SecurityActivity` to a `SecurityFragment`.
- Combined `ViewSeedActivity` and `VerifyWriteDownFragment` since they are the same.
- Refactored `DecryptSeedLiveData` to move this logic into `SecurityFunctions`.
- Removed some unused resources.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/livedata/DecryptSeedLiveData.kt
```diff
@@ -1,88 +0,0 @@
...
-                            }
-                        }
-                        value = Resource.loading(null)
-                        decryptSeedTask!!.decryptSeed(wallet.keyChainSeed, wallet.keyCrypter, encryptionKey)
-                    }
-                }
-            }
-            value = Resource.loading(null)
-            deriveKeyTask!!.deriveKey(wallet, password)
-        }
```
Syn-McJ, line 86: I think it's better to have this logic as a couple of `suspend` calls in the `SecurityFunctions` class. That way, we can use it from anywhere - not just the UI thread.

wallet/src/de/schildbach/wallet/ui/main/WalletActivity.java
```diff
@@ -148,14 +147,6 @@ protected void onResume() {
...
         detectUserCountry();
-        showBackupWalletDialogIfNeeded();
-    }
-
-    private void showBackupWalletDialogIfNeeded() {
-        if (showBackupWalletDialog) {
-            BackupWalletDialogFragment.show(getSupportFragmentManager());
-            showBackupWalletDialog = false;
-        }
     }
```
Syn-McJ, line 150: Looks like we don't need this anymore. `showBackupWalletDialog` is never set to `true`.

wallet/src/de/schildbach/wallet/ui/backup/BackupWalletDialogFragment.java
```diff
@@ -253,17 +247,12 @@ public void onDismiss(final DialogInterface dialog) {
...
         showView.setOnCheckedChangeListener(null);
+        alertDialog.dismiss();
 
         wipePasswords();
         super.onDismiss(dialog);
     }
 
-    @Override
-    public void onCancel(final DialogInterface dialog) {
-        activity.finish();
```
Syn-McJ, line 263: Not sure why we were finishing `SecurityActivity` on the dialog dismissal. Seems like just dismissing is fine.

wallet/src/de/schildbach/wallet/ui/DecryptSeedWithPinDialog.kt
```diff
@@ -15,49 +16,50 @@ import org.bitcoinj.wallet.DeterministicSeed
  */
 @AndroidEntryPoint
 class DecryptSeedWithPinDialog(
-    private var onSuccessOrDismiss: ((DeterministicSeed?) -> Unit)?
+    private var onSuccessOrDismiss: ((Array<String>) -> Unit)?
```
HashEngineering, line 19: This is an interesting change.  We must not need the `DeterministicSeed` object, only the phrase.

wallet/src/de/schildbach/wallet/ui/backup/BackupWalletDialogFragment.java
HashEngineering: This was taken from Bitcoin Wallet, where I think an activity was created that displays the backup wallet dialog.  

However, we are not creating an activity to only display this dialog:

Reference:
https://github.com/bitcoin-wallet/bitcoin-wallet/blob/main/wallet/src/de/schildbach/wallet/ui/backup/BackupWalletActivity.java#L28-L40

wallet/src/de/schildbach/wallet/ui/main/WalletActivity.java
HashEngineering: Correct.

HashEngineering (approved):
LGTM.

PR #1108, opened 2023-04-06
feat: redesign balance header [NMA-1483]

Syn-McJ:
We want to update the UI of our balance header on the Home screen.

- Updated the UI.
- Created a datastore config for UI settings and migrated "autohide balance" preference.
- Added a way to scroll back on top of the tx history by tapping "Home" button or the Dash logo.
- Removed some unused code.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/ui/CurrencyTextView.java
```diff
@@ -45,25 +41,14 @@ public class CurrencyTextView extends AppCompatTextView {
     private MonetaryFormat format = null;
     private boolean alwaysSigned = false;
     private RelativeSizeSpan prefixRelativeSizeSpan = null;
-    private ScaleXSpan prefixScaleXSpan = null;
-    private ForegroundColorSpan prefixColorSpan = null;
     private RelativeSizeSpan insignificantRelativeSizeSpan = null;
     private boolean applyMarkup = true;
-    private int prefixColor = 0;
```
Syn-McJ, line 52: We never use this.
Given that Rostislav wants to remove the variable font size in the currency texts, we might not need this class at all in the future. But for now, I just cleaned up the prefix code.

common/src/main/res/layout/app_bar_main.xml
```diff
@@ -1,49 +0,0 @@
...
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            tools:text="Synchronizing with network. 3 months behind."
-            android:textColor="@color/white"
-            app:fontFamily="@font/inter_regular"
-            android:textSize="12sp" />
-
-    </RelativeLayout>
-
-</com.google.android.material.appbar.AppBarLayout>
```
Syn-McJ, line 49: This is some relic that's only used in sevaral two-pane layouts. Not sure if we still need those layouts (it's probably better to handle tablet UI separately), but I replaced `app_bar_main` with `app_bar_general` there for now.

HashEngineering (approved):
Looks good.

Great job at removing legacy code from older versions.

PR #1114, opened 2023-04-20
feat(dashpay): platform sync scheduling [NMA-1411]

Syn-McJ:
We want to publish tx metadata at a random time in the future.

- Moved platform push out of `updateTransactionMetadata` and added a randomized time cutoff for pushing.
- Removed some unnecessary `async` DAOs. All DB access methods are either `suspend` or return `Flow`.
- Cleaned up some coroutine work around `PlatformSyncService`.
- Combined metadata table update to a single call.
- Added a datastore config for DashPay. Migrated some prefs.
- Some DI-related refactoring.
- Applied `ktlint` here and there.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/Extensions.kt
```diff
@@ -1,46 +0,0 @@
...
-        curContext = (curContext as ContextWrapper).baseContext
-    }
-    return if (curContext is LifecycleOwner) {
-        curContext
-    } else {
-        null
-    }
-}
-
-fun <T> LiveData<T>.observeOnce(lifecycleOwner: LifecycleOwner, observer: Observer<T>) {
```
Syn-McJ, line 39: These extensions are probably not needed anymore. `observeOnce` is replaced with `suspend` calls and the rest aren't used.

wallet/src/de/schildbach/wallet/service/WalletTransactionMetadataProvider.kt
```diff
@@ -121,17 +121,26 @@ class WalletTransactionMetadataProvider @Inject constructor(
...
+
+        if (existing != null) {
             log.info("txmetadata for $txId exists, only do update")
-            update()
+            update(existing)
         } else {
             log.info("txmetadata for $txId does not exist, perform insert, then update")
-            insertTransactionMetadata(txId, isSyncingPlatform)
-            update()
+            insertTransactionMetadata(txId, isSyncingPlatform)?.let { update(it) }
```
Syn-McJ, line 143: These changes are lifted from DashDirect branch.

.github/workflows/manual_distribution.yml
```diff
@@ -121,9 +121,6 @@ jobs:
         gem install google-cloud-storage
         rm  -rf wallet/assets/explore/*
         ruby ./exploredata.rb "${{ github.event.inputs.flavor }}"
-
-    - name: Test
-      run: bundle exec fastlane test flavor:"${{ github.event.inputs.flavor }}" type:"${{ github.event.inputs.type }}" version_code:"${{ env.build_number }}" store_pass:"${{ secrets.SIGNING_STORE_PASS }}"
```
Syn-McJ, line 126: Something wrong with the tests, cannot even build them. Will check later.

Syn-McJ: Fixed

wallet/src/de/schildbach/wallet/ui/RestoreWalletFromFileViewModel.kt
```diff
@@ -16,24 +16,30 @@
...
 import org.bitcoinj.wallet.Wallet
 import org.slf4j.LoggerFactory
+import javax.inject.Inject
 
-class RestoreWalletFromFileViewModel(application: Application) : AndroidViewModel(application) {
+@HiltViewModel
+class RestoreWalletFromFileViewModel @Inject constructor(
+    private val walletApplication: WalletApplication,
+    private val dashPayConfig: DashPayConfig
+) : ViewModel() {
```
HashEngineering, line 39: hopefully this doesn't cause trouble with merging with master

HashEngineering (approved):
Looks great.

wallet/src/de/schildbach/wallet/data/TransactionMetadataDocumentDao.kt
```diff
@@ -37,60 +37,51 @@ interface TransactionMetadataDocumentDao {
...
-            FROM transaction_metadata_platform 
-            WHERE txId = :txId
-                AND memo is NOT NULL
-        )
-        """)
+        WHERE txId = :txId 
+            AND memo is NOT NULL
+        ORDER BY timestamp DESC
+        LIMIT 1
+    """)
```
Syn-McJ, line 54: The old query had a problem: it didn't filter by `txId` correctly because the filtering was only done inside the subquery, and the main query was returning any row that would match `WHERE timestamp =` condition. This caused several transactions to display the same memo.

The old query could also be modified like this, but the simplified version should be slightly more performant:

```
@Query("""
    SELECT memo 
    FROM transaction_metadata_platform 
    WHERE timestamp = (
        SELECT MAX(timestamp) 
        FROM transaction_metadata_platform 
        WHERE txId = :txId
            AND memo is NOT NULL
    ) AND txId = :txId
""")
```

PR #1116, opened 2023-04-21
Feature core 19 masternode keys fixes

HashEngineering:

Some improvements, fixes and changes to what is displayed for each key type

This will be merged to the masternode-keys branch, which will be merged to the feature-core-19 branch before being merged to master.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeyViewHolder.kt
```diff
@@ -41,7 +43,7 @@ class MasternodeKeyViewHolder(val binding: MasternodeKeyRowBinding) : RecyclerVi
             keypairUsage.text = when (usage?.status) {
                 AuthenticationKeyStatus.CURRENT -> {
                     val ipAddress = if (usage.address != null) {
-                        usage.address?.addr.toString()
+                        usage.address?.addr.toString().removePrefix("/")
```
HashEngineering, line 46: DashJ puts a `/` in front of IP addresses

build.gradle
```diff
@@ -3,7 +3,7 @@ buildscript {
         kotlin_version = '1.6.10'
         coroutines_version = '1.6.1'
         ok_http_version = '4.9.1'
-        dashjVersion = '19.0-SNAPSHOT'
+        dashjVersion = '19.0-MNKEY-SNAPSHOT'
```
HashEngineering, line 6: current a special "branch" of dashj is required

wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeysViewModel.kt
```diff
@@ -40,25 +53,60 @@ import javax.inject.Inject
...
+                } else {
+                    false
+                }
+            }
+            if (masternodeKeysUsed > 0) {
+                initKeyChainInfo()
+                _newKeysUsed.value = true
+            } else {
+                _newKeysUsed.value = false
+            }
```
HashEngineering, line 108: Here we are using liveData to tell the UI to update based on any changes in masternode key usage found during blockchain sync.

wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeysViewModel.kt
```diff
@@ -40,25 +53,60 @@ import javax.inject.Inject
...
+        get() = _newKeysUsed
 
     init {
         if (authenticationGroup.hasKeyChains()) {
             initKeyChainInfo()
         }
+
+        walletData.observeAuthenticationKeyUsage()
+            .onEach { usage -> refreshKeyChains(usage) }
+            .launchIn(viewModelScope)
```
HashEngineering, line 86: This is where we observe masternode key usage changes.

wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeyChainFragment.kt
```diff
@@ -64,26 +64,33 @@ class MasternodeKeyChainFragment : Fragment(R.layout.fragment_masternode_key_cha
...
+                (binding.keyList.layoutManager as LinearLayoutManager).scrollToPositionWithOffset(
+                    position,
+                    0
+                )
+            }
+        }
+        viewModel.newKeysFound.observe(viewLifecycleOwner) { isAdded ->
+            if (isAdded == true) {
+                masternodeKeyChainAdapter.notifyDataSetChanged()
+            }
```
HashEngineering, line 93: Here the we can react to the event and update the UI.

Syn-McJ: The usage of LiveData is fine here since the fragment observes an asynchronous event. However, an improvement can be made to eliminate the `notifyDataSetChanged` calls, which are not very performant on large lists since it forces the `RecycleView` to redraw everything.

Instead of observing "listWasChanged" event which is represented by `newKeysFound` livedata here and then notifying the adapter, the fragment can observe the list itself and then provide it to the adapter. 
The viewModel can hold a `LiveData<List<MasternodeKeyInfo>>` or another `data class` that holds all the necessary info for each row. This list is updated in the viewModel when necessary, and the fragment submits it to the adapter when the list is changed, for example in the `TransactionGroup` classes:

```
// In the viewModel
private val _transactions = MutableLiveData<List<TransactionRowView>>()
val transactions: LiveData<List<TransactionRowView>>
    get() = _transactions

// In the fragment
viewModel.transactions.observe(viewLifecycleOwner) { transactions ->
    adapter.submitList(transactions)
}
```

Then the adapter can derive from `ListAdapter` to have a `DiffCallback`, like in the `TransactionAdapter`.
The `DiffCallback` will have `areItemsTheSame` method which compares unique item IDs (for example, the `masternodeKey` field)  and `areContentsTheSame` method which will compare all of the contents (with `data class` it's easy - just check if they are equal). 

This will ensure that the `RecycleView` only redraws the changed rows and the changed content in those rows. If it sees that a new unique ID is appeared in the list (or removed), it will add/remove a new viewHolder to the list. If it sees that no items are added/removed, but some item have their contents changed (because the `data class` have changed fields), it will update the content in that row specifically.

Another advantage of this approach is that the adapter doesn't need to know much about the specifics of our MasternodeKeys feature, and only interacts with an abstract representation of the row view itself.

wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeysViewModel.kt
Syn-McJ: Since the `refreshKeyChains` is executed in `viewModelScope`, it will run on UI thread. This is probably fine if the method executes fast.
If not, it's better to `launchIn` another scope, for example, a custom working scope that this viewModel will hold like this:

```
private val workerJob = SupervisorJob()
val viewModelWorkerScope = CoroutineScope(Dispatchers.IO + workerJob) 
```

Syn-McJ (approved):
Looks good. Some potential improvements, but not essential.

PR #1118, opened 2023-04-28
fix: request PIN code for balance if launched from the lock screen

Syn-McJ:
See title.

- Added PIN code check.
- Fixed Dash format for the balance header - it was showing trailing zeros which we don't want.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
LGTM

PR #1121, opened 2023-05-09
fix: core 19 masternode list sync speed and update translations

HashEngineering:

Remove the rather old mnlistdiff.dat bootstrap file.  This is to avoid problems with the Core 19 upgrade.  Later, a new bootstrap should be added after Core 19 upgrade is complete.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


market/market-promo-text-sv.txt
```diff
@@ -1 +1 @@
-Dina egna dash, i din egen ficka!
\ No newline at end of file
+Your own Dash, in your own pocket!
```
Syn-McJ, line 1: All of the market promos are changed from translated to English versions. Not sure if this is a good idea. Where is it used exactly?

Syn-McJ (approved):
Looks good, I just have a minor concert about `market/market-promo-text.txt` translations.

market/market-promo-text-sv.txt
HashEngineering: This is not used anywhere in the app.  Text Files have to be manually imported into the Google Play Store listings.  The promo-text file probably wasn't used by us at any time.

The other text files are more concerning.  

Also, there are some empty XML files.  

I wonder if something changed with the Transifex Client.  Is it working differently?  Why would we be getting English in what should be translated in other languages?

HashEngineering: I took all of these changes out since they were from inactive translations.

PR #1131, opened 2023-06-08
chore(dashdirect): retries in case of error

Syn-McJ:
Given some timing issues DashDirect API has, we want to retry in the case PaymentStatus request returns an error. Subsequent requests might return a different result.

- Initial delay increased to 1s
- Ticker delay increased to 1.5s
- Retry counter added for 3 retries in the case of an error

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
looks good.

PR #1135, opened 2023-06-16
chore: upgrade low storage warning

Syn-McJ:
Our "low storage" warning uses deprecated APIs and needs upgrading.

- Upgraded to the new storage APIs for Android 8+
- Restored Time Skew dialog
- Removed other unused dialogs
- Fixed duplicated MainViewModel instance
- Upgraded some dependencies

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/main/MainViewModel.kt
+++ b/wallet/src/de/schildbach/wallet/ui/main/MainViewModel.kt
@@ -20,7 +20,6 @@ package de.schildbach.wallet.ui.main
 import android.content.ClipDescription
 import android.content.ClipboardManager
 import android.content.SharedPreferences
-import android.util.Log
 import androidx.annotation.VisibleForTesting
 import androidx.core.os.bundleOf
 import androidx.lifecycle.*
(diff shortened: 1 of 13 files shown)


wallet/src/de/schildbach/wallet/ui/main/MainViewModel.kt
```diff
@@ -240,6 +242,19 @@ class MainViewModel @Inject constructor(
         }
     }
 
+    suspend fun getDeviceTimeSkew(): Long {
+        return try {
+            val systemTimeMillis = System.currentTimeMillis()
+            val result = Constants.HTTP_CLIENT.head("https://www.dash.org/")
+            val networkTime = result.headers.getDate("date")?.time
```
Syn-McJ, line 249: Here we're comparing the device's time with the time dash.org returns in headers.

Initially, I tried to request time from an NTP server and parse the binary response with some ChatGPT-assisted code, but it proved to be not reliable. For a more reliable solution, we would probably need to use a dedicated NTP library, but it might be overkill.

wallet/src/de/schildbach/wallet/ui/main/WalletTransactionsFragment.kt
```diff
@@ -49,7 +49,7 @@ class WalletTransactionsFragment : Fragment(R.layout.wallet_transactions_fragmen
         private const val HEADER_ITEM_TAG = "header"
     }
 
-    private val viewModel by viewModels<MainViewModel>()
+    private val viewModel by activityViewModels<MainViewModel>()
```
Syn-McJ, line 52: This was creating a separate MainViewModel instance, so we kept two in memory. Not good.

HashEngineering (approved):
Excellent work.

PR #1139, opened 2023-06-19
fix(explore): cleanup old databases

Syn-McJ:
Old Explore databases aren't being deleted after a sync.

- Timestamp suffix and associated logic is removed
- Journal mode set to TRUNCATE to avoid creating write-ahead logging files
- Previous database files are cleaned up before an update

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
LGTM

PR #1141, opened 2023-06-21
fix: use scrCompat to prevent crashes on api 23

Syn-McJ:
There are some icons with gradient (e.g. new Backup screen) that cause crashes on API level 23.
We need to use srcCompat for these.

- Use `srcCompat` for the icons with gradient
- Simplify some icons by only keeping the background gradient

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Looks good to me.

PR #1143, opened 2023-06-21
fix: send coins crash

Syn-McJ:
When the system restores SendCoinsFragment after killing the app process, the payment intent ends up being not initialized.

- Re-initialize the payment intent from the saved state.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.kt
+++ b/wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.kt
@@ -85,9 +85,9 @@ class SendCoinsFragment: Fragment(R.layout.send_coins_fragment) {
             requireActivity().finish()
         }
 
-        if (savedInstanceState == null) {
-            viewModel.initPaymentIntent(args.paymentIntent)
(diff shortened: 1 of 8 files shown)


wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.kt
```diff
@@ -106,6 +106,10 @@ class SendCoinsFragment: Fragment(R.layout.send_coins_fragment) {
                 .add(R.id.enter_amount_fragment_placeholder, fragment)
                 .commitNow()
             enterAmountFragment = fragment
+        } else {
+            enterAmountFragment = childFragmentManager.findFragmentById(
+                R.id.enter_amount_fragment_placeholder
+            ) as EnterAmountFragment
```
Syn-McJ, line 112: If `savedInstanceState` isn't null we still want to call `initPaymentIntent` to restore the intent from the saved state, although it is still not necessary to do a `EnterAmountFragment` transaction - only to restore it into the field value.

HashEngineering (approved):
impressive work

PR #1154, opened 2023-07-09
fix: render the platform private/public key (base64) correctly

HashEngineering:
* The last byte was always set to zero (0x00)

Fix the HPMN Operator private / public keypair rendering in base64 format.
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/more/masternode_keys/MasternodeKeysViewModel.kt
```diff
@@ -309,7 +309,7 @@ class MasternodeKeysViewModel @Inject constructor(
         val bytes = ByteArray(64)
         decryptedKey.privKeyBytes.copyInto(bytes, 0, 0, 32)
         // public key bytes have a 0x00 prefix byte that is ignored
-        decryptedKey.pubKey.copyInto(bytes, 32, 1, 32)
+        decryptedKey.pubKey.copyInto(bytes, 32, 1, 33)
```
HashEngineering, line 312: `pubKey` is 33 bytes. In this format, we need the last 32 bytes copied.  The parameters for `copyInto` are indexes, rather than lengths.

Syn-McJ (approved):
Looks good

PR #1174, opened 2023-07-27
fix: Arabic decimals

Syn-McJ:
In Arabic locales, there is a different comma character used as a decimal separator, which is distinct from the comma in other locales. This causes a crash in the Amount Input control.

- Add Arabic comma in `formatFiatWithoutComma` logic
- Some other RTL fixes

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
```diff
@@ -129,15 +129,14 @@ object GenericUtils {
...
         return if (fiatValueContainsCommaWithDecimal) {
-            fiatValue.replace(
-                ",".toRegex(),
-                ""
-            )
+            fiatValue.replace(",", "")
         } else {
-            fiatValue.replace(",".toRegex(), ".")
+            fiatValue.replace(",", ".")
+                .replace("٫", ".")
```
Syn-McJ, line 139: At first, I wanted to parse the input in a "proper" manner with the current device locale, however, this proved problematic because `formatFiatWithoutComma` often receives input in a US formatted manner.

At some point, we might want to unify the way decimal separators are presented to the users in different locales, but it will require a bit of work and extensive testing. For now, this fix should suffice.

HashEngineering (approved):

PR #1187, opened 2023-08-18
fix: stuck currency code & cleanup obsolete currencies

Syn-McJ:
When the user opens Receive screen, the currency code is set, and changing it from the settings won't have an effect until relaunch.
Additionally, we need to filter out obsolete currencies and not save them in the database.

- Selected currency code is moved from `Configuration` to the `WalletUIConfig` datastore.
- A bunch of refactoring related to the above change - e.g. observing local currency instead of just fetching it.
- Clean up obsolete currencies during the migration from shared prefs to datastore.
- Move currency detection from the activity to the viewModel.
- Replace the old Currency Picker in the settings with the new dialog version.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/data/WalletUIConfig.kt
```diff
@@ -0,0 +1,122 @@
...
+            // previous versions of the app (prior to 7.3.3) may have stored an obsolete
+            // currency code in the preferences.  Let's change to the most up to date.
+            val sharedPreferences = context.getSharedPreferences(sharedPreferencesName, Context.MODE_PRIVATE)
+            val currentValue = sharedPreferences.getString(WalletUIConfig.SELECTED_CURRENCY.name, null)
+            currentValue?.let {
+                val fixedValue = CurrencyInfo.getUpdatedCurrency(CurrencyInfo.getOtherName(currentValue))
+                sharedPreferences.edit().putString(WalletUIConfig.SELECTED_CURRENCY.name, fixedValue).apply()
+            }
+            // The database might have obsolete currencies as well
+            exchangeRates.cleanupObsoleteCurrencies()
```
Syn-McJ, line 110: This will run once on migration and ensure we never have obsolete currency names in the datastore. 
This way, we don't have to call `getOtherName` or `getUpdatedCurrency` every time.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
```diff
@@ -48,56 +44,6 @@ object GenericUtils {
...
-     * For currencies used by multiple countries, we set a locale with any country using the currency
-     * If the currentCurrencySymbol equals the currency code, we just use the currency code, otherwise we
-     * get the symbol
-     * @param currencyCode
-     * @return
-     */
-    fun setCurrentCurrencySymbolWithCode(currencyCode: String): String? {
-        var currentLocale = Locale("", "")
-        var currentCurrencySymbol: String? = ""
-        when (currencyCode.lowercase()) {
```
HashEngineering, line 67: Was this method unnecessary or handled by standard android functions?

HashEngineering (approved):
Looks great.  Before merging, I want final confirmation from the CTX team about their server status.

common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
Syn-McJ: This was only used in the old `EnterAmountFragment` to figure out where the Dash symbol should go; we use `isCurrencyFirst` for it now which makes use of Androids `NumberFormat`.

integrations/coinbase/src/main/java/org/dash/wallet/integration/coinbase_integration/viewmodels/CoinbaseConvertCryptoViewModel.kt
```diff
@@ -103,7 +102,7 @@ class CoinbaseConvertCryptoViewModel @Inject constructor(
 
         val tradesRequest = TradesRequest(
             valueToConvert.toFormattedStringNoCode(),
-            userPreference.exchangeCurrencyCode!!,
+            walletUIConfig.get(WalletUIConfig.SELECTED_CURRENCY) ?: Constants.USD_CURRENCY,
```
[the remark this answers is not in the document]
Syn-McJ: Yes, that'd be better

HashEngineering (approved):

PR #1188, opened 2023-08-18
fix: avoid leaking balance info by error message or continue button

Syn-McJ:
When the user launches Send screen from the locked state, the balance info can still be leaked because we show an "Insufficient funds" error and block the Continue button.

- Don't show the error until the user authenticates.
- Don't block the button until the user authenticates (except for zero amount).

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Excellent.

PR #1212, opened 2023-10-02
chore(coinbase): transition to v3 Buy API

Syn-McJ:
Coinbase v2 Buy APIs are being shutdown, so we need to switch to v3. 

- Move the Coinbase feature out of the separate activity into the `nav_home` graph.
- Rename packages to align with other integrations.
- Replace `v2/accounts` with `api/v3/brokerage/accounts` calls.
- Replace `v2/accounts/{account_id}/buys` with `api/v3/brokerage/orders` calls.
- Rearrange Buy flow: no need to create and commit an order as the `/orders` endpoint executes the order immediately.
- Other refactoring to improve some unneeded/convoluted code.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
Excellent work.

PR #1217, opened 2023-10-10
feat(coinjoin): pick privacy level

Syn-McJ:
The user must be able to pick privacy level for CoinJoin mixing.

- Repurpose the `CreateUsernamePrivacyFragment` for the new flow.
- Adjust the UI and logic to match the latest design.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/service/CoinJoinService.kt
```diff
@@ -66,9 +66,10 @@ enum class CoinJoinMode {
 }
 
 interface CoinJoinService {
+    var mode: CoinJoinMode
+    val mixingStatus: MixingStatus
+
     fun needsToMix(amount: Coin): Boolean
-    fun getMode(): CoinJoinMode
-    fun setMode(mode: CoinJoinMode)
```
Syn-McJ, line 71: I think in Kotlin we don't need dedicated setter/getter methods

wallet/src/de/schildbach/wallet/ui/coinjoin/CoinJoinLevelViewModel.kt
```diff
@@ -0,0 +1,69 @@
...
+            AnalyticsConstants.CoinJoinPrivacy.USERNAME_PRIVACY_BTN_CONTINUE,
+            mapOf(AnalyticsConstants.Parameter.VALUE to mode.name)
+        )
+
+        coinJoinService.mode = mode
+        coinJoinService.prepareAndStartMixing() // TODO: change the logic if needed
+    }
+
+    suspend fun stopMixing() {
+//        coinJoinService.stopMixing() // TODO expose stop method?
```
Syn-McJ, line 63: Stopping the mixing isn't exposed in the `coinJoinService`, don't know if that's what we want. I'll leave this to you @HashEngineering

wallet/src/de/schildbach/wallet/service/CoinJoinService.kt
HashEngineering: Yeah, this was probably due to thinking in Java.

wallet/src/de/schildbach/wallet/ui/coinjoin/CoinJoinLevelViewModel.kt
HashEngineering: Yes, we will have to change how we tell **CoinJoinService** what to do.  With this PR we are adding a means for the user to stop the process.  There will be other events that will activate or deactivate the CoinJoinService was well.

HashEngineering (approved):

PR #1219, opened 2023-10-16
fix(dashpay): only set tx icon in updateIcon method 

Syn-McJ:
`TransactionResultViewBinder` has some extra code related to setting the icon that should be removed.

- Remove all icon-setting code except for updateIcon

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


HashEngineering (approved):
LGTM

PR #1240, opened 2023-12-16
fix: eliminate race condition when updating BlockchainState

HashEngineering:
## Screenshots / Videos

Resolve the issue where "Syncing..." appears for a long time:

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt
+++ b/wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt
@@ -23,7 +23,7 @@ import dagger.hilt.android.qualifiers.ApplicationContext
 import de.schildbach.wallet.Constants
 import de.schildbach.wallet.database.dao.BlockchainStateDao
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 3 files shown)


Syn-McJ (approved):

wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt
```diff
@@ -173,7 +166,11 @@ class BlockchainStateDataProvider @Inject constructor(
                 }
 
                 val validMNsCount = if (mnlist.size() != 0) {
-                    mnlist.validMNsCount
+                    var virtualMNCount = 0
+                    mnlist.forEachMN(true) { entry ->
+                        virtualMNCount += if (entry.isHPMN) 4 else 1
+                    }
+                    virtualMNCount
```
HashEngineering, line 173: EvoNodes have 4x the collateral (4000 instead of 1000) and receive 4x times the block rewards, so we will count them as 4 regular masternodes.  

Each block only pays out a single masternode or 1/4 of a EvoNode.  EvoNodes receive 4 blocks of rewards in a row.  APY calculations are based on a single block.

Without this fix, we are using the number of actual masternode servers which will inflate the APY by around 10%.

Future improvements:
1.  We could base the APY on a number of blocks in recent history such as the past 100.
2.  Masternode reward calculations could be moved into dashj, but not sure which class would get the functions (Block or Masternode List)

Syn-McJ (approved):
Looks good

PR #1247, opened 2024-01-12
feat(dashpay): add coinjoin error handling to enter amount screen

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.kt
```diff
@@ -175,7 +175,13 @@ class SendCoinsFragment: Fragment(R.layout.send_coins_fragment) {
             errorMessage = when (dryRunException) {
                 is Wallet.DustySendRequested -> getString(R.string.send_coins_error_dusty_send)
                 is InsufficientMoneyException -> if (!requirePinForBalance || userAuthorizedDuring) {
-                    getString(R.string.send_coins_error_insufficient_money)
+                    getString(
+                        if (viewModel.isCoinJoinInsufficentMoneyException) {
```
Syn-McJ, line 179: Looks like a typo in "Insufficient".

Also, a call to viewModel property here looks a bit weird since those are supposed to reflect the state of the view, but this property will return the same value whenever CoinJoin is on, even if there is no error.

I would improve this by moving it to the ViewModel where `dryRunException` is assigned and using a different exception type for CoinJoin. Something like `InsufficientCoinJoinMoneyException` that derives from `InsufficientMoneyException` similar to `LeftoverBalanceException`.

Syn-McJ (approved):
Looks good, just a small improvement suggestion.

wallet/src/de/schildbach/wallet/ui/send/SendCoinsFragment.kt
HashEngineering: Thanks.  I will try this.

HashEngineering: @Syn-McJ see #[6d5b983](https://github.com/dashpay/dash-wallet/pull/1247/commits/6d5b9837300fee790f49a5abe796177389b12496)

Syn-McJ (approved):
Looks good

PR #1251, opened 2024-01-26
Feature maya integration address

HashEngineering:

Support getting deposit addresses for other exchanges to use in Maya

How it is done:

`ExchangeIntegrationProvider` is injected into Maya to obtain a list of exchanges that is generated by the implementation in `wallet`.

* Coinbase had all of the required API's.  All that was added was a means to cache the accounts and their addresses.
* Uphold support required a new API.  A few other API's were implemented as `suspend` functions.  These new methods will also save the accounts and addresses
* ExchangeConfig is a class that will manage saving and retrieving accounts and addresses.  Currently it saves a map of currency to account number as one map and account number to address in the second map.  Perhaps it would be easier to simply have a single map to associate a currency with an address.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/WalletApplication.java
```diff
@@ -393,7 +396,7 @@ private void initUphold() {
         UpholdConstants.CLIENT_ID = BuildConfig.UPHOLD_CLIENT_ID;
         UpholdConstants.CLIENT_SECRET = BuildConfig.UPHOLD_CLIENT_SECRET;
         UpholdConstants.INSTANCE.initialize(Constants.NETWORK_PARAMETERS.getId().contains("test"));
-        UpholdClient.init(getApplicationContext(), authenticationHash);
+        UpholdClient.init(getApplicationContext(), authenticationHash, upholdConfig);
```
HashEngineering, line 399: I didn't like this idea, but seems necessary due to legacy code.

integrations/uphold/src/main/java/org/dash/wallet/integrations/uphold/data/UpholdCardAddressList.kt
```diff
@@ -0,0 +1,17 @@
...
+    val type: String
+) {
+    fun isEmpty() = addresses.isEmpty()
+    fun firstOrNull() = addresses.firstOrNull()
+}
+
+data class UpholdCardAddress2(
+    val format: String,
+    val value: String
+)
```
HashEngineering, line 17: This needs a different name.  This is a different structure compared to the other `UpholdCardAddress`.

wallet/src/de/schildbach/wallet/service/ExchangeIntegrationListProvider.kt
```diff
@@ -0,0 +1,186 @@
...
+                            "coinbase",
+                            false,
+                            null,
+                            currency,
+                            R.string.coinbase,
+                            R.drawable.ic_coinbase
+                        )
+                    )
+                }
+            }
```
HashEngineering, line 95: The idea is to look in the cache (the config for that exchange), then query accounts and then an address.  The exchange repository or client will save any results to the cache.

common/src/main/java/org/dash/wallet/common/ui/address_input/AddressInputFragment.kt
```diff
@@ -210,7 +210,6 @@ abstract class AddressInputFragment : Fragment(R.layout.fragment_address_input)
                     )
                 }
             )
-            notifyItemRangeInserted(0, source.size)
```
HashEngineering: this was causing a crash when the list was refeshed.  With the adapter that is used, we don't need to notify about updates.

integrations/uphold/src/main/java/org/dash/wallet/integrations/uphold/ui/UpholdPortalFragment.kt
```diff
@@ -70,13 +73,22 @@ class UpholdPortalFragment : Fragment(R.layout.fragment_integration_portal) {
                 }
             }
 
+            // TODO: this does not work, how can we tell this activity to close itself
+            if (action == LOGIN_AND_CLOSE) {
+                intent.extras?.putString(INTENT_ACTION, action)
+            }
+            println("auth: $intent")
+            println("auth: extras ${intent.extras}")
```
HashEngineering, line 81: this will be removed.

integrations/uphold/src/main/java/org/dash/wallet/integrations/uphold/ui/UpholdPortalFragment.kt
```diff
@@ -154,7 +166,13 @@ class UpholdPortalFragment : Fragment(R.layout.fragment_integration_portal) {
             lifecycleScope.launch {
                 viewModel.refreshBalance()
                 viewModel.checkCapabilities()
+                if (action == LOGIN_AND_CLOSE) {
+                    findNavController().popBackStack()
+                }
```
HashEngineering, line 171: this does nothing, but this is not within an observer.

common/src/main/java/org/dash/wallet/common/integrations/Integration.kt
```diff
@@ -0,0 +1,40 @@
...
+    @StringRes
+    val name: Int,
+    @DrawableRes
+    val iconId: Int
+)
+
+interface ExchangeIntegrationProvider {
+
+    suspend fun getDepositAddresses(currency: String): List<ExchangeIntegration>
+    // fun observeDepositAddresses(currency: String) : Flow<List<ExchangeIntegration>>
```
HashEngineering, line 38: For now this is a get function and not observe.

If we make an observer, then we need to monitor for Login/Logout of the exchange to refresh the deposit address list.

the ViewModel or Fragment of the Exchange Portal Fragment needs to initiate a "refresh" after observing a change in Login Status.

integrations/uphold/src/main/java/org/dash/wallet/integrations/uphold/data/UpholdAddress.kt
```diff
@@ -14,28 +14,9 @@
...
-
-    public void setLine2(String line2) {
-        this.line2 = line2;
-    }
-
-}
+data class UpholdAddress(
+    val format: String,
+    val value: String
+)
```
HashEngineering, line 22: This class was never used previously for its contents.  The original code didn't know what line one and two were.  Now we know what they are and will use them.

Syn-McJ (approved):
Looks good. Some commented code left

PR #1253, opened 2024-02-02
feat: use DashSpend service as a replacement for DashDirect

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/data/entity/GiftCard.kt
```diff
@@ -21,10 +21,11 @@ import androidx.room.Entity
 import androidx.room.PrimaryKey
 import com.google.zxing.BarcodeFormat
 import org.bitcoinj.core.Sha256Hash
-
+// TODO: need migration for 12->13 for addition of id column?
 @Entity(tableName = "gift_cards")
 data class GiftCard(
     @PrimaryKey var txId: Sha256Hash,
+    var id: String? = "",
```
Syn-McJ, line 28: Because the gift card table is in production - yes, we will need a migration.
(This TODO might not be needed as it looks like the migration is added already).

features/exploredash/src/main/res/values/strings-explore-dash.xml
```diff
@@ -164,64 +164,70 @@
 
     <!-- purchase_gift_card -->
     <!-- TODO: marked as non-translatable until gift cards are re-integrated -->
```
Syn-McJ, line 166: TODO can be removed

wallet/src/de/schildbach/wallet/ui/payments/PaymentsReceiveFragment.kt
```diff
@@ -57,22 +55,18 @@ class PaymentsReceiveFragment : Fragment(R.layout.fragment_payments_receive) {
...
 
         binding.receiveInfo.setOnSpecifyAmountClicked {
-            analytics.logEvent(AnalyticsConstants.SendReceive.SPECIFY_AMOUNT, mapOf())
             findNavController().navigate(PaymentsFragmentDirections.paymentsToReceive())
         }
         binding.receiveInfo.setOnAddressClicked {
-            analytics.logEvent(AnalyticsConstants.SendReceive.COPY_ADDRESS, mapOf())
         }
         binding.receiveInfo.setOnShareClicked {
-            analytics.logEvent(AnalyticsConstants.SendReceive.SHARE, mapOf())
```
Syn-McJ, line 75: Don't we need these?

Syn-McJ (approved):
Looks good. There is some alanytics removed - please recheck if it's correct.

wallet/src/de/schildbach/wallet/ui/payments/PaymentsReceiveFragment.kt
HashEngineering: Yes, we do.  They were added back.

PR #1263, opened 2024-02-29
fix: remove memoryleaks in Coinbase and Explore

HashEngineering:

Remove memory leaks related never unsubscribing from:
1. `setOnApplyWindowInsetsListener`
2. `onBackPressedDispatcher.addCallback`

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

### DIFF
--- a/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/explore/SearchFragment.kt
+++ b/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/explore/SearchFragment.kt
@@ -26,6 +26,7 @@ import android.os.Looper
 import android.view.View
 import android.view.inputmethod.InputMethodManager
 import androidx.activity.OnBackPressedCallback
+import androidx.activity.addCallback
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.core.animation.doOnEnd
 import androidx.core.content.ContextCompat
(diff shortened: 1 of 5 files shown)


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/explore/SearchFragment.kt
```diff
@@ -554,19 +563,12 @@ class SearchFragment : Fragment(R.layout.fragment_search) {
...
-                        }
-                    }
-                }
-            )
+        onBackPressedCallback = requireActivity().onBackPressedDispatcher.addCallback(viewLifecycleOwner) {
+            hardBackAction.invoke()
+            if (isMerchantTopic) {
+                viewModel.logEvent(AnalyticsConstants.Explore.MERCHANT_DETAILS_BACK_BOTTOM)
+            }
+        }
```
HashEngineering, line 571: code was simplified to be more kotlin than Java.  Docs were chanced to ensure that the parameter set to to`true` would be `true` in kotlin which uses a default value for the parameter.

I will need to make these changes separately to the Maya integration.

Syn-McJ (approved):
Looks good. Profiler reports no leaks related to `OnBackPressedCallback`.

There is, however, still a leak in the `CoinbaseBuyDashFragment`. Looks like it's happening because we're attaching `EnterAmountViewModel` to the `MainActivity` with `by activityViewModels()` and it remains in memory along with the `EnterAmountFragment` after exiting the Coinbase module. 

We don't have this issue in CrowdNode because it lives in a separate activity, but will probably get it in Maya.

This can be solved by scoping the `EnterAmountViewModel` to the fragment in which `EnterAmountFragment` is hosted in. `EnterAmountFragment` would need to accept the `viewModelStore` owner and get the `EnterAmountViewModel` using the provided owner. 

We can skip it for now and perhaps come back to it in one of the Maya stories.

HashEngineering:
I will see about removing the `EnterAmount*` related memory leaks soon.

PR #1267, opened 2024-03-08
fix: prevent NPE after corrupt wallet requires backup restoration

HashEngineering:

This fixes a crash that one user had 900+ times.  

If the wallet file becomes corrupted, the backup wallet file is restored, but during that process `wallet` is null and will result in an NPE when the blockchain is reset.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


Syn-McJ (approved):

Setarland (commented):
>

PR #1273, opened 2024-03-22
chore(coinbase): upgrade to payment methods v3

Syn-McJ:
payment-methods v2 is being shut down.

- Upgraded to payment_methods v3.
- Related changes.
- Disabled auth limits check since `auth_limits` is null
- Removed unneeded `Build.VERSION_CODES.N` checks 

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

- [] I have added or updated relevant unit/integration/functional/e2e tests

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
+++ b/common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
@@ -17,10 +17,12 @@
 
 package org.dash.wallet.common.util
 
+import android.os.Build
 import android.os.LocaleList
 import java.math.RoundingMode
 import java.text.NumberFormat
-import java.util.*
+import java.util.Currency
+import java.util.Locale
 
 /**
  * @author Andreas Schildbach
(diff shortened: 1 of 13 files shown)


common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
```diff
@@ -45,7 +54,7 @@ object GenericUtils {
     }
 
     fun getDeviceLocale(): Locale {
-        val countryCode = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+        val countryCode = if (!isRunningUnitTest) {
             LocaleList.getDefault()[0].country
         } else {
             Locale.getDefault().country
```
Syn-McJ, line 60: While we don't need the `Build.VERSION_CODES.N` check anymore, the unit test fails on `LocaleList.getDefault()[0]`, and mocking the static method doesn't seem to work. This workaround is somewhat ugly, but will do for now.

HashEngineering (commented):
LGTM

PR #1280, opened 2024-04-18
feat: add jetpack compose 

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


build.gradle
```diff
@@ -1,22 +1,22 @@
...
         ok_http_version = '4.9.1'
         dashjVersion = '20.0.4'
-        hiltVersion = '2.45'
+        hiltVersion = '2.51'
+        hiltCompilerVersion = '1.2.0'
         hiltWorkVersion = '1.0.0'
         workRuntimeVersion='2.7.1'
         firebaseVersion = '32.1.1'
-        roomVersion = '2.4.3'
+        roomVersion = '2.6.1'
```
HashEngineering, line 12: This was changed in hopes it might eliminate this:
``` 
e: [ksp] [MissingType]: Element 'org.dash.wallet.common.data.RoomConverters' references a type that is not present
e: [ksp] androidx.room.RoomKspProcessor was unable to process 'org.dash.wallet.features.exploredash.ExploreDatabase' because not all of its dependencies could be resolved. Check for compilation errors or a circular dependency with generated code.
e: Error occurred in KSP, check log for detail
```

Similar error: https://issuetracker.google.com/issues/274889425

common/src/main/java/org/dash/wallet/common/ui/components/MenuItem.kt
```diff
@@ -0,0 +1,195 @@
...
+    title: String,
+    subtitle: String? = null,
+    details: String? = null,
+    icon: Int? = null, // Assuming you use resource IDs for icons
+    showInfo: Boolean = false,
+    showChevron: Boolean = false,
+    isToggled: (() -> Boolean)? = null, // Change to a lambda that returns a Boolean
+    onToggleChanged: ((Boolean) -> Unit)? = null,
+    action: (() -> Unit)? = null
+) {
```
HashEngineering, line 99: from ChatGPT based on iOS MenuItem

common/src/main/java/org/dash/wallet/common/ui/components/MenuItem.kt
```diff
@@ -0,0 +1,195 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import org.dash.wallet.common.R
+
+//@Composable
+//fun MenuItemIcon(icon: Int?) {
```
HashEngineering, line 43: I was trying to fix a problem that was resolved by some build.gradle changes, so we don't need this commented code.

features/exploredash/build.gradle
```diff
@@ -10,9 +10,9 @@ plugins {
 
 android {
     defaultConfig {
-        compileSdk 33
+        compileSdk 34
         minSdkVersion 24
-        targetSdkVersion 33
+        targetSdkVersion 34
```
HashEngineering, line 15: Android SDK 34 was required for many of compose libs.

common/src/main/java/org/dash/wallet/common/ui/components/MenuItem.kt
```diff
@@ -0,0 +1,195 @@
...
+@Composable
+fun MenuItem(
+    title: String,
+    subtitle: String? = null,
+    details: String? = null,
+    icon: Int? = null, // Assuming you use resource IDs for icons
+    showInfo: Boolean = false,
+    showChevron: Boolean = false,
+    isToggled: (() -> Boolean)? = null, // Change to a lambda that returns a Boolean
+    onToggleChanged: ((Boolean) -> Unit)? = null,
```
Syn-McJ, line 97: ChatGPT comments can be removed.

I think it came up with something weird here and below with `toggleState`. 
`isToggled` should not be lambda if we already have `onToggleChanged` for reporting state change. It should just be `isToggled: Boolean = false,`.

The reason the binding in iOS code is nullable is to indicate if we need to show the toggle in the first place. If we don't need the toggle, binding is null. If we do and it's not null, we bind it to the caller's state.

In JC, we can either use `MutableState` or split it in two like it's sort of done here, but indicate visibility with the `onToggleChanged` callback. If it's nullable, don't show the toggle. Like this:

 ```
onToggleChanged?.let {
    Switch(
        checked = isToggled,
        onCheckedChange = { newState ->
            onToggleChanged?.invoke(newState)
        },
        modifier = Modifier.size(60.dp)
    )
}
```

**Review by Syn-McJ — CHANGES_REQUESTED** [review]
ChatGPT generated code can use some fixes around toggle, but otherwise looks good

File: wallet/src/de/schildbach/wallet/ui/more/ToolsFragment.kt [hunk 6]
```diff
@@ -191,3 +212,11 @@ class ToolsFragment : Fragment(R.layout.fragment_tools) {
...
     }
 }
+
+@Composable
+fun Greeting(text: String) {
+    Text(
+        text = "$text!", color = Color(0xFF0F9D58),
+        fontSize = 30.sp
+    )
+}
```
Syn-McJ, line 222: Not needed

common/src/main/java/org/dash/wallet/common/ui/components/MenuItem.kt
HashEngineering: thanks

wallet/src/de/schildbach/wallet/livedata/EncryptWalletLiveData.kt
```diff
@@ -74,10 +74,12 @@ class EncryptWalletLiveData(
     @SuppressLint("StaticFieldLeak")
     internal inner class EncryptWalletTask : AsyncTask<Any, Void, Resource<Wallet>>() {
 
+        @Deprecated("Deprecated in Java")
         override fun onPreExecute() {
```
HashEngineering, line 78: these were added after running the 1.8 to 1.9 migration tool.

wallet/src/de/schildbach/wallet/ui/WalletUriHandlerActivity.kt
```diff
@@ -49,11 +53,23 @@ class WalletUriHandlerActivity : AppCompatActivity() {
...
+            var resultIntent: Intent? = null
+            if (result.resultCode == Activity.RESULT_OK) {
+                val data = result.data
+                val requestData = intent.data
+                val transactionHash = BitcoinIntegration.transactionHashFromResult(data)
+                resultIntent = WalletUri.createPaymentResult(requestData, transactionHash)
+            }
+            setResult(result.resultCode, resultIntent)
+            finish()
+        }
```
HashEngineering, line 71: replaces deprecated onActiivtyResult

HashEngineering: This is not used internally, but when apps externally request data using `dashwallet://...`

Syn-McJ (approved):
Looks good

PR #1284, opened 2024-05-01
feat: warn user of stale exchange rates

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/data/ExchangeRatesConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+            PREFERENCES_NAME,
+            walletDataProvider
+        ) {
+        companion object {
+            private const val PREFERENCES_NAME = "exchange_rates_config"
+            val EXCHANGE_RATES_RETRIEVAL_TIME = longPreferencesKey("exchange_rates_retrieval_time")
+            val EXCHANGE_RATES_RETRIEVAL_FAILURE = booleanPreferencesKey("exchange_rates_retrieval_error")
+            val EXCHANGE_RATES_PREVIOUS_RETRIEVAL_TIME = longPreferencesKey("exchange_rates_previous_retrieval_time")
+        }
+    }
```
HashEngineering, line 25: At first I was going to put these in WalletUIConfig, but that would have created a circular reference.

common/src/main/java/org/dash/wallet/common/data/entity/ExchangeRate.kt
```diff
@@ -34,16 +34,25 @@ open class ExchangeRate : Parcelable {
...
     @Ignore
     private var currencyName: String? = null
 
     @Ignore
     private var currency: Currency? = null
 
+    constructor(currencyCode: String, rate: String?, retrievalTime: Long) {
+        this.currencyCode = currencyCode
+        this.rate = rate
+        this.retrievalTime = retrievalTime
```
HashEngineering, line 49: CTX does return a field that contains the time their servers last retrieved this particular rate.

It is possible for CTX to return rates, but they are stale.

common/src/main/java/org/dash/wallet/common/data/entity/ExchangeRate.kt
```diff
@@ -79,13 +88,16 @@ open class ExchangeRate : Parcelable {
...
+            currencyName =
+                if (currencyCode.length == 3) {
+                    try {
+                        getCurrency().displayName
+                    } catch (x: IllegalArgumentException) {
+                        currencyCode
+                    }
+                } else {
                     currencyCode
                 }
```
HashEngineering, line 100: Here, I am sure that the ktlint plugin on automatically formatted this file upon save.  I have turned off this feature.

common/src/main/java/org/dash/wallet/common/ui/components/MyTheme.kt
```diff
@@ -0,0 +1,48 @@
...
+object MyTheme {
+    val ToastBackground = Color(0xff191c1f).copy(alpha = 0.9f)
+
+    val Body2Regular = TextStyle(
+        fontSize = 14.sp,
+        lineHeight = 20.sp,
+        fontFamily = FontFamily.Default, // FontFamily(Font(R.font.inter)) // crashes,
+        fontWeight = FontWeight(400),
+        color = Color.White
+    )
```
HashEngineering, line 38: This is a start for a "stylesheet" file, there probably is a better way to do this, perhaps a standard way that it is done with Compose.

wallet/src/de/schildbach/wallet/rates/ExchangeRatesRepository.kt
```diff
@@ -71,10 +87,16 @@ class ExchangeRatesRepository @Inject constructor(
                 }
 
                 if (rates.isNotEmpty()) {
+                    previousRates.clear()
+                    previousRates.addAll(exchangeRatesDao.getAll())
```
HashEngineering, line 91: here we save the previous rates in memory to compare with the new rates.

common/src/main/java/org/dash/wallet/common/data/ExchangeRatesConfig.kt
Syn-McJ: `WalletUIConfig` might be better to leave for UI settings that can be changed by the user, or closely related preferences.

wallet/src/de/schildbach/wallet/ui/main/WalletActivityExt.kt
```diff
@@ -179,13 +199,14 @@ object WalletActivityExt {
...
-            }
-        )
+        val storageManagerIntent =
+            Intent(
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    StorageManager.ACTION_MANAGE_STORAGE
+                } else {
+                    Settings.ACTION_MANAGE_APPLICATIONS_SETTINGS
+                }
+            )
```
HashEngineering, line 209: Is there a way to get the `ktlint` plugin for Android Studio to use our rules in this project?


This was another case where I was using the new ktlint plugin to reformat a file on save, which would actually change a lot in `wallet` and in other modules may apply rules that we don't use.

common/src/main/java/org/dash/wallet/common/ui/components/MyTheme.kt
Syn-McJ: Usually in JC you would set a MaterialTheme with a `typography` field, however, the names of standard text styles aren't going to match with our typography sheet in Figma, so it's better to have named styles as you did here.

wallet/src/de/schildbach/wallet/ui/main/WalletActivityExt.kt
Syn-McJ: I think if you set `.editorconfig` as the baseline file in ktlint plugin settings, it should pickup our rules.
I removed ktlint plugin from the studio because after some update it broke and started to complain about everything.

Syn-McJ (approved):
Looks good

PR #1285, opened 2024-05-16
fix: syncing 31

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/build.gradle
```diff
@@ -220,6 +221,15 @@ android {
...
 
+    signingConfigs {
+        debug {
+            storeFile file('keystore/debug.keystore')
+            storePassword "android"
+            keyAlias "androiddebugkey"
+            keyPassword "android"
+        }
+    }
+
```
Syn-McJ, line 232: Didn't we remove the leaked debug key recently?

Syn-McJ (approved):
Looks good, apart from the debug key, please check the comment

wallet/build.gradle
HashEngineering: Yes, thanks for catching this.  I needed to add this back for working with an install that was built using the old debug key.  It will be removed before merging.

PR #1309, opened 2024-09-28
feat(dashpay): add mix dash first warning dialog

HashEngineering:
## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


wallet/src/de/schildbach/wallet/ui/more/MoreFragment.kt
```diff
@@ -172,7 +176,16 @@ class MoreFragment : Fragment(R.layout.fragment_more) {
         binding.errorUpdatingProfile.cancel.setOnClickListener { dismissProfileError() }
         binding.editUpdateSwitcher.isVisible = false
         binding.joinDashpayBtn.setOnClickListener {
-            startActivity(Intent(requireContext(), CreateUsernameActivity::class.java))
+            lifecycleScope.launch {
+                val shouldShowMixDashDialog = withContext(Dispatchers.IO) { createIdentityViewModel.shouldShowMixDash() }
+                //withContext(Dispatchers.Main) {
```
Syn-McJ, line 181: I guess this is commented because it wasn't clear if needed or not. `lifecycleScope` will execute in the main thread and `shouldShowMixDash()` is dispatched in IO with the result being returned back to the main thread, so `Dispatchers.Main` context indeed isn't needed.

`viewModelScope` by default also runs in the main thread.

Syn-McJ (approved):
Looks good

wallet/src/de/schildbach/wallet/ui/more/MoreFragment.kt
HashEngineering: I will remove the `withContext(Dispatchers.Main)` comment

PR #1318, opened 2024-10-29
fix(dashpay): username request voting period bugs

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

### DIFF
--- a/wallet/src/de/schildbach/wallet/ui/username/voting/RequestUserNameViewModel.kt
+++ b/wallet/src/de/schildbach/wallet/ui/username/voting/RequestUserNameViewModel.kt
@@ -35,6 +35,7 @@ import de.schildbach.wallet.service.CoinJoinMode
 import de.schildbach.wallet.ui.dashpay.CreateIdentityService
 import de.schildbach.wallet.ui.dashpay.PlatformRepo
 import de.schildbach.wallet.ui.dashpay.work.BroadcastIdentityVerifyOperation
+import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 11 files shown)


wallet/src/de/schildbach/wallet/ui/username/voting/RequestUserNameViewModel.kt
```diff
@@ -247,7 +253,7 @@ class RequestUserNameViewModel @Inject constructor(
     }
 
     fun checkUsername(requestedUserName: String?) {
-        viewModelScope.launch {
+        viewModelScope.launch(Dispatchers.IO) {
```
Syn-McJ, line 256: In general, it's better not to dispatch coroutines in IO scope at the top level, but go down to where the actual work is done and call it inside `withContext(Dispatchers.IO)` so that the top-level callers don't have to worry about threading, correct context and switching back to the main thread to update UI.

Syn-McJ (approved):
Looks good. Some coroutine improvements can be made

wallet/src/de/schildbach/wallet/ui/username/voting/RequestUserNameViewModel.kt
HashEngineering: thanks for this idea.

HashEngineering: I will remove this `(Dispatchers.IO)` and use `withContext` in the next PR

PR #1393, opened 2025-05-15
Prepare 11.2.2

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added new and updated localized strings for CoinJoin transaction statuses, block explorer selection, copyright notices, and account management in multiple languages.
  - Expanded UI text for gift card options, filtering, and status messages in several localizations.

- **Bug Fixes**
  - Improved localization by correcting or updating the translation for the "Discount" sort option in various languages.
  - Added exception handling to prevent crashes during time skew updates on system time changes.

- **Chores**
  - Updated dependency versions and Explore database file paths to new versions.
  - Cleaned up resource files by removing unnecessary blank lines.

- **Refactor**
  - Enhanced fiat amount formatting for gift card purchases.
  - Adjusted dust threshold logic and improved coroutine support for asset lock transactions.
  - Streamlined initialization flow in staking activity.
  - Updated coroutine scope usage in payment receive fragment.
  - Set explicit min and max amounts in receive fragment input.

### DIFF
--- a/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ctxspend/CTXSpendViewModel.kt
+++ b/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ctxspend/CTXSpendViewModel.kt
@@ -126,7 +126,7 @@ class CTXSpendViewModel @Inject constructor(
 
             val response = repository.purchaseGiftCard(
                 merchantId = it,
-                fiatAmount = amountValue.toBigDecimal().toDouble().toString(),
+                fiatAmount = MonetaryFormat.FIAT.noCode().format(amountValue).toString(),
                 fiatCurrency = "USD",
                 cryptoCurrency = Constants.DASH_CURRENCY
             )
--- a/features/exploredash/src/main/res/values-ar/strings-explore-dash.xml
+++ b/features/exploredash/src/main/res/values-ar/strings-explore-dash.xml
@@ -123,6 +123,7 @@
     <string-array name="sort_by_options_names">
         <item>اسم</item>
         <item>مسافة</item>
+        <item>Discount</item>
     </string-array>
 
     <plurals name="radius_miles">
--- a/features/exploredash/src/main/res/values-ar/strings-explore-dash.xml
+++ b/features/exploredash/src/main/res/values-ar/strings-explore-dash.xml
@@ -93,7 +93,6 @@
     <string name="explore_filtered_by">تمت تصفيتها بواسطة:</string>
     <string name="explore_pay_gift_card">ادفع ببطاقة الهدايا</string>
     <string name="explore_sort_by">ترتيب حسب</string>
-
     <string-array name="merchants_filter_options">
         <item>عبر الانترنت</item>
         <item>قريب</item>
@@ -124,7 +123,6 @@
     <string-array name="sort_by_options_names">
         <item>اسم</item>
         <item>مسافة</item>
-        <item>Discount</item>
     </string-array>
 
     <plurals name="radius_miles">
--- a/wallet/src/de/schildbach/wallet/service/platform/TopUpRepository.kt
+++ b/wallet/src/de/schildbach/wallet/service/platform/TopUpRepository.kt
@@ -91,6 +91,7 @@ class TopUpRepositoryImpl @Inject constructor(
 ) : TopUpRepository {
     companion object {
         private val log = LoggerFactory.getLogger(TopUpRepositoryImpl::class.java)
+        private const val MIN_DUST_FACTOR = 10L
     }
 
     private val platform = platformRepo.platform
@@ -109,7 +110,7 @@ class TopUpRepositoryImpl @Inject constructor(
             Constants.DASH_PAY_FEE
         }
         val balance = walletDataProvider.wallet!!.getBalance(Wallet.BalanceType.ESTIMATED_SPENDABLE)
-        val emptyWallet = balance == fee && balance <= (fee + Transaction.MIN_NONDUST_OUTPUT)
+        val emptyWallet = balance == fee && balance <= (fee + Transaction.MIN_NONDUST_OUTPUT.multiply(MIN_DUST_FACTOR))
         val cftx = blockchainIdentity.createAssetLockTransaction(
             fee,
             keyParameter,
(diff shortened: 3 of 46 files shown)


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/ctxspend/CTXSpendViewModel.kt
```diff
@@ -126,7 +126,7 @@ class CTXSpendViewModel @Inject constructor(
 
             val response = repository.purchaseGiftCard(
                 merchantId = it,
-                fiatAmount = amountValue.toBigDecimal().toDouble().toString(),
+                fiatAmount = MonetaryFormat.FIAT.noCode().format(amountValue).toString(),
```
HashEngineering, line 129: If we had a fiat value of 0.00999981 that should be 0.01

`amountValue.toBigDecimal().toDouble().toString()` would give 0.00999981.  CTX would truncate that to 0.00

`MonetaryFormat.FIAT.noCode().format(amountValue).toString()` will give 0.01

features/exploredash/src/main/res/values-ar/strings-explore-dash.xml
```diff
@@ -124,7 +123,6 @@
     <string-array name="sort_by_options_names">
         <item>اسم</item>
         <item>مسافة</item>
-        <item>Discount</item>
```
HashEngineering, line 127: whoops, `tx` overwrote the these changes.

wallet/src/de/schildbach/wallet/ui/staking/StakingActivity.kt
```diff
@@ -60,10 +60,12 @@ class StakingActivity : LockScreenActivity() {
...
 
         binding = ActivityStakingBinding.inflate(layoutInflater)
-        lifecycleScope.launch { navController = setNavigationGraph() }
+        lifecycleScope.launch {
+            navController = setNavigationGraph()
+            viewModel.observeOnlineAccountStatus().observe(this@StakingActivity, ::handleOnlineAccountStatus)
+        }
 
         viewModel.navigationCallback.observe(this, ::handleNavigationRequest)
-        viewModel.observeOnlineAccountStatus().observe(this, ::handleOnlineAccountStatus)
```
HashEngineering, line 66: there was a race condition.  if the observer ran before the `navController` was initialized, then it would crash.  This was a rare crash.

wallet/src/de/schildbach/wallet/ui/payments/ReceiveFragment.kt
```diff
@@ -64,7 +65,8 @@ class ReceiveFragment : Fragment(R.layout.fragment_receive) {
                 }
             )
         }
-
+        enterAmountViewModel.setMaxAmount(Coin.ZERO)
+        enterAmountViewModel.setMinAmount(Coin.ZERO)
```
HashEngineering, line 69: If the user goes to the Gift Card purchase flow, these limits will be set to match min/max of the card and the wallet balance and will limit what can be entered in the Receive screen.  Therefore we need to reset the limits to zero which allows any number to be entered.

Not sure if there is a different way to clear the values with enterAmountViewModel by attaching it to a particular navigation graph.  It needs to have a scope larger than a single fragment.  With activityViewModels it stays alive while MainActivity is alive.

wallet/src/de/schildbach/wallet/service/platform/TopUpRepository.kt
```diff
@@ -91,25 +92,27 @@ class TopUpRepositoryImpl @Inject constructor(
...
             Constants.DASH_PAY_FEE_CONTESTED
         } else {
             Constants.DASH_PAY_FEE
         }
-        val balance = walletDataProvider.wallet!!.getBalance(Wallet.BalanceType.ESTIMATED_SPENDABLE)
-        val emptyWallet = balance == fee && balance <= (fee + Transaction.MIN_NONDUST_OUTPUT)
+        val balance = walletDataProvider.observeSpendableBalance().first()
+        val emptyWallet = balance == fee ||
+                (balance >= fee && balance <= (fee + Transaction.MIN_NONDUST_OUTPUT.multiply(MIN_DUST_FACTOR)))
+        Context.propagate(walletDataProvider.wallet!!.context)
```
HashEngineering, line 115: If a user has about 0.032 DASH and then mixes them to get just above 0.03 DASH, this code wouldn't set emptyWallet to `true`, then the username creation process would get stuck and report an insufficient funds exception in the logs.  the only way to recover was to deposit more DASH and mix a bit more so that the fee could be covered.

Our goal is to empty the wallet if the username creation fee is just a bit higher than the requirement (0.03 in this case) to avoid such errors.

In this case, we were looking at the wrong balance (total) instead of spendable.

There still might be a way for this to fail.

wallet/src/de/schildbach/wallet/ui/payments/PaymentsReceiveFragment.kt
```diff
@@ -98,7 +98,7 @@ class PaymentsReceiveFragment : Fragment(R.layout.fragment_payments_receive) {
             binding.receiveInfo.setProfile(it?.username, it?.displayName, it?.avatarUrl, it?.avatarHash)
         }
 
-        lifecycleScope.launch {
+        viewLifecycleOwner.lifecycleScope.launch {
```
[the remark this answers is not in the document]
Syn-McJ: Right, `lifecycleScope` is tied to Fragments own lifecycle while `viewLifecycleOwner.lifecycleScope` to the view's lifecycle which can be destroyed while the fragment is still fine.

We probably have similar issues in other fragments where I'm mixing these two concepts, in some cases it's fixed with keeping a local `binding` variable, but using `viewLifecycleOwner.lifecycleScope` for coroutines that modify UI is probably a better solution.

wallet/src/de/schildbach/wallet/ui/payments/ReceiveFragment.kt
Syn-McJ: I remember thinking on this problem in the context of leaking EnterAmountFragment. We might need a separate story if we want to fix this at some point.

Moving EnterAmountFragment to Jetpack Compose can also solve the problem, since we can encapsulate both UI and behavior into a single component that's built from smaller ones. In this case, Gift Card Buy screen doesn't actually need to be shared with other similar screens like Send, Coinbase and Crowdnode since it's quite different (fiat-only input, fixed denominations), and instead it can be built anew but with same sub-components.

Syn-McJ (approved):
Looks good

PR #1398, opened 2025-05-22
Fix CTXSpend discount rounding issue

Syn-McJ:
## Summary
- send fiat to listeners for EnterAmount input
- store fiat amount in view model
- compare purchase limits using fiat
- show discount for the exact fiat amount entered

## Testing
- `./gradlew test --no-daemon` *(fails: No route to host)*

## Summary by CodeRabbit

- **New Features**
  - Added support for tracking and displaying fiat currency amounts alongside Dash amounts when entering values.
  - Users now receive real-time updates for both Dash and fiat values during amount entry.

- **Bug Fixes**
  - Improved app stability by preventing UI updates when the relevant screen is not visible or properly attached.

- **Refactor**
  - Updated internal logic to use fiat amounts for purchase limit checks and UI feedback.


HashEngineering (approved):
Excellent work.

PR #1400, opened 2025-06-05
fix: use original icon hash for id

Syn-McJ:
To maintain compatibility with iOS in the case of uploading icon metadata to the platform, we need to use the same IDs for the icons. Currently, the ID hash is calculated from the resized icon data, while it is nearly impossible to resize the icon in the same way across platforms.

- Use the original icon data to calculate ID instead of the resized one.

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency of icon image handling and hashing for transaction metadata across platforms. This ensures more reliable display and identification of transaction icons.


HashEngineering (approved):
LGTM

PR #1405, opened 2025-06-24
feat(dashspend): piggycards terms & conditions

Syn-McJ:
Use should be able to accepts PiggyCards terms and conditions

- Refactor the login dialog and some other code to support multiple card options
- Some bug fixing

## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:


common/src/main/java/org/dash/wallet/common/transactions/TransactionObserver.kt
```diff
@@ -13,13 +13,18 @@ suspend fun Transaction.waitToMatchFilters(vararg filters: TransactionFilter) {
         transactionConfidenceListener = TransactionConfidence.Listener { _, _ ->
             if (filters.isEmpty() || filters.any { it.matches(this) }) {
                 confidence.removeEventListener(transactionConfidenceListener)
-                continuation.resume(Unit)
+
+                if (continuation.isActive) {
+                    continuation.resume(Unit)
+                }
```
Syn-McJ, line 19: Looks like there is an occasional "already resumed" crash here

PR #1406, opened 2025-06-29
feat(dashspend): piggycards login

Syn-McJ:
We need to add OTP login for PiggyCards provider

- Implement PiggyCards API layer
- Refactor DashSpend to account for multiple gift card providers
- Rename CTXSpend -> DashSpend where makes sense
- Remove `ResourceResponse` from CTX API

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/repository/DashSpendRepository.kt
```diff
@@ -0,0 +1,31 @@
...
+interface DashSpendRepository {
+    val userEmail: Flow<String?>
+
+    suspend fun signup(email: String): Boolean
+    suspend fun login(email: String): Boolean
+    suspend fun verifyEmail(code: String): Boolean
+    suspend fun isUserSignedIn(): Boolean
+    suspend fun logout()
+//    suspend fun refreshToken(): Boolean TODO
+}
```
Syn-McJ, line 31: Only authentication methods are unified across ctx and piggycards repos for now, but everything else should be moved too to better conform to open-closed principle

PR #1414, opened 2025-07-22
feat(dashspend): piggycards provider picker

Syn-McJ:
We need UI for picking a gift card provider if multiple are available.

- Refactor Merchant/ATM details screen to Jetpack Compose
- Implement the new design for merchant details
- Add gift card provider picker view
- Add GiftCardProvider dao

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:


features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreDatabase.kt
```diff
@@ -128,7 +128,7 @@ abstract class ExploreDatabase : RoomDatabase() {
                             log.info("onOpenPrepackagedDatabase")
                         }
                     }
-                ).addMigrations(ExploreDatabaseMigrations.migration1To2, ExploreDatabaseMigrations.migration2To3)
+                ).addMigrations(ExploreDatabaseMigrations.migration1To2, ExploreDatabaseMigrations.migration2To3)//, ExploreDatabaseMigrations.migration3To4, ExploreDatabaseMigrations.migration4To5)
```
HashEngineering: a line is commented here.

Do we need a dummy migration for v3 to v4?

HashEngineering (changes requested):
one item to look at, otherwise good.

features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ExploreDatabase.kt
Syn-McJ: During my testing, I found that the dummy 3-4 migration wasn't necessary. But it might be better to have to for consistency and to avoid any potential issues.

HashEngineering (approved):
LGTM

PR #1501, opened 2026-06-25
feat(payment): support BIP21 op_return and callback URI parameters

sag333ar:
Parses two new BIP21 URI parameters and routes them through the existing payment flow:

  - `op_return=<hex>`: hex-encoded payload (up to 80 bytes) attached to the transaction as a 0-value OP_RETURN output. Used for cross-chain memos (THORChain/Maya), order IDs, and similar payee-side correlation data.

  - `callback=<https-url>`: https URL opened in a Chrome Custom Tab after a successful send so the merchant can continue its checkout flow. Only https is accepted at parse time to prevent a URI from invoking arbitrary app handlers.

The OP_RETURN payload is built via `ScriptBuilder.createOpReturnScript` and appended to the existing `outputs[]` array, so the existing broadcast path picks it up with no further changes. The dust check in bitcoinj's `TransactionOutput.isDust` already exempts 0-value OP_RETURN scripts, so no special-case handling was needed there.

`mergeWithEditedValues` now preserves auxiliary outputs (e.g. OP_RETURN) and the callback URL when the user edits the amount in the send screen — previously the non-primary outputs were dropped.

Both parameters are optional and additive — URIs that don't carry them behave identically to before.

Mirrors the corresponding iOS change in dashpay/dashsync-iOS and dashpay/dashwallet-ios.

## How Has This Been Tested?

- [ ] QA (Mobile Team)

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added support for BIP21 payment links with an optional secure callback URL.
  * After a successful payment, the app can now continue to the provided callback in a browser tab.

* **Bug Fixes**
  * Improved payment editing so changing the amount keeps the rest of the original outputs intact.
  * Preserved callback details when updating an existing payment.

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java
+++ b/common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java
@@ -24,6 +24,7 @@ import android.text.TextUtils;
 import com.google.common.io.BaseEncoding;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.Locale;
 
 import javax.annotation.Nullable;
 
@@ -192,6 +193,14 @@ public final class PaymentIntent implements Parcelable {
 
     public String source = "";
 
+    /**
+     * BIP21 {@code callback=<https-url>} parameter. When set, the wallet UI opens this URL after
+     * a successful send so the merchant / payee-side service can continue its checkout flow.
+     * Validated as {@code https} at parse time; arbitrary schemes are rejected.
+     */
+    @Nullable
+    public String callback = null;
+
     private static final Logger log = LoggerFactory.getLogger(PaymentIntent.class);
 
     public PaymentIntent(@Nullable final Standard standard, @Nullable final String payeeName,
@@ -265,15 +274,75 @@ public final class PaymentIntent implements Parcelable {
 
     public static PaymentIntent fromBitcoinUri(final BitcoinURI bitcoinUri) {
         final Address address = bitcoinUri.getAddress();
-        final Output[] outputs = address != null ? buildSimplePayTo(bitcoinUri.getAmount(), address) : null;
+        final Output[] baseOutputs = address != null ? buildSimplePayTo(bitcoinUri.getAmount(), address) : null;
         final String bluetoothMac = (String) bitcoinUri.getParameterByName(Bluetooth.MAC_URI_PARAM);
         final String paymentRequestHashStr = (String) bitcoinUri.getParameterByName("h");
         final byte[] paymentRequestHash = paymentRequestHashStr != null ? base64UrlDecode(paymentRequestHashStr) : null;
         final String dashPayUsername = bitcoinUri.getUser();
 
-        return new PaymentIntent(PaymentIntent.Standard.BIP21, null, null, outputs, bitcoinUri.getLabel(),
-                bluetoothMac != null ? "bt:" + bluetoothMac : null, null, bitcoinUri.getPaymentRequestUrl(),
-                paymentRequestHash, null, dashPayUsername);
+        // BIP21 op_return=<hex>: attach the payload as a 0-value OP_RETURN output so
+        // payee-side services (e.g. THORChain swap memos) can correlate the on-chain tx
+        // with their off-chain order. Invalid payloads are silently dropped per BIP21
+        // unknown-param convention.
+        final Output opReturnOutput = parseOpReturnParam(
+                (String) bitcoinUri.getParameterByName("op_return"));
+
+        // BIP21 callback=<https-url>: URL the wallet opens after a successful send so
+        // the merchant can continue its checkout flow.
+        final String callbackUrl = parseCallbackParam(
+                (String) bitcoinUri.getParameterByName("callback"));
+
+        final Output[] outputs;
+        if (baseOutputs == null) {
+            outputs = (opReturnOutput != null) ? new Output[]{opReturnOutput} : null;
+        } else if (opReturnOutput != null) {
+            outputs = new Output[baseOutputs.length + 1];
+            System.arraycopy(baseOutputs, 0, outputs, 0, baseOutputs.length);
+            outputs[baseOutputs.length] = opReturnOutput;
+        } else {
+            outputs = baseOutputs;
+        }
+
+        final PaymentIntent intent = new PaymentIntent(PaymentIntent.Standard.BIP21, null, null,
+                outputs, bitcoinUri.getLabel(),
+                bluetoothMac != null ? "bt:" + bluetoothMac : null, null,
+                bitcoinUri.getPaymentRequestUrl(), paymentRequestHash, null, dashPayUsername);
+        intent.callback = callbackUrl;
+        return intent;
+    }
+
+    @Nullable
+    private static Output parseOpReturnParam(@Nullable final String hex) {
+        if (hex == null || hex.isEmpty()) return null;
+        // 80-byte cap = Dash standard relay policy. Even-length + hex-only required
+        // before decode; use req-op_return= to make an invalid value invalidate the
+        // whole request (handled by BitcoinURI).
+        if (hex.length() % 2 != 0 || !hex.matches("[0-9a-fA-F]+")) {
+            log.info("dropping op_return param: not valid hex");
+            return null;
+        }
+        try {
+            final byte[] payload = Constants.HEX.decode(hex.toLowerCase(Locale.ROOT));
+            if (payload.length < 1 || payload.length > 80) {
+                log.info("dropping op_return param: payload size {} out of range (1..80)", payload.length);
+                return null;
+            }
+            return new Output(Coin.ZERO, ScriptBuilder.createOpReturnScript(payload));
+        } catch (final IllegalArgumentException x) {
+            log.info("dropping op_return param: decode failed");
+            return null;
+        }
+    }
+
+    @Nullable
+    private static String parseCallbackParam(@Nullable final String raw) {
+        if (raw == null || raw.isEmpty()) return null;
+        // https only: arbitrary schemes would let a URI invoke other app handlers.
+        if (!raw.toLowerCase(Locale.ROOT).startsWith("https://")) {
+            log.info("dropping callback param: not a valid https URL");
+            return null;
+        }
+        return raw;
     }
 
     private static final BaseEncoding BASE64URL = BaseEncoding.base64Url().omitPadding();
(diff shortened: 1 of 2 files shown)


farbodghasemlu:
Hey, I was going through the parsing path on this and ran into what looks like a crash in the op_return plus normal payment case, which is basically the main swap flow, so I wanted to write it up before it merges.

The root of it is `hasAddress()` in `PaymentIntent`:

```java
public boolean hasAddress() {
    if (outputs == null || outputs.length != 1)
        return false;

    final Script script = outputs[0].script;
    return script.isSentToAddress() || script.isPayToScriptHash() || script.isSentToRawPubKey();
}
```

It only considers an intent to "have an address" when there's exactly one output. That assumption was fine before, but now `fromBitcoinUri` appends the op_return as a second output, so a normal `dash:address?amount=...&op_return=...` URI ends up with `outputs.length == 2`. The moment that happens, `hasAddress()` returns false and `getAddress()` throws, since it does `if (!hasAddress()) throw new IllegalStateException()`.

That alone wouldn't be fatal if every caller checked first, and `initPaymentIntent` in `SendCoinsBaseViewModel` does exactly that:

```kotlin
if (paymentIntent.hasAddress()) {
    _address.postValue(paymentIntent.getAddress(Constants.NETWORK_PARAMETERS).toBase58())
}
```

But the confirmation path in `SendCoinsFragment.showPaymentConfirmation()` calls it unguarded:

```kotlin
val address = viewModel.basePaymentIntent.getAddress(Constants.NETWORK_PARAMETERS)?.toBase58() ?: return
```

The `?.` and `?: return` look like they cover the bad case, but they only handle a null return. `getAddress()` doesn't return null here, it throws, so the exception sails straight past the null check and takes down the confirmation screen. And because this is precisely the THORChain/Maya swap shape (real address plus a memo), it'd fire on the exact flow the PR is meant to enable.

Repro, using any valid mainnet address (a fake one gets rejected in address validation earlier and never reaches this code, so it has to be real):

```
dash:<address>?amount=0.01&op_return=68656c6c6f
```

I went looking for anything that strips the op_return output before confirmation and couldn't find it, so I think it reaches `getAddress()` with two outputs. I haven't put it on a device yet, so worth a quick check there to confirm.

On fixing it, there are two ways to go. The quick one is to guard the confirmation call the same way `initPaymentIntent` does, but that leaves `getAddress()` brittle for any future caller that forgets the check, and there are a few of them. The cleaner fix is to let `hasAddress()` and `getAddress()` tolerate extra data outputs and just pick the first real pay-to-address output, which keeps the single-output behaviour identical and the BIP70 no-address case still returning false:

```java
public boolean hasAddress() {
    return findAddressScript() != null;
}

public Address getAddress(NetworkParameters params) {
    final Script script = findAddressScript();
    if (script == null)
        throw new IllegalStateException();

    return script.getToAddress(params, true);
}

@Nullable
private Script findAddressScript() {
    if (outputs == null)
        return null;

    for (final Output output : outputs) {
        final Script script = output.script;
        if (script.isSentToAddress() || script.isPayToScriptHash() || script.isSentToRawPubKey())
            return script;
    }
    return null;
}
```

Feel free to just grab that if it's useful. Or if you'd rather, I'm happy to open a PR against your branch with the fix plus some PaymentIntent parsing tests, since that area doesn't have much coverage yet.

thepastaclaw:
⛔ Blockers found — Phase 2 deferred (commit 617126de) · triage: normal
_Canonical validated blockers: 1_

common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java
```diff
@@ -265,15 +274,75 @@ public static PaymentIntent from(final String address, @Nullable final String ad
...
+
+    @Nullable
+    private static String parseCallbackParam(@Nullable final String raw) {
+        if (raw == null || raw.isEmpty()) return null;
+        // https only: arbitrary schemes would let a URI invoke other app handlers.
+        if (!raw.toLowerCase(Locale.ROOT).startsWith("https://")) {
+            log.info("dropping callback param: not a valid https URL");
+            return null;
+        }
+        return raw;
```
thepastaclaw, line 345: **🟡 Suggestion: Callback parsing accepts malformed HTTPS-prefixed values**

parseCallbackParam() only checks whether the raw value starts with https://. Values such as https://, https://  evil.com, or strings containing whitespace therefore survive parsing and are later passed to the browser/custom-tab launch path. The scheme restriction is useful, but the callback should also be parsed as an absolute URI and require a non-empty host so malformed attacker-controlled payment links cannot trigger a failed or unexpected post-send navigation.

```suggestion
    @Nullable
    private static String parseCallbackParam(@Nullable final String raw) {
        if (raw == null || raw.isEmpty()) return null;
        final URI uri;
        try {
            uri = new URI(raw);
        } catch (final URISyntaxException x) {
            log.info("dropping callback param: malformed URL");
            return null;
        }
        if (!"https".equalsIgnoreCase(uri.getScheme()) || uri.getHost() == null || uri.getHost().isEmpty()) {
            log.info("dropping callback param: not a valid https URL");
            return null;
        }
        return uri.toString();
    }
```

<sub>source: `glm-5.3-flash` (phase1-reviewer: general, security-auditor)</sub>

thepastaclaw (changes requested):
## Preliminary review — Phase 1 blocker gate

The BIP21 OP_RETURN implementation breaks the existing single-output address invariant. A normal address-plus-OP_RETURN URI reaches an unguarded getAddress() call during confirmation and throws, so the PR's primary use case crashes the wallet. The same additional output also makes Send-MAX requests enter dashj's single-output empty-wallet path and fail; callback validation accepts malformed HTTPS-prefixed values.

Validated blockers were found by the Phase-1 review and confirmed by a fresh verifier. Phase 2 is deferred until a fresh same-head revalidation clears the blocker gate.

🔴 1 blocking | 🟡 2 suggestion(s)

### 2 finding(s) not shown inline (the lines are not part of this PR's diff)

**🔴 Blocking: Address-plus-OP_RETURN intents crash during payment confirmation**
`common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java:414-427`

hasAddress() returns false whenever outputs.length is not exactly one, while getAddress() throws when hasAddress() is false. fromBitcoinUri() now appends the OP_RETURN output, so a normal URI such as dash:<address>?amount=0.01&op_return=68656c6c6f produces two outputs and causes getAddress() to throw. SendCoinsFragment.showPaymentConfirmation() calls this method directly at line 288; the safe-looking ?. only handles a null return and cannot catch the exception. This deterministically breaks the address-plus-memo flow introduced by this PR. Update address detection to find the first address-bearing output while ignoring auxiliary outputs, preserving the existing no-address behavior.

```suggestion
    public boolean hasAddress() {
        return findAddressScript() != null;
    }

    public Address getAddress(NetworkParameters params) {
        final Script script = findAddressScript();
        if (script == null)
            throw new IllegalStateException();

        return script.getToAddress(params, true);
    }

    @Nullable
    private Script findAddressScript() {
        if (outputs == null)
            return null;

        for (final Output output : outputs) {
            final Script script = output.script;
            if (script.isSentToAddress() || script.isPayToScriptHash() || script.isSentToRawPubKey())
                return script;
        }
        return null;
    }
```

<sub>source: `glm-5.3-flash` (phase1-reviewer: general, security-auditor)</sub>

**🟡 Suggestion: Send-MAX with OP_RETURN enters dashj's single-output empty-wallet path**
`wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt:448`

The empty-wallet flag is derived from mayEditAmount && walletBalance == paymentIntent.amount without considering auxiliary outputs. For an OP_RETURN URI, the payment amount still equals the wallet balance, so Send-MAX sets emptyWallet=true even though the transaction has both the payment output and the OP_RETURN output. dashj rejects this with its requirement that an empty-wallet transaction have exactly one output. Restrict emptyWallet to intents with only the primary output, or implement an OP_RETURN-aware maximum-send calculation that does not use the empty-wallet mode.

```suggestion
        sendRequest.emptyWallet = mayEditAmount && paymentIntent.outputs.size == 1 && walletBalance == paymentIntent.amount
```

<sub>source: `glm-5.3-flash` (phase1-reviewer: general, security-auditor)</sub>


<details>

Source: reviewer 1: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-gate-verifier`, role: `verifier`)

- Triage: `normal` by `gpt-6-astra` (effort low) — The change modifies payment intent parsing and send-flow behavior, including auxiliary transaction outputs and post-payment callback handling, so it affects funds movement and requires more than a trivial review despite being relatively contained.
- Phase 1 reviewers: `glm-5.3-flash` — general (completed, effort max); agent `phase1-reviewer`, `glm-5.3-flash` — security-auditor (completed, effort max); agent `phase1-reviewer`
- Phase 1 model: `glm-5.3-flash` — zai quota: 5h 74% left, weekly 86% left; passed over `gemini-3.8-flash-high` (antigravity below 15% reserve: weekly 11% left, 5h 100% left)
- Fresh verifier: `gpt-6-astra` — verifier; agent `astra-gate-verifier`
- Phase 2 reviewers: **not run (deferred by blocker gate)**

</details>

<details>

```

In `common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java`:
- [BLOCKING] common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java:414-427: Address-plus-OP_RETURN intents crash during payment confirmation
  hasAddress() returns false whenever outputs.length is not exactly one, while getAddress() throws when hasAddress() is false. fromBitcoinUri() now appends the OP_RETURN output, so a normal URI such as dash:<address>?amount=0.01&op_return=68656c6c6f produces two outputs and causes getAddress() to throw. SendCoinsFragment.showPaymentConfirmation() calls this method directly at line 288; the safe-looking ?. only handles a null return and cannot catch the exception. This deterministically breaks the address-plus-memo flow introduced by this PR. Update address detection to find the first address-bearing output while ignoring auxiliary outputs, preserving the existing no-address behavior.
- [SUGGESTION] common/src/main/java/org/dash/wallet/common/data/PaymentIntent.java:337-345: Callback parsing accepts malformed HTTPS-prefixed values
  parseCallbackParam() only checks whether the raw value starts with https://. Values such as https://, https://  evil.com, or strings containing whitespace therefore survive parsing and are later passed to the browser/custom-tab launch path. The scheme restriction is useful, but the callback should also be parsed as an absolute URI and require a non-empty host so malformed attacker-controlled payment links cannot trigger a failed or unexpected post-send navigation.

In `wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt`:
- [SUGGESTION] wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt:448: Send-MAX with OP_RETURN enters dashj's single-output empty-wallet path
  The empty-wallet flag is derived from mayEditAmount && walletBalance == paymentIntent.amount without considering auxiliary outputs. For an OP_RETURN URI, the payment amount still equals the wallet balance, so Send-MAX sets emptyWallet=true even though the transaction has both the payment output and the OP_RETURN output. dashj rejects this with its requirement that an empty-wallet transaction have exactly one output. Restrict emptyWallet to intents with only the primary output, or implement an OP_RETURN-aware maximum-send calculation that does not use the empty-wallet mode.
```
</details>

PR #1505, opened 2026-06-30
feat: fix about me code issues

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed profile photo zooming so invalid crop areas no longer create thin artifact slivers.
  * Improved contact and user screens to refresh correctly after navigating away and back.
  * Made transaction result screens finish reliably after the related sheet is dismissed, even when the device rotates.
  * Corrected how contact rows and payment amounts are displayed and filtered.
  * Reduced logging of sensitive profile image details and improved image loading efficiency.

PR #1504, opened 2026-06-30
feat(dash-dex): finish buy swap flow

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

## Summary by CodeRabbit

* **New Features**
  * Added a new **Dash DEX** shortcut and home-screen entry point for quicker access.
  * Introduced a new buy/sell DEX flow with dedicated amount, refund address, and receive screens.
  * Added clearer address entry and balance controls, plus localized balance visibility labels.

* **Bug Fixes**
  * Improved address validation to better reject invalid crypto addresses.
  * Refined search/no-results messaging and fixed layout sizing in amount selection.
  * Updated receive screens to show QR, expiry guidance, and copy actions more reliably.

PR #1500, opened 2026-06-19
feat: update about-me activity list UI to designs

HashEngineering:
## How Has This Been Tested?

- [x] QA (Mobile Team)

## Checklist:

- [ ] I have performed a self-review of my own code and added comments where necessary
- [ ] I have added or updated relevant unit/integration/functional/e2e tests

## Summary by CodeRabbit

* **New Features**
  * Added DashPay user interaction bottom sheet displaying user profile, request status, and activity history.
  * Activity section supports filtering by type (All, Received, Sent) with notifications organized by calendar date.
  * Displays absolute timestamps and formatted payment amounts for transactions.
  * Supports contact requests, payments, and transaction detail viewing.

Issue #1523, opened 2026-07-28
SDK cutover: merge v11.9.0 into the cutover baseline (currently forked from v11.8.2)


bfoss765:
## Context
The SDK-cutover work (`build-11105` worktree / every `11.10.x` QA build) **forked from `v11.8.2`** and has **never incorporated `v11.9.0`**. v11.9.0 = v11.8.2 + **144 commits** that are entirely missing from the cutover line.

Verified: `git merge-base --is-ancestor <v11.8.2> <cutover>` = YES (11.8.2 is the base); `git merge-base --is-ancestor <v11.9.0> <cutover>` = NO. The `versionName "11.8.2"` fallback default in `build.gradle` reflects the true baseline; the `11.10.x` version number is only an `-P` override.

## What's missing (the v11.9.0 changeset)
- **DEX** buy/sell swaps (SwapKit `/track`, swap status on home, provider explorer links, decimal quantization)
- **Maya** improvements (conversion-result full-height sheet, settlement explorer link, config-change error retention)
- **DashSpend / gift cards** (generated merchant icons, gift-card cache fixes)
- **About-screen redesign** (#1498 / #1503 / #1505)
- **Identity / crash / ANR fixes** (legacy identity-cache CBOR tolerance, Topper ANR → Dispatchers.IO, PIN-change crash, ContactsFragment back-nav, etc.)

## Task
Merge `v11.9.0` into the cutover branch before it ships as `11.10.x`. Expect **heavy conflicts** — v11.9.0 is DEX/Maya/send-heavy and the cutover modified `SendCoinsFragment`, `SendCoinsViewModel`, `BuyCredits*`, and the DEX files. This is a dedicated, focused merge effort (not to be folded into a point-fix build).

## Relationship to Phase 1B (#1520)
#1520 lists the v11.9.0 merge as a **prerequisite** for a few of its items (SwapKit isn't even in the cutover build without it). This issue is the **dedicated tracking issue for the merge itself**; #1520 covers migrating the fail-closed send/integration paths that the merge brings in.

## Acceptance
`v11.9.0` is fully merged into the cutover line, conflicts resolved, the app builds all flavors, and the DEX/Maya/DashSpend/About features from 11.9.0 are present and functional on a cutover build.

bfoss765:
@HashEngineering flagging a schema-version conflict that needs handling as part of this merge.

**`AppDatabase` version on the cutover branch is 20; on v11.9 it is 21.**

[...]

PR #1525, opened 2026-07-30
merge: master (Dash DEX, about-me redesign) into kotlin-sdk-phase1

HashEngineering:
## What this is

A true merge commit (`f73bf13e2`, parents `b6f6f2109` + `d4c04952b`) bringing `origin/master` into the Kotlin SDK phase1 line, with master's dashj/bitcoinj-based additions re-expressed on phase1's neutral (dashj-free) module APIs.

## What master brought in

- Dash DEX buy-swap flow completion (#1504), swap tracking + Maya screen redesign (#1511), many sell-swap fixes (#1519)
- About-me / contact details redesign (#1486, #1500, #1503, #1505)

## Conflict + auto-merge fallout resolution

Only one file had textual conflicts (`SwapKitApiAggregator.kt`), but ~20 more files auto-merged master's bitcoinj code into dashj-free modules. All were re-expressed on the neutral APIs, keeping master's behavior:

- **SwapKitApiAggregator**: kept phase1's neutral types (FiatValue, InsufficientFundsException, string-address sendCoins); adopted master's max-sell sweep quoting, pre-broadcast sweep re-check, error mapping (errorMessageRes / isAmountTooLowError), HydratedSnapshot refactor, lastFiatExchangeRate tracking
- **SwapOrder / SwapOrderDao**: keyed by neutral `TxId` instead of `Sha256Hash`; call sites fixed (conversion viewmodel, WalletTransactionMetadataProvider, TransactionResultViewModel)
- **Example-address generation** (AddressGenerator, MayaCryptoCurrency): random `SessionExampleKey` + vendored Base58/SHA-256 replaces ECKey/Sha256Hash; BTC validation via `AddressUtils.verify`
- **Foreign-chain parsers** (Solana/XRP/TRON/Cardano/XRD + Bech32/Zcash intent parsers): vendored Base58/AddressFormatException
- **MayaBlockchainApi** stays interface-only in the maya module; master's CancellationException rethrow (double-swap guard) ported to the wallet-side impl
- **ContactsFragment**: phase1's per-view ContactsIdentityGate kept; master's onDestroyView flag-reset workaround dropped as redundant
- Misc: segwitHrp rename, currentReceiveAddressString(), AddressNetwork.fromId(networkId), LocalDashColors in SettingsScreen, OP_RETURN memo extraction in intent-parser tests

## Verification

- `:common`, `:integrations:maya`, `:wallet` (_testNet3Debug) compile
- All common + maya unit tests pass, including the new 22-test example-address suite


HashEngineering:
@coderabbitai - review this only concerning the merge changes.

HashEngineering:
### Heads-up: this merge silently reverts a `* 1000` in `NotificationItemPayment.getDate()`

Flagging because the resolution changes behaviour rather than just picking a side, so it should be a conscious call rather than something discovered later.

**What conflicted.** `88a7d7516` (post-cutover completeness) added SDK-only row support to `NotificationItemPayment` and, in the same edit, reintroduced a `* 1000` on `getDate()`:

```kotlin
// 88a7d7516
override fun getDate() = tx?.let { it.updateTime.time * 1000 } ?: (correctedDisplay!!.time * 1000)
```

master has the opposite, with a comment explaining why (I confirmed master's file is byte-identical to what this branch carried in):

```kotlin
// master
// updateTime.time is already epoch milliseconds; every other NotificationItem returns
// millis too, so the sort key and relative-time display stay consistent across types.
override fun getDate() = tx!!.updateTime.time
```

**Why I went with master's scale.** Every other row type returns true epoch millis:

| Type | `getDate()` |
|---|---|
| `NotificationItemContact` | `usernameSearchResult.date` |
| `NotificationItemStub` | `System.currentTimeMillis()` |
| `NotificationItemUserAlert` | `createdAt` |

`Transaction.updateTime` is a `java.util.Date`, so `.time` is already millis, and `88a7d7516`'s own comment describes `correctedDisplay.time` as "epoch-millis". Multiplying either by 1000 puts payment rows ~1000x into the future relative to contact/alert rows in the mixed-type sort, and breaks relative-time rendering ("in 55,000 years" territory). Within payment rows only, sorting still looks fine — which is probably why it wasn't noticed.

**How it is resolved here** (`wallet/src/de/schildbach/wallet/data/NotificationItemPayment.kt:43`) — phase1's new SDK-only support kept, master's scale restored:

```kotlin
val txId: Sha256Hash
    get() = tx?.txId ?: Sha256Hash.wrap(correctedDisplay!!.rowId)

override fun getId() = txId.toString()

// Both sources are ALREADY epoch milliseconds ... (do not reintroduce a `* 1000`)
override fun getDate() = tx?.updateTime?.time ?: correctedDisplay!!.time
```

The `txId` property and null-`tx` handling from `88a7d7516` are preserved in full — a post-cutover row still opens the detail dialog without a dashj `Transaction`.

**If the `* 1000` was deliberate**, say so and I'll revert this hunk — but then the other three row types need the same treatment, or mixed-type ordering stays wrong. My read is that it was an accidental carry-over from the older code the SDK-only support was written against.

Not otherwise verifiable at runtime: it only manifests in mixed-type notification sort order, which the emulator smoke test I ran wouldn't surface.

bfoss765 (approved):
Reviewed in depth (schema, cutover-machinery preservation, dashj re-introduction, local build+tests). Approving.

- **AppDatabase**: lands on v21 with master's additive 20→21 migration (`swap_orders` + one ADD COLUMN). QA devices at v20 migrate non-destructively (display caches preserved — history stays on screen through the SDK sync); shipped v11.9 devices at v21 get a byte-identical schema hash. The divergence we feared on #1523 never existed — v20 was shared history.
- **Cutover work preserved**: the PR contains `4978b2920` as a direct parent; `SendCoinsTaskRunner`/`CutoverCoordinator`/`BlockchainServiceImpl`/`CutoverUiDataService` are byte-identical to the current head, and a grep of removed lines for cutover/sync/balance/parity terms is empty.
- **No dashj re-introduction** into common/features/integrations; engine choice stays below the seam.
- Local `:wallet:compile_testNet3DebugKotlin` and full `:wallet:test_testNet3DebugUnitTest` pass. The red CI `build` check is the known unresolvable pinned-AAR issue inherited from the base (see PR comment thread), not a defect here.

Two doc-only nits for any follow-up, not blocking: the merge reverted `IdentityCacheTolerance.kt`'s updated KDoc to stale text, and `PlatformService.kt` carries a duplicated stale KDoc block above `getContactIdentity`.

PR #1530, opened 2026-07-30
fix(ci): repair ktlint violations and the uphold unit-test org.json classpath

HashEngineering:
Makes `feat/kotlin-sdk-phase1` pass CI. Both failures are pre-existing on phase1 — they only became visible now because the branch has never had a PR of its own, so [#1525](https://github.com/dashpay/dash-wallet/pull/1525) is the first CI run over this code. Neither was introduced by that merge.

## `check` — ktlintCheck: 23 violations in 3 modules

CI reported only crowdnode because the first failing task aborts the run; the rest were hidden behind it. Full list:

| Module | Violations |
|---|---|
| `integrations/maya` (main) | 14 import-ordering + 1 unused import |
| `integrations/maya` (test) | 2 missing final newline, 1 import-ordering, 1 max-line-length |
| `integrations/crowdnode` (test) | 1 import-ordering |
| `features/exploredash` (main) | 1 import-ordering |

All applied with `ktlintFormat` except the over-length line, which the formatter can't fix: the Shelley test address in `AddressParserTest` is now split across two literals to fit 120 columns. I verified the concatenation is byte-identical to the original, so the assertion is unchanged. (The 125-char line inside the adjacent raw string is untouched — ktlint exempts raw-string content, and editing it would change the parser's test input.)

## `build` — 8 of 11 uphold unit tests failed

```
java.lang.RuntimeException: Method getJSONObject in org.json.JSONObject not mocked.
  at ...uphold.data.UpholdApiException.getErrorsWithException(UpholdApiException.java:81)
```

`android.jar` ships a **stub** `org.json` whose methods throw under JVM unit tests. A real implementation had been reaching the test classpath transitively from `org.dashj:dashj-core` — its POM declares `org.json:json:20231013` at runtime scope. Phase1 removed dashj from this module, and jjwt's own copy is deliberately excluded (it must not be packaged into the APK), so nothing supplied it any more.

Fix: `testImplementation 'org.json:json:20231013'` — test-scope only, so the APK is unaffected — and the stale `//provided by Android natively` comment now says why unit tests are the exception. **Production code was never affected**: a real device supplies `org.json`, which is why this only ever showed up in unit tests.

## Verification

- `./gradlew ktlintCheck` — clean
- `./gradlew testDebugUnitTest` — 118 tests, 0 failures across common, exploredash, coinbase, crowdnode, maya, uphold (uphold now 11/11, previously 8 failing)

Both commands match what CI runs (`ktlint.yml` runs `ktlintCheck`; `dashwallet.yml` runs `fastlane test` → `testDebugUnitTest`).

Once this lands on phase1, re-merging or rebasing #1525 should turn it green.


bfoss765 (approved):
Reviewed the full diff: 19 files are pure ktlint formatting (no logic changes — verified no non-import added lines). The two substantive changes are safe: (1) uphold/build.gradle adds org.json only at test scope with the APK exclude retained, restoring the real impl that dashj-core used to supply transitively; (2) the AddressParserTest Shelley address split is byte-identical to the original. No wallet-module changes, so no conflict with in-flight work. LGTM.

Issue #1522, opened 2026-07-27
SDK cutover — Phase 3: SDK owns the wallet / keys / persistence — remove dashj


bfoss765:
## Goal

Retire dashj as the **wallet / key / persistence authority**, then remove the `org.bitcoinj` dependency entirely. Today the SDK **mirrors** the wallet — it derives a parallel SDK-side wallet from the same mnemonic (`service/platform/sdk/SecurityGuardMnemonicProvider.kt:122` reads `wallet.keyChainSeed`), but dashj's `Wallet`/`WalletEx` remains the source of truth for keys, seed, encryption, and persistence. Phase 3 makes the SDK **own** these, replacing the mirror, so dashj can be deleted.

Source: the 6-domain dashj-dependency audit of `build-11105`. This is the largest, structural phase and is gated on Phases 1 and 2.

## Foundation the SDK must own (currently dashj-only — "Pile A")

1. **The canonical wallet object** — `WalletApplication.wallet` / `WalletEx` (`wallet/src/de/schildbach/wallet/WalletApplication.java:194,824`). L1 balance/UTXO/keys must live in the SDK, not a mirror.
2. **Wallet-file persistence** — `WalletProtobufSerializer` read/write, `saveToFile`, `autosaveToFile`, `walletToProto` (`WalletApplication.java:711,858,958,966`; `util/WalletUtils.java:33-41`; `util/Crypto.java:30`). No SDK serializer today.
3. **BIP39 seed ownership** — `keyChainSeed` / `DeterministicSeed` is the source of truth; the SDK's `PlatformMnemonicProvider` reads it out. SDK must own seed generation/storage.
4. **Wallet encryption / spending-password crypto** — `KeyCrypterScrypt`, `deriveKey`, `wallet.encrypt/decrypt`, seed `decrypt` (`livedata/EncryptWalletLiveData.kt:135-177`, `ui/dashpay/PlatformRepo.kt:151-183`).
5. **Standalone HD derivation** — `HDKeyDerivation`, `ChildNumber`, `BIP44_PATH` (`ui/RestoreWalletFromSeedViewModel.kt:73,87`, `Constants.java:96,319`).
6. **Masternode / identity auth keychains** — `AuthenticationKeyChain(Group)`, `AuthenticationGroupExtension`, `Ed25519DeterministicKey` (`WalletApplication.java:494-530`, `ui/more/masternode_keys/MasternodeKeysViewModel.kt:141-227`). No SDK masternode-key service today.
[...]

Issue #1521, opened 2026-07-27
SDK cutover — Phase 2: retire dashj as the consensus / network engine


bfoss765:
## Goal

Retire dashj as the **consensus / network engine**. Today, even after a committed cutover, the dashj `BlockChain` / `SPVBlockStore` / `DashSystem` objects are still constructed and disk-loaded every launch (`wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.kt:1446-1448,1621-1622`), the wallet stays attached to an idle `BlockChain`, and several consensus reads still resolve against dashj — returning **stale/zero** because nothing feeds them post-cutover. Phase 2 makes the SDK the source of all consensus data and stops constructing the dashj network engine entirely.

Source: the 6-domain dashj-dependency audit of `build-11105`.

## What still reads dashj post-cutover (and must get an SDK feed)

1. **Chainlock height** — `BlockchainState.chainlockHeight` is frozen at the cutover snapshot; `SdkBlockchainStateMapper` never sets it (documented gap, `service/BlockchainStateDataProvider.kt:153,215`). Consumers read stale/zero:
   - `ui/DashPayUserActivityViewModel.kt:167`
   - `ui/dashpay/notification/NotificationsViewModel.kt:73`
   - `ui/transactions/TransactionGroupViewModel.kt:100`
   - **To do:** SDK chainlock-height feed → `SdkBlockchainStateMapper` populates `chainlockHeight`.

2. **InstantSend / ChainLock tx confidence** — `ui/transactions/TxResourceMapper.kt:166-185` reads dashj `TransactionConfidence.isChainLocked` / `IX_LOCKED`; not updated for SDK-sourced txs (no peergroup delivering IS/CL signatures).
   - **To do:** SDK-sourced IS/CL status per tx, surfaced on the confidence/row model.

3. **Masternode list (identity voting-key lookup)** — `service/platform/PlatformBroadcastService.kt:302` queries the dashj MN list (`masternodeListManager.masternodeList.getMasternodesByVotingKey`), empty post-cutover (bridge manager, `BlockchainServiceImpl.kt:1693-1695`).
   - **To do:** resolve voting keys from the SDK masternode/quorum source.

## Engine teardown

4. Once (1)-(3) are SDK-fed, stop constructing the idle dashj network engine on launch:
[...]

Issue #1520, opened 2026-07-27
SDK cutover — Phase 1B: migrate fail-closed / dashj-only send & integration paths (BIP70, coin-selection, asset-lock, Maya, CrowdNode, SwapKit)


bfoss765:
## Context

The SDK cutover routes **simple** L1 sends through the SDK post-cutover (`SdkL1SendService` + `CoreSendAllNative`, bridged back via `SdkBridgedTransactionFactory`). Everything else in the send funnel currently **fails closed** — `SendCoinsTaskRunner.cutoverSendRoute` (`wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt:123-139,898-923`) throws `IllegalStateException` once the dashj peergroup is held, because the tx would be dashj-built and cannot be broadcast.

**Phase 1A** (core L1/L2 sync + username/invite creation) completes when invites work. The items below are the remaining Phase-1 functions that still build dashj transactions or use dashj-only signing; each needs an SDK-routed equivalent (or an explicit decision to keep it dashj-backed and gate it out of cutover). Split into **Phase 1B** for hand-off.

Source of the file:line references: the 6-domain dashj-dependency audit of `build-11105`.

---

### 1. BIP70 / BIP270 direct payment
- **Where:** `SendCoinsTaskRunner.kt:640-788` (`PaymentProtocol.createPaymentMessage`, `wallet.completeTx`, `tx.unsafeBitcoinSerialize()`, `wallet.freshAddress(REFUND)`, `maybeCommitTx`), UI in `ui/send/PaymentProtocolFragment.kt` / `PaymentProtocolViewModel.kt`.
- **Now:** fails closed post-cutover.
- **To do:** SDK path to build + sign + broadcast a payment-request tx (single/multi output), produce the serialized tx + refund address for the BIP70 `Payment` message, and commit the returned tx via `SdkBridgedTransactionFactory`.

### 2. Custom coin-selection sends
- **Where:** `transactions/DashjSelectors.kt` (`ByAddressCoinSelector`, `ExactOutputsSelector`), `payments/MaxOutputAmountCoinSelector.kt`, `payments/ChainLockedCoinSelector.kt`. Routed to fail-closed at `SendCoinsTaskRunner.kt:183,527-535`.
- **Now:** SDK exposes only `SelectionStrategy` (all / max) — no equivalent for "spend only these inputs / addresses".
[...]

PR #1531, opened 2026-07-31
feat(cutover): BIP70/BIP270 via the SDK deferred build/broadcast surface (Phase 1B item 1)

HashEngineering:
## What

Implements the BIP70 item of #1520 (Phase 1B item 1) under the **replace-then-delete policy** (management directive, final): the BIP70/BIP270 direct-pay leg is implemented on the Kotlin SDK's deferred build/broadcast surface, and the dashj implementation it replaces is **deleted in this same PR**. dashj remains in this flow only as Phase-3 foundation (#1522: the wallet object as display/confidence record, keys, persistence) plus one transition-scoped guard that retires with Phase 2 (#1521).

> **Note for reviewers:** earlier revisions of this PR *gated* the SDK route on the committed cutover and kept the dashj leg intact ("provably inert pre-cutover"). That design was reviewed favorably — and then superseded by the management policy above. The final revision routes BIP70 through the SDK **unconditionally** and removes the dashj leg (`e4f34bd3e`, −473/+129).

Commits in review order:

1. **SDK deferred route** — `buildSignedPayment` (platform#4185 token surface in the pinned AAR) builds + signs with inputs reserved; the BIP70 `Payment` message carries the raw signed bytes; merchant **ack** → `broadcastSigned` → bridge commit via `SdkBridgedTransactionFactory` (the #1520-prescribed record mechanism); nack/transport failure → `releaseReservation`; post-ack never releases (double-pay hazard). Recipients via pure `extractBip70Recipients` (P2PKH/P2SH only; anything else fails closed with the typed `SendNotSdkRoutableException`).
2. **Refund address from the SDK** — `refund_to` from the SDK's Room address-pool mirror (lowest unused external, the KotlinExampleApp/iOS pattern); swaps to `coreWallet().nextReceiveAddress()` when dashpay/platform#4260 lands in a pinned AAR.
3. **The preview IS the payment** — `PaymentProtocolViewModel` builds the deferred payment at preview time (exact fee shown), submits that same tx on confirm, releases on abandon; `Bip70AckedDisplayException` marks post-ack display failures non-retryable.
4. **Test tooling** — `scripts/bip70-test-server.py` + CLAUDE.md recipe (local invoice server; drives the full scanned-invoice flow via `adb reverse` + a `dash:?r=` intent).
5. **Post-rebase alignment** — typed `SendNotSdkRoutableException`; base's own test-file fixes adopted.
6. **Double-submission race fix** (review finding) — `AtomicReference.getAndSet(null)` take + synchronous single-flight guard in `sendPayment()`, with `PaymentProtocolViewModelRaceTest` covering duplicate-confirm and retry-after-failure.
7. **`e4f34bd3e` — the deletion** — dashj `directPay` leg, ViewModel dashj preview/send branches, `baseSendRequest`/fragment `SendRequest` plumbing, `isTransactionOnNetwork` bloom rescue, and the 15 legacy dashj-path tests. Plus the **reservation mirror** (transition-only, tagged for the Phase 2 kill list): the SDK's reserved input outpoints are locked in the foundation wallet via `Wallet.lockOutput` — the same mechanism that protects CrowdNode coins from the background CoinJoin mixer — so dashj-side spenders on not-yet-cut-over installs cannot double-select them mid-preview; unlocked on release/ack, with a regression test on real fixture outpoints.

## Accepted per policy (explicit, so nobody is surprised)

- BIP70 requires the SDK funding gate on **all** installs — cold start / mid-first-scan purchases fail with the typed "not synced" UX where dashj would previously have paid.
- `ROLLBACK_CUTOVER` no longer restores a dashj BIP70 leg.
- ~~The pre-cutover mempool timeout rescue (dashj bloom view) is gone~~ **Rebuilt SDK-side and field-verified** (`437cf5ffa`, corrected in `4f28e70e2`): on a transport failure after the Payment POST, the wallet polls the SDK store for ~20s; a row for that txid proves the transaction reached the network — the engine only records transactions it has observed, and a built-but-unbroadcast payment has no row (verified against real txids) — so the payment completes as paid. An explicit nack never rescues.

  *Two defects the on-device test caught and fixed:* the first version keyed on `context >= instantSend`, which can never fire — the pinned engine takes an externally broadcast tx `context 0 → 2` and never surfaces an InstantSend context or `isInstantLocked`, even when the network has IS-locked it. And the ViewModel used to re-arm the preview after an ambiguous failure: because the engine cannot see the mempool spend of the released inputs, the rebuild selects **different** inputs, so a retry tap would pay the merchant twice (reproduced: rebuild `ed08074b` vs on-chain `2c3be7d8`). The preview now re-arms only on a definitive nack.

## Verified

- On-device, testnet, committed cutover: CTX gift card tx `62aa6214…0864`; scanned-invoice preview flow tx `7d9bd5fb…965f` (exact fee at preview, same txid on confirm, both on the explorer); cancel path tx `a6cc9ead…9d24` (build → back out → `l1DeferredRelease`, no POST, no broadcast); and the **timeout rescue** tx `ed08074b…6d39` — a test server broadcast the payment to testnet then returned HTTP 500, the engine observed the transaction **1.3 s** later, and the payment completed as paid with the reservation kept and the tx bridged (on-chain, `txlock: true`). The earlier runs predate the deletion commit; the SDK path they exercised is unchanged by it.
- Unit: payments / send-UI / SDK suites green, including the AAR canary (v41int11 pin), the race tests, and the new lock-mirror test.

## CI note

The build job fails resolving the pinned `dash-sdk-android` snapshot (local-Maven only); the base branch fails identically. Format check is green.

## Phase 2/3 follow-ups this flow will need (tracked, not in scope)

Lock mirror removal (#1521), IS/CL badges for bridged txs via the SDK feed (#1521 item 2), bridge/display migration off the dashj `Transaction` (#1522), Room→FFI refund-address swap (platform#4260), neutral-types sweep (#1522 items 10/14–16).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,32 @@ adb install wallet/build/outputs/apk/dash-wallet-_testNet3-debug.apk
 ./gradlew build
 ```
 
+### BIP70 payment-request testing (device/emulator)
+
+`scripts/bip70-test-server.py` (stdlib Python, no dependencies) serves a
+local BIP70 invoice and acks the returned Payment — the fastest way to
+exercise the scanned-invoice flow (`PaymentProtocolFragment`: fetch →
+preview → confirm → Payment POST → ACK) end to end. By default the invoice
+pays 0.01 tDASH back to the Dash testnet faucet
+(`yjSvwyLB5X4dqQqVMPMu6UdrFpYZ3u9v5U`); pass your own receive address for a
+fee-only self-pay.
+
+```bash
+# Terminal 1: serve the invoice (defaults: faucet address, 0.01 tDASH, :8330)
+python3 scripts/bip70-test-server.py
+
+# Terminal 2: bridge the port and open the invoice in the wallet
+adb reverse tcp:8330 tcp:8330
+adb shell am start -a android.intent.action.VIEW \
+    -d "dash:?r=http://127.0.0.1:8330/invoice" hashengineering.darkcoin.wallet_test
+```
+
+Post-cutover (`CUT_OVER`) watch logcat for `l1DeferredBuild` when the
+preview opens (SDK builds + reserves the tx, exact fee shown) and
+`l1DeferredBroadcast` of the same txid after confirm — see
+`SendCoinsTaskRunner.directPayViaSdk`. The server can reply `nack` (edit
+`do_POST`) to exercise the reservation-release path.
+
 ### Code Quality
 ```bash
 # Format Kotlin code with ktlint
--- a/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dashspend/dialogs/GiftCardDetailsViewModel.kt
+++ b/features/exploredash/src/main/java/org/dash/wallet/features/exploredash/ui/dashspend/dialogs/GiftCardDetailsViewModel.kt
@@ -274,7 +274,7 @@ class GiftCardDetailsViewModel @Inject constructor(
                                     cancelTicker()
                                     state
                                 } else if (giftCard.redeemUrl?.isNotEmpty() == true) {
-                                    log.error("CTXSpend returned a redeem url card: not supported")
+                                    log.info("CTXSpend returned a redeem url card")
                                     updateGiftCardWithURL(index = 0, giftCard.redeemUrl, giftCard.redeemUrlChallenge)
                                     val newState = uiState.value.copy(
                                         status = giftCard.status,
--- a/scripts/bip70-test-server.py
+++ b/scripts/bip70-test-server.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+"""Minimal BIP70/BIP270 test server for the Dash testnet wallet (stdlib only).
+
+Serves a one-output PaymentRequest and acks the returned Payment — enough to
+exercise the wallet's full scanned-invoice flow (PaymentProtocolFragment:
+fetch -> preview -> confirm -> Payment POST -> ACK) against a device or
+emulator, with no external service and no real network.
+
+Endpoints:
+  GET  /invoice  -> application/dash-paymentrequest (PaymentRequest protobuf)
+  POST /pay      -> application/dash-paymentack     (echoes the Payment in an ACK)
+
+Usage (device connected via adb):
+  python3 scripts/bip70-test-server.py [dest_address] [amount_duffs] [port]
+  adb reverse tcp:8330 tcp:8330
+  adb shell am start -a android.intent.action.VIEW \
+      -d "dash:?r=http://127.0.0.1:8330/invoice" hashengineering.darkcoin.wallet_test
+
+Defaults: pay 1_000_000 duffs (0.01 tDASH) to the Dash testnet faucet
+(https://faucet.testnet.networks.dash.org/ — same hot wallet as
+faucet.thepasta.org, the FAUCET_URL in de.schildbach.wallet.Constants), so
+test payments recycle back into the faucet pool. Pass your own wallet's
+receive address instead for a self-pay that only costs the fee.
+
+Post-cutover (state CUT_OVER) the wallet routes this through the SDK deferred
+build/broadcast surface: expect `l1DeferredBuild` in logcat when the preview
+opens (exact fee, inputs reserved), and `l1DeferredBroadcast` of the SAME
+txid after confirm+ACK. Pre-cutover it exercises the legacy dashj path.
+"""
+import sys
+import time
+from http.server import BaseHTTPRequestHandler, HTTPServer
+
+# The Dash testnet faucet's hot wallet (faucet.testnet.networks.dash.org /
+# faucet.thepasta.org — verified on-chain: 1 tDASH fan-out payouts).
+FAUCET_ADDRESS = "yjSvwyLB5X4dqQqVMPMu6UdrFpYZ3u9v5U"
+
+B58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+
+
+def b58decode_check(s: str) -> bytes:
+    n = 0
+    for ch in s:
+        n = n * 58 + B58.index(ch)
+    raw = n.to_bytes(25, "big")
+    return raw[:-4]  # version byte + hash160 (checksum not re-verified here)
+
+
+def p2pkh_script(address: str) -> bytes:
+    payload = b58decode_check(address)
+    h160 = payload[1:21]
+    return b"\x76\xa9\x14" + h160 + b"\x88\xac"
+
+
+def varint(n: int) -> bytes:
+    out = b""
+    while True:
+        b = n & 0x7F
+        n >>= 7
+        out += bytes([b | (0x80 if n else 0)])
+        if not n:
+            return out
+
+
+def field(num: int, wire: int, payload: bytes) -> bytes:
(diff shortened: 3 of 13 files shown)


HashEngineering:
@coderabbitai review

bfoss765:
Flagging CodeRabbit's finding at `PaymentProtocolViewModel.kt:231` as worth addressing before merge: the read-then-null of `deferredPayment` across overlapping coroutines is a genuine double-submission race on the SDK deferred-broadcast surface (two rapid confirmations could submit the same payment twice). A simple fix is an atomic getAndSet(null) (e.g. `AtomicReference`) or confining the read-clear to a single dispatcher. The detekt SwallowedException in the BIP70 test is minor by comparison. Happy to help if useful — this PR stacks on the cutover branch we maintain. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

HashEngineering:
**Note on the red CI build:** the failure is inherited from the base branch, not introduced here. The build job fails with `Could not find org.dashj:dash-sdk-android:0.1.0-v41int11-SNAPSHOT` — that pinned snapshot AAR only exists in local Maven (published from the platform repo's kotlin-sdk package), so CI cannot resolve it. `feat/merge-kotlin-sdk-master` fails its own CI identically at the same head (93c68c32e). The format check is green on this PR.

Local verification on the PR head: unit suites 686/687 (the one failure is the pre-existing pinned-AAR `coreSendMutex` canary), plus on-device testnet runs under a committed cutover — CTX gift-card purchase tx `62aa6214…0864`, and a scanned-invoice preview-screen payment tx `7d9bd5fb…965f` (built at preview with the exact 260-duff fee, same txid submitted on confirm, acked, broadcast, bridged; both visible on the testnet explorer).

bfoss765 (changes requested):
Requesting changes on two items; everything else looks solid — the deferred build→broadcast contract is implemented correctly (reservations released under `NonCancellable` on pre-ack HTTP failure and nack, never after ack), the runner-level tests are good, and gating rather than deleting the dashj BIP70 leg matches this branch's established pattern.

**1. Double-submission race — still live at `PaymentProtocolViewModel.kt` ~:223-231.** `val prebuilt = deferredPayment; … deferredPayment = null` runs inside `viewModelScope.launch(Dispatchers.IO)` on a plain `var`, and the fragment has no in-flight guard. Two rapid `sendPayment()` calls both read the same reservation: both submit the same txid (no direct double-spend), but the loser's exception path calls `createBaseSendRequest` → builds a fresh reservation for the same invoice and can post `Resource.error` over the winner's success — the user sees "failed" for a paid invoice with the payment re-armed, and one retry tap is a genuine double-pay with different inputs. Fix is small: take the payment atomically (`AtomicReference.getAndSet(null)` or read+clear on main before the launch) plus an is-sending guard, with a regression test.

**2. Rebase needed — the PR currently cannot merge.** The base moved (#1525 just merged, plus our send-path work `eccd9e7d9`): the one real conflict is `SdkL1SendService.kt`, where both sides independently added `ACCOUNT_TYPE_TAG_STANDARD`/`STANDARD_ACCOUNT_TAG_BIP44` — a forced merge is a duplicate-declaration compile error. `SendCoinsTaskRunner` auto-merges cleanly with the intendedRecipient/typed-exception work. Good news: the 29 failing tests at this PR's current base are all already fixed on the new base, so they disappear with the rebase.

Post-rebase nit, non-blocking: throw `SendNotSdkRoutableException` rather than a bare `IllegalStateException` for the not-routable case, for consistency with the new typed send exceptions.

Since the conflict is with code we wrote, happy to handle the rebase + constant dedupe ourselves if that's easier — say the word; the race fix and its test are probably best done by whoever picks it up first, just flag here to avoid duplicate work.

HashEngineering:
@bfoss765 all three items addressed — two were already on the head when your review landed (it appears to have caught the pre-rebase tip):

1. **Double-submission race — fixed in `24b70aac3`.** `deferredPayment` is now `AtomicReference`-backed with a `getAndSet(null)` take in `sendPayment()`, plus a synchronous single-flight `compareAndSet` guard so a duplicate confirm is dropped before any coroutine races (this also closes the pre-existing dashj-path double-tap hazard, where a duplicate would have built a genuinely second transaction). `PaymentProtocolViewModelRaceTest` covers both your scenarios: duplicate-confirm → exactly one submission and no loser rebuild over the winner; retryable failure → guard released, retry submits the rebuilt reservation. Also tidied the detekt `SwallowedException` in the BIP70 test.
2. **Rebase — already done** (`282d7e5c2`, base `bf580425e`): the `SdkL1SendService.kt` constant collision was deduped exactly as you describe (kept the base's `ACCOUNT_TYPE_TAG_STANDARD`/`STANDARD_ACCOUNT_TAG_BIP44`, my block keeps only `ADDRESS_POOL_TAG_EXTERNAL`), and the 29 base-inherited test failures are gone.
3. **The nit — already done** in the same rebase push: the non-routable throw uses `SendNotSdkRoutableException`.

Payments/SDK/send-UI suites green on the current head, including the AAR canary against the v41int11 pin. Ready for re-review.

HashEngineering:
@bfoss765 heads-up before your re-review: the design basis changed after your review, by management directive (final): **replace-then-delete** — when a function moves to the SDK, the dashj implementation is deleted in the same PR; dashj stays only as Phase-3 foundation (#1522), engine retirement is Phase 2 (#1521).

Concretely, `e4f34bd3e` removes the gated dashj BIP70 leg you reviewed (−473/+129): `directPay` now routes through the SDK unconditionally, the ViewModel's dashj dry-run preview and `baseSendRequest` are gone, and the 15 legacy dashj-path tests went with them. So the "gating rather than deleting" property you called out — and the pre-cutover inertness argument — no longer exist by design; the PR description has been rewritten to match, including an explicit list of the regressions accepted under the policy (funding-gate availability on all installs, rollback not covering BIP70, no mempool timeout rescue).

One new piece worth your eyes: because not-yet-cut-over installs still have live dashj-side spenders (manual sends, the background CoinJoin mixer) that can't see the SDK's input reservation, the build now mirrors the reserved outpoints into `Wallet.lockOutput` for the reservation's lifetime — transition-only, tagged in-code for the Phase 2 kill list, with a regression test. Your race-fix items from the earlier review are also in (`24b70aac3`, atomic take + single-flight guard + tests).

HashEngineering:
@bfoss765 heads-up: this merged at head \`437cf5ffa\`, five minutes before I pushed \`4f28e70e2\` — the field-test correction, so it is **not** in \`c3742451b\`. Follow-up opened as #1534 (cherry-picks cleanly).

Two defects in what landed, from the on-device run:

1. 🔴 **Double-pay on retry** — the ViewModel re-arms the preview after an ambiguous transport failure; the engine can't see the mempool spend of the released inputs, so the rebuild picks different inputs. Reproduced on testnet: rebuild \`ed08074b\` vs already-on-chain \`2c3be7d8\` — a retry tap pays the merchant twice. #1534 re-arms only on a definitive nack.
2. **The rescue can never fire** as merged (keys on \`context >= instantSend\`, which the engine never surfaces — it goes 0 → 2), so it just adds ~12s to every failure. #1534 keys on row existence instead; verified working, 1.3s to detect a server broadcast.

Note also that local verification of #1534 is blocked: the merged line pins v41int13, which isn't in local Maven, the pin can't be downgraded (the shielded-invite path needs int13's typed error), and the int13 source isn't pushed anywhere I can fetch. If you can share the AAR or push its branch I'll re-run the suites against it.

PR #1533, opened 2026-08-03
feat(display): restore grouped display of historical CoinJoin mixing transactions

bfoss765:
## What

Restores the pre-removal home-screen UX for wallets holding historical CoinJoin mixing transactions: instead of rendering as dozens of individual "Internal" rows scattered through history, they collapse back into one dedicated **"Mixing Transactions"** group row per day — the same rows the pre-removal UI showed.

**Display-only.** No mixing functionality (engine, service, config) is reintroduced, and the mixed-funds migration logic (`CoinJoinFundsMigrationService`, `MixedFundsMigrationDialogFragment`) is untouched — the CoinJoin account-type tag is only *read* for classification.

## Why they showed as "Internal"

The dashj-era grouping infrastructure survived the CoinJoin removal (`CoinJoinTxWrapperFactory` / `CoinJoinMixingTxSet` are still wired into `TxDisplayCacheService`, and `tx_group_cache` still models groups). The gap is the **post-cutover SDK display pipeline**: `CutoverUiDataService` plans one `tx_display_cache` row per SDK record, and its planner renders both `INTERNAL`- and `COINJOIN`-direction records with the "Internal" title. On a post-cutover restore the held dashj wallet never sees these txs at all, so the surviving wrapper grouping never runs for them — every historical mixing round became its own "Internal" row written by the SDK writer.

## How

In `CutoverUiDataService`'s sync pass (both the Room-snapshot and engine-event feeds — one shared code path):

1. **Classify** a record as historical mixing when:
   - `direction == COINJOIN` (the SDK's Rust classifier affirmatively tagged a mixing round), or
   - `direction == INTERNAL` **and** the txid created at least one TXO on a CoinJoin account (`accounts.accountType == 1`, the `ACCOUNT_TYPE_TAG_COIN_JOIN` tag space, via the same `txos → core_addresses → accounts` join `SdkAssetLockFundingPreflight` uses) — denomination creation / collateral / combine-dust.
   - Exclusions mirror dashj's `CoinJoinTransactionType` rules: a tx that merely *spends* mixed funds keeps its individual row (`Send` parity), and asset-lock-funding records (shield/top-up/invite, which can return change to the CoinJoin account) are carved out via the existing kind resolver.
2. **Collapse** them into per-day group rows using the exact dashj wrapper conventions — rowId `coinjoin_<ISO date>` (same as `CoinJoinMixingTxSet.id`), the existing "Mixing Transactions" string, `ICON_COINJOIN` on the sent background, `FLAG_COINJOIN` (ALL filter only, like pre-removal), value = Σ member nets. Merging is incremental, so a single engine-event record extends an existing day row (memo/rate preserved) instead of rebuilding it.
3. **Persist membership** to `tx_group_cache` as `TYPE_COINJOIN`, so later sync passes, the IS-lock path, and the dashj-side writers all recognize the members as grouped (`isMultiTxGroupRow`), and remove the previously-scattered individual member rows in the same Room transaction (new `TxDisplayCacheDao.upsertGroupRows`).

### Both-writer coexistence (preserve-guard safe)

- Txs already grouped by a dashj rebuild (upgraded installs) are excluded up front via the existing `grouped` check — nothing is double-counted; both writers converge on the *same* per-day rowId.
- Group rows are **never** marked SDK-authoritative and members are never re-planned individually, so `mergePreservingSdkStamped` semantics are unchanged and a dashj rebuild (which has the richer wrapper view) still wins for wallets where dashj holds the txs.
- On upgraded installs, tapping the group row opens the existing `TransactionGroupDetailsFragment` via the group cache as before. On a fresh post-cutover restore (dashj wallet empty) the row renders correctly but the detail sheet can't materialize dashj `Transaction`s — the tap logs and no-ops, same as the current lazy-load fallback. Can be followed up with an SDK-backed group detail if desired.

## Tests

8 new host-JVM tests in `CutoverUiDataServiceTest`: classification rules (COINJOIN direction, CoinJoin-account-funded INTERNAL, spend/send exclusions), dashj id-convention parity, per-local-day collapse and rendering, incremental merge preserving memo/rate, and two service-level end-to-end tests (mixing records → one group row + membership + scattered-row cleanup, while a plain internal move keeps its row; already-grouped txs are never regrouped).

- `:wallet:test_testNet3DebugUnitTest` — 1200 tests, 0 failures
- `:wallet:compile_testNet3DebugKotlin` — clean

### DIFF
--- a/wallet/src/de/schildbach/wallet/database/dao/TxDisplayCacheDao.kt
+++ b/wallet/src/de/schildbach/wallet/database/dao/TxDisplayCacheDao.kt
@@ -91,6 +91,23 @@ interface TxDisplayCacheDao {
     @Query("DELETE FROM tx_display_cache")
     suspend fun deleteAll()
 
+    /** Delete specific rows by rowId. */
+    @Query("DELETE FROM tx_display_cache WHERE rowId IN (:rowIds)")
+    suspend fun deleteByIds(rowIds: List<String>)
+
+    /**
+     * Atomically upsert group rows while removing the individual rows their member
+     * transactions previously rendered as — used when historical CoinJoin mixing
+     * transactions collapse into their per-day "Mixing" group row. One transaction so
+     * the pager never observes the intermediate state (both the group row AND the
+     * individual member rows, or neither).
+     */
+    @Transaction
+    suspend fun upsertGroupRows(rows: List<TxDisplayCacheEntry>, removeRowIds: List<String>) {
+        if (rows.isNotEmpty()) insertAll(rows)
+        if (removeRowIds.isNotEmpty()) deleteByIds(removeRowIds)
+    }
+
     /**
      * Atomically replace the entire cache: delete all existing rows and insert the new ones
      * in a single transaction.  This prevents Room's InvalidationTracker from firing between
--- a/wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverUiDataService.kt
+++ b/wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverUiDataService.kt
@@ -797,6 +797,122 @@ internal fun planL1InstantLockRowUpdate(
     return updated.takeIf { it != existing }
 }
 
+// ── Historical CoinJoin mixing grouping (display-only) ────────────────
+
+/**
+ * The rowId/groupId prefix of a per-day mixing group — the SAME convention the
+ * dashj-era wrapper uses ([de.schildbach.wallet.transactions.coinjoin.CoinJoinMixingTxSet]:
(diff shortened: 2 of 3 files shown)


bfoss765:
On-device verification (S21, build 11.10.46 from this branch): clean install → restore of a historical mixing wallet (~273 mixing rounds). Result: the 22 July day renders as a single **273 transactions / Mixing Transactions / +2.871929** group row with the CoinJoin icon — previously 273 scattered 'Internal' rows on 11.10.45. Non-mixing Sent/Received rows around it stay individual, current-day genuine internals (self-send) stay individual, and restore/identity-recovery were unaffected. Known residue: a handful of individual zero/collateral 'Sent' rows below the group — matches dashj-era behavior (spends deliberately ungrouped); flagging in case product wants them folded in as a follow-up.

PR #1534, opened 2026-08-03
fix(bip70): field-test corrections that missed the #1531 merge — dead rescue + double-pay on retry

HashEngineering:
Follow-up to #1531. This is the **field-test correction commit that missed that merge by five minutes** — #1531 was merged at head `437cf5ffa` (17:35 EDT); this fix was pushed at 17:40. It cherry-picks cleanly onto the merged line.

The on-device test (Pixel 8, testnet, real broadcasts) found two defects in what was merged. One is a funds hazard.

## 1. Double-pay on retry after an ambiguous failure 🔴

`PaymentProtocolViewModel` re-arms the preview after **any** non-acked failure, including an ambiguous transport failure. The SDK engine cannot see the mempool spend of the just-released inputs, so the rebuild selects **different** inputs — a second, independently-valid payment for the same invoice.

Reproduced on testnet, not theorised:
- original `2c3be7d8…` — broadcast by the merchant, later mined;
- post-failure rebuild `ed08074b…` — a *different* tx. A retry tap would have paid the merchant twice.
- Independently confirmed: a fresh preview build for the same invoice after `2c3be7d8` mined produced `ed08074b` again, i.e. that rebuild really was a distinct payment, not a re-serialisation.

**Fix:** re-arm only on a definitive nack (server refused; nothing on the network). An unresolved transport failure leaves the flow un-armed (`canSendPayment == false`) so no blind retry is possible; the user restarts the payment.

## 2. The timeout rescue as merged can never fire

It keys on `TransactionContext >= instantSend`. Measured on the pinned engine: an externally broadcast wallet-relevant tx goes `context 0 (mempool) → 2 (inBlock)` and **never** surfaces an InstantSend context or `isInstantLocked`, even when the network has IS-locked it (`txlock: true` on the explorer). Net effect as merged: the rescue never rescues and adds ~12 s to every failure.

**Fix:** key on the row's **existence**, which is safe here because the engine only records transactions it has observed and a built-but-unbroadcast payment has no row — verified against real txids (`a6cc9ead`, `ed08074b` absent while never sent; server-broadcast `2c3be7d8` present). We have not broadcast at that point in the flow, so an observed tx was broadcast by the BIP70 server. Window widened to ~20 s.

**Verified working after the fix:** server broadcast the payment then returned HTTP 500 → engine observed it at context 0 in **1.3 s** → `completing as paid` → reservation kept → bridged; tx `ed08074b…` on-chain with `txlock: true`.

## Tests

Four arms in `SendCoinsTaskRunnerBIP70Test` / `PaymentProtocolViewModelRaceTest`: observed → paid; never observed → release; nack → re-arm; ambiguous → stays un-armed.

## Verification status — please read

These suites are green **on the pre-merge base** (int11). I could not compile or run them against this branch's merged base locally: it pins `0.1.0-v41int13-SNAPSHOT`, which is not in local Maven, and the pin cannot be downgraded — the merged shielded-invite path requires int13's typed `ShieldedInviteAlreadyClaimed`. The int13 source is not pushed to any branch I can reach on dashpay/bfoss765/HashEngineering, so I can't build it either. Happy to re-run everything the moment the int13 AAR (or its source branch) is available — flagging it rather than implying verification I don't have.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved payment recovery after network timeouts by detecting transactions observed in the mempool or wallet engine.
  * Prevented ambiguous transport failures from automatically rebuilding or retrying payments.
  * Limited automatic retry preparation to definitive payment rejections, reducing duplicate payment attempts.

### DIFF
--- a/wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
+++ b/wallet/src/de/schildbach/wallet/payments/SendCoinsTaskRunner.kt
@@ -331,12 +331,13 @@ class SendCoinsTaskRunner @Inject constructor(
         private val log = LoggerFactory.getLogger(SendCoinsTaskRunner::class.java)
 
         /**
-         * Poll cadence for the BIP70 IS-lock timeout rescue (~12s total)
-         * — the successor of the dashj rescue's 0/1/3/5s loop, widened
-         * because an InstantSend lock takes a moment longer to observe
-         * than raw bloom relay did.
+         * Poll cadence for the BIP70 timeout rescue (~20s total) — the
+         * successor of the dashj rescue's 0/1/3/5s loop. Mempool arrival
+         * was observed within seconds in the field test; the extra
+         * headroom covers relay/persistence jitter on a slow network.
          */
-        private val IS_LOCK_RESCUE_DELAYS_MS = listOf(1_000L, 2_000L, 3_000L, 3_000L, 3_000L)
+        private val BROADCAST_OBSERVE_DELAYS_MS =
+            listOf(1_000L, 2_000L, 3_000L, 3_000L, 3_000L, 4_000L, 4_000L)
     }
     private val paymentIntentParser = DashPaymentIntentParser(org.dash.wallet.common.payments.parsers.AddressNetwork.fromId(NETWORK_PARAMETERS.id))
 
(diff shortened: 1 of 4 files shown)


HashEngineering:
@coderabbitai review

PR #1539, opened 2026-08-14
fix(dash-dex): improve enter amount and other parts of the flow

HashEngineering:
## Summary
- Various Figma-alignment and correctness fixes to the Maya/DEX convert-crypto flow (enter-amount, address input, receive screen, currency picker), plus a handful of unrelated crash/ANR/thread-safety fixes picked up along the way.
- Most recent: align the vertical currency picker in `EnterAmount` with the Figma design (node 38680:47341) — remove the default segmented-picker pill background/shadow, fix rounded corners clipping option text, and add spacing between stacked options.

## Test plan
- [ ] Build and run testnet debug (`./gradlew clean assemble_testNet3Debug -x test`)
- [ ] Verify Maya convert-crypto enter-amount screen: currency picker (DASH/BNB) renders as plain stacked labels, no pill background, no clipped text, visible spacing between options
- [ ] Verify Maya address-input, DEXReceiveScreen match Figma at default and enlarged font scale
- [ ] Sanity-check lock screen PIN text in dark mode, CoinJoin/CrowdNode single-tx group opening, Imgur profile picture upload failure path

## Summary by CodeRabbit

- **New Features**
  - Added improved middle-ellipsis support for long menu subtitles and wallet addresses.
  - Added configurable spacing for segmented picker options.
  - Added new Maya conversion labels and maximum-amount messaging.

- **UI Improvements**
  - Refreshed the Maya receive screen with updated QR layout, deposit information, expiry warning, and address rows.
  - Redesigned the Maya conversion direction cards and balance display.
  - Updated currency picker styling with a cleaner, flatter appearance.
  - Simplified conversion screen titles and improved text overflow handling.


bfoss765 (approved):
Reviewed at head `38086cc94` (review round 1). Compile-verified: `:common` and `:integrations:maya` build clean at this head.

The `MiddleEllipsisText` approach is the right call — width-measured middle truncation for addresses (both ends stay checkable) instead of a fixed character count, and the existing end-ellipsis path is untouched unless a call site opts in. The `SegmentedPicker` padding additions default to zero so every existing picker renders byte-identically. Both `@Preview` device specs matching a measured real S22 is a nice touch.

Approving. Three non-blocking notes, none worth holding the PR for:

1. `middleEllipsizeToFit` trims one character per iteration, measuring each candidate — worst case O(n) `TextMeasurer.measure` calls. Fine for a ~40–60 char address behind a `remember`, but if this ever gets reused for longer strings (memos, URIs), a binary search over the drop count would keep it constant-ish. Maybe worth a one-line comment so a future caller knows the intended domain.

2. `maxAmountErrorMessage`'s no-exchange-rate fallback now returns the bare string "Max" with no amount. Rare path (no fiat rate yet), but as a user-facing error a lone "Max" reads broken — consider keeping the old full-sentence string for that fallback only.

3. i18n nit: `"${getString(R.string.maya_max_amount_error)} $fiatAmount"` concatenates in code, which locks the word order for locales that would phrase it differently. A positional format (`<string name="maya_max_amount_error">Max %s</string>`) would let translators reorder. The pre-existing code had the same pattern, so this is not a regression — just the moment to fix it while the string is new.

HashEngineering:
@coderabbitai

HashEngineering:
@coderabbitai review

PR #1543, opened 2026-08-23
fix: MO-995 | retry the SDK wallet bind and hold the cutover until it succeeds

bfoss765:
## Field failure (MO-995 — Andrei's sync outage, both devices)

During fresh-wallet setup the Kotlin SDK's `createWallet` failed with `UserNotAuthenticatedException` from the Android keystore — a `setUnlockedDeviceRequired` key denied because Keystore2 believed the device was locked (sometimes **falsely**, with the screen demonstrably unlocked). The SDK rolled its wallet back cleanly, but the app then stranded the user with **no sync engine at all**, permanently:

1. `CutoverCoordinator` commits `DUAL_RUNNING -> CUT_OVER` at fresh-wallet setup **before** the bind ever runs (log: `cutover state DUAL_RUNNING -> CUT_OVER (fresh-wallet setup (restore/new))`), so dashj is held (`Phase 5d cutover gate: dashjEngineMayStart=false … dashjHeldByCutover=true`).
2. `SdkWalletBinder` is single-shot per trigger — one failed pass (`SDK wallet binding pass failed; dashj behavior unchanged`) never reruns, while `CutoverUiDataService`'s 5 s wait loop waits for "a SINGLE bound SDK wallet" forever without ever re-invoking the bind.
3. Nothing is surfaced: the Network monitor shows a dead "Not started" / "Network engine not started" (`L1SyncStage.IDLE`).

### 1. Hold-then-rollback on the cutover commit
Deferring the fresh-wallet commit to bind success is not possible in this design: the commit is what routes the fresh-wallet launch (`BlockchainServiceImpl` resolves the engine gate once at service onCreate, right after `setWallet`, while the first bind pass only runs when platform sync starts) — a deferred commit would start dashj on every fresh wallet and then land mid-launch, leaving **both** SPV engines live for the session. So the commit stays immediate and the escape hatch is `CutoverCoordinator.rollbackForFailedBind`: after **5 consecutive failed bind passes** the gate rolls `CUT_OVER` back to `DUAL_RUNNING`, and `BlockchainServiceImpl` — which now observes `CUTOVER_STATE` live (mirroring the existing dashj-diagnostic observer) — un-holds and starts the dashj fallback engine mid-launch. The rollback is skipped while the device is *provably* locked (`KeyguardManager.isDeviceLocked`): a genuinely-locked keystore denial is expected, heals on unlock, and must not flip engines on every locked-screen background start.

**Invariant:** the gate always ends with dashj allowed OR the SDK wallet bound — never both held.

### 2. Re-armable bind
`SdkWalletBinder` now tracks the failed-pass state (`bindRetryPending: StateFlow<Boolean>` + `consecutiveBindFailures`; a throw that leaves no bound wallet arms it, any pass that binds clears it, gate-skipped passes count neither way). The new `SdkBindRetryService`:
- is driven by `CutoverUiDataService`'s existing 5 s bound-wallet wait loop (which runs exactly while the cutover holds dashj and nothing is bound — the stranded state) and re-runs the bind on a capped **5s/15s/30s/60s then hourly** ladder, reset on app foreground;
- registers a runtime `ACTION_USER_PRESENT` receiver (`RECEIVER_NOT_EXPORTED`) that fires an **immediate** bind retry on the next device unlock — the exact heal condition for the keystore false-locked class.

### 3. Surface the failure
While a bind failure is pending retry, the Network monitor renders the new `L1SyncStage.SETUP_RETRYING`: stage "Wallet setup incomplete — retrying", connection row "Unlock your device to finish wallet setup" — instead of the dead "Not started". IDLE-only override (any real scan progress means the wallet bound and the honest stage wins); the dashj regime ignores the flag so post-rollback dashj progress renders normally. No new UI surfaces.

## Test evidence

`./gradlew :wallet:compile_testNet3DebugKotlin` — BUILD SUCCESSFUL.

`./gradlew :wallet:test_testNet3DebugUnitTest` scoped to the touched classes — **235 tests, 0 failures**:
- `SdkBindRetryServiceTest` (new) — 13 tests: the ladder, backoff gating, foreground reset, unlock-receiver arming + immediate heal, rollback threshold, rollback held while provably locked, and an end-to-end replay of the outage over a **real** binder + **real** coordinator asserting the never-both-held invariant (persistent failure ⇒ `dashjEngineMayStart()==true`; heal ⇒ cutover stays committed).
- `CutoverCoordinatorTest` — 26 (3 new: rollback from CUT_OVER, no-op from DUAL_RUNNING, never regresses SETTLED)
- `CutoverUiDataServiceTest` — 80 (1 new: the wait loop consults the retry every poll and stops once bound)
- `SdkWalletBinderTest` — 70 (regression, unchanged)
- `L1SyncStatusServiceTest` — 38 (3 new: SETUP_RETRYING replaces dead IDLE; real progress beats the flag; dashj regime unaffected)
- `NetworkMonitorViewModelTest` — 8 (stage/connection string mappings incl. the new stage)

## Scope note

The SDK-side hardening — a typed keystore error from `createWallet` plus an internal retry — is a deliberately separate follow-up; this PR is the app-side containment that guarantees the user always has a working sync engine and a truthful status readout.

## Summary by CodeRabbit

* **New Features**
  * Added automatic wallet setup retrying with progressively longer retry intervals.
  * Retries resume promptly when the app returns to the foreground or the device is unlocked.
  * Persistent setup failures can safely restore the wallet’s fallback operating mode.

* **Bug Fixes**
  * Network monitoring now clearly reports incomplete wallet setup and active retries.
  * Improved wallet binding recovery during service operation.

* **Tests**
  * Added coverage for retry timing, recovery, rollback protection, and status reporting.

### DIFF
--- a/wallet/res/values/strings.xml
+++ b/wallet/res/values/strings.xml
@@ -199,6 +199,7 @@
     <string name="network_monitor_stage_filters">Scanning block filters</string>
     <string name="network_monitor_stage_synced">Synced</string>
     <string name="network_monitor_stage_error">Sync error</string>
+    <string name="network_monitor_stage_setup_retrying">Wallet setup incomplete — retrying</string>
     <string name="network_monitor_progress">%1$d%%</string>
     <string name="network_monitor_headers_label">Block headers</string>
     <string name="network_monitor_filters_label">Block filters</string>
@@ -210,6 +211,7 @@
     <string name="network_monitor_connection_searching">Searching for peers…</string>
     <string name="network_monitor_connection_idle">Network engine not started</string>
     <string name="network_monitor_connection_error">Connection error</string>
+    <string name="network_monitor_connection_setup_retrying">Unlock your device to finish wallet setup</string>
     <string name="network_monitor_peer_details_note">The wallet engine does not report individual peer connections.</string>
     <string name="network_monitor_dashj_hint">Peer and block lists come from the dashj diagnostic engine. Turn on dashj sync in Tools to view them.</string>
     <string name="block_row_mining_difficulty_adjustment">Mining difficulty adjustment</string>
--- a/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.kt
+++ b/wallet/src/de/schildbach/wallet/service/BlockchainServiceImpl.kt
@@ -207,6 +207,11 @@ class BlockchainServiceImpl : LifecycleService(), BlockchainService {
         override fun onStart(owner: LifecycleOwner) {
             log.info("App moved to foreground")
             isAppInBackground = false
+            // MO-995: a pending SDK bind retry escapes the hourly backoff
+            // tail the moment the user is actually looking at the app.
+            if (::sdkBindRetryService.isInitialized) {
+                sdkBindRetryService.noteAppForeground()
+            }
         }
 
         override fun onStop(owner: LifecycleOwner) {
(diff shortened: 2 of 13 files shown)


HashEngineering:
@coderabbitai review

HashEngineering:
@coderabbitai full review

PR #1544, opened 2026-08-24
fix(crowdnode): detect linked account when API confirmation was self-paid

HashEngineering:
## Problem

After a reinstall or data clear, wallets holding a **linked online CrowdNode account whose 0.00054321 API confirmation was paid from the same wallet** can never restore the account. `restoreStatus()` fails silently on every launch, `signUpStatus` stays `NotStarted`, and since v11.8 (withdraw-only gating) the Explore staking entry is hidden entirely — locking the user out of their CrowdNode balance in the app.

Reported via support ticket: a user with 7.19 DASH on CrowdNode (verified via `IsApiAddressInUse`/`GetBalance`) lost the staking entry after restoring their wallet.

## Root cause

`getApiAddressConfirmationTx()` scans for the confirmation with `CoinsReceivedTxFilter`, whose `isEntirelySelf || getValue < 0` guard (added in a82387e84, Mar 2023, shipped in v10.3.0, to keep CoinJoin self-transfers from matching CrowdNode filters) also rejects the confirmation when the account's registered address lives in the same wallet — every input and output is "mine", so the tx is classified as not-incoming and discarded before the amount check.

This topology is not an edge case of user error: whenever a CrowdNode online account is anchored to an address from the same mobile wallet being linked (i.e. the user's only wallet is the phone), the confirmation is necessarily a self-send. Live linking always worked (the live handler uses `CrowdNodeAPIConfirmationTx`, a guard-free `CoinsToAddressTxFilter`) — only restore-from-blockchain was broken, which is why the failure stayed latent until a data loss.

## Fix

- `CoinsReceivedTxFilter`: new opt-in `includeSelfTransfers` constructor parameter (default `false` — behavior unchanged for all existing users, including `PossibleWelcomeResponse`/`PossibleAcceptTermsResponse`).
- `getApiAddressConfirmationTx()`:
  - passes `includeSelfTransfers = true`; false positives remain impossible because the existing cross-check still requires the matched 54321-duff output to have been **spent to the CrowdNode address** (the forward), which no CoinJoin or coincidental tx does;
  - picks the account address from the output with the exact confirmation amount instead of the first received output, which in a self-paid confirmation could be the change output.

## Regression test

`CrowdNodeBlockchainApiTest` rebuilds the reported wallet state from the **real mainnet transactions** of the affected account (funding → self-paid confirmation `c1858a45…` → forward `6128f9d5…`), with a fake `TransactionBag` controlling address ownership:

- self-paid confirmation is found — **fails on the pre-fix code** (verified), passes with the fix;
- externally-paid confirmation is still found (normal cross-wallet link, passes before and after);
- a confirmation never forwarded to CrowdNode returns null (no false positives).

## Testing notes

QA restore testing in June could not have caught this: API-signup accounts restore via `tryRestoreSignUp()` (unaffected), and link tests confirmed from a second wallet are genuine incoming txs (unaffected). Reproducing requires: link an online account and pay the confirmation **from the same wallet**, then reinstall + restore from phrase. On v11.9.0 the staking entry disappears; with this fix it is restored.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved confirmation detection for self-funded transactions.
  * Correctly identifies wallet-owned outputs matching the confirmed amount.
  * Continues excluding transactions with negative net value.
  * Rejects confirmations that were not forwarded to the service.

* **Tests**
  * Added coverage for self-paid, externally paid, and unforwarded confirmation scenarios.

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/transactions/filters/CoinsReceivedTxFilter.kt
+++ b/common/src/main/java/org/dash/wallet/common/transactions/filters/CoinsReceivedTxFilter.kt
@@ -24,14 +24,15 @@ import org.dash.wallet.common.transactions.TransactionUtils.isEntirelySelf
 
 open class CoinsReceivedTxFilter(
     private val bag: TransactionBag,
-    private val coins: Coin
+    private val coins: Coin,
+    private val includeSelfTransfers: Boolean = false
 ): TransactionFilter {
     var toAddress: Address? = null
         private set
 
     override fun matches(tx: Transaction): Boolean {
         // this check prevents a CoinJoin TX from being marked as a Crowdnode TX
-        if (tx.isEntirelySelf(bag) || tx.getValue(bag).signum() < 0) {
+        if (!includeSelfTransfers && (tx.isEntirelySelf(bag) || tx.getValue(bag).signum() < 0)) {
             // Not an incoming transaction
             return false
         }
--- a/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeBlockchainApi.kt
+++ b/integrations/crowdnode/src/main/java/org/dash/wallet/integrations/crowdnode/api/CrowdNodeBlockchainApi.kt
@@ -27,7 +27,6 @@ import org.dash.wallet.common.services.LeftoverBalanceException
 import org.dash.wallet.common.services.SendPaymentService
 import org.dash.wallet.common.transactions.ByAddressCoinSelector
 import org.dash.wallet.common.transactions.ExactOutputsSelector
-import org.dash.wallet.common.transactions.TransactionUtils
 import org.dash.wallet.common.transactions.filters.CoinsReceivedTxFilter
 import org.dash.wallet.common.transactions.filters.LockedTransaction
 import org.dash.wallet.common.transactions.filters.TxWithinTimePeriod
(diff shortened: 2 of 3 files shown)


thepastaclaw:
⛔ Final review complete — 1 blocking finding(s) (commit f693b3d4) · triage: normal · Phase 2 only (queue backlog)

common/src/main/java/org/dash/wallet/common/transactions/filters/CoinsReceivedTxFilter.kt
```diff
@@ -24,14 +24,15 @@ import org.dash.wallet.common.transactions.TransactionUtils.isEntirelySelf
...
     var toAddress: Address? = null
         private set
 
     override fun matches(tx: Transaction): Boolean {
         // this check prevents a CoinJoin TX from being marked as a Crowdnode TX
-        if (tx.isEntirelySelf(bag) || tx.getValue(bag).signum() < 0) {
+        if (!includeSelfTransfers && (tx.isEntirelySelf(bag) || tx.getValue(bag).signum() < 0)) {
             // Not an incoming transaction
             return false
         }
```
thepastaclaw, line 38: **🔴 Blocking: Preserve negative-value rejection for non-self transactions**

When `includeSelfTransfers` is true, this condition skips both the entirely-self check and the `getValue(bag) < 0` check. As a result, a non-self transaction with negative net wallet value can match if it contains any wallet-owned output worth exactly 54321 duffs. `getApiAddressConfirmationTx()` enables this opt-in during account restoration, so a mixed-input or otherwise net-negative transaction could be treated as the confirmation candidate even though the existing safeguard intentionally excludes negative-value transactions. The self-transfer exception should bypass rejection only for transactions that are actually entirely wallet-owned; negative net value must remain rejected for all other transactions.

```suggestion
        val isSelfTransfer = tx.isEntirelySelf(bag)
        if ((!includeSelfTransfers && isSelfTransfer) ||
            (!isSelfTransfer && tx.getValue(bag).signum() < 0)) {
            // Not an incoming transaction
            return false
        }
```

thepastaclaw (changes requested):

The PR correctly enables recovery of confirmations paid entirely from the restored wallet and adds regression coverage for self-paid, externally paid, and unforwarded confirmations. However, the opt-in filter currently disables the negative-value rejection for every transaction, allowing mixed-input or otherwise net-negative transactions to qualify based solely on an exact-value wallet output.

🔴 1 blocking

<details>

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a contained CrowdNode transaction-detection logic fix with focused regression coverage, but it does not alter consensus, funds movement, cryptography, key handling, network deserialization, or storage migrations.
- Phase 1 reviewers: **not run (skipped for throughput: 11 PRs queued, above the 10 limit)**

</details>

<details>

```

In `common/src/main/java/org/dash/wallet/common/transactions/filters/CoinsReceivedTxFilter.kt`:
- [BLOCKING] common/src/main/java/org/dash/wallet/common/transactions/filters/CoinsReceivedTxFilter.kt:35-38: Preserve negative-value rejection for non-self transactions
  When `includeSelfTransfers` is true, this condition skips both the entirely-self check and the `getValue(bag) < 0` check. As a result, a non-self transaction with negative net wallet value can match if it contains any wallet-owned output worth exactly 54321 duffs. `getApiAddressConfirmationTx()` enables this opt-in during account restoration, so a mixed-input or otherwise net-negative transaction could be treated as the confirmation candidate even though the existing safeguard intentionally excludes negative-value transactions. The self-transfer exception should bypass rejection only for transactions that are actually entirely wallet-owned; negative net value must remain rejected for all other transactions.
```
</details>

PR #1548, opened 2026-08-26
fix: classify a stale funding reservation as NotBroadcast on the atomic send path

HashEngineering:
## What

Adds a typed `StaleReservationToken` arm to `classifyCoreSendFailure`, so a stale funding reservation on the **atomic** send path classifies `NotBroadcast` instead of `Ambiguous`.

## Why

[dashpay/platform#4309](https://github.com/dashpay/platform/pull/4309) age-guards the finalized-transaction handle path (`finalize` → `broadcastTransaction`) against the same reservation bound the deferred BIP70 surface already used, and deliberately **reuses native code 34** (`ErrorStaleReservationToken`) rather than minting a new one.

Until then, code 34 could only arrive from the deferred surface, so the arm lived only in `classifyDeferredBroadcastFailure`. After #4309 it also reaches:

- `SdkL1SendService.sendToAddress` → `ManagedPlatformWallet.sendToAddresses` (the plain send)
- `CoreSendAllNative` → `ManagedCoreWallet.broadcastTransaction` (send-all / CoinJoin drain)

Neither is covered by the deferred classifier, so the error fell through `classifyCoreSendFailure`'s message-prefix rules to `classifyBroadcastFailure` and came back **`Ambiguous`**.

That is wrong in the direction that costs the user most. The SDK contract guarantees this refusal happens *before* the broadcaster is touched, and the guard releases the still-owned reservation owner-guarded on the way out, so the inputs are immediately reselectable. Classifying it `Ambiguous` shows a payment that provably never left the device as "may be on the network" **and** blocks the safe dashj fallback — the exact failure mode the existing typed arms (`CoreInsufficientFunds`, `SigningKeyUnavailable`, `TransactionBroadcastRejected`) were added to prevent.

## Changes

- `SdkL1SendService.kt` — typed `StaleReservationToken` arm in `classifyCoreSendFailure`, placed with the other definitively-pre-network arms, plus KDoc explaining why both classifiers keep an arm (the deferred one stays ahead only to name the deferred surface in the reason string).
- `SdkL1SendServiceTest.kt` — `classify_staleReservationToken_isNotBroadcast_onTheAtomicSendPath`, and code 34 added to the deferred-table parity list.

## Testing

`:wallet:testProdDebugUnitTest --tests SdkL1SendServiceTest` — **71 tests, 0 failures.**

Verified non-vacuous: with the production arm reverted, the new test fails (`AssertionError`) and it is the only failure. The existing deferred-path test still passes without it, confirming the two classifiers are independently covered.

The new test also asserts the ambiguous sibling `TransactionBroadcastUnconfirmed` (code 20) keeps its `Ambiguous` classification, so the pair cannot collapse into one rule.

## Notes

- Safe to land **before** #4309. Against the current AAR the arm is unreachable, so this is a no-op until the native guard ships.
- A second gap from #4309 is **not** addressed here and needs a platform-side change first: `PlatformWalletError::InputMidBroadcast` (the new build refusal when a coin selection picks an input held by an in-flight dispatch) maps to `ErrorUnknown` (99) and so also classifies `Ambiguous`. Fixing it by message-prefix matching would be fragile — the message embeds an outpoint — so the right fix is claiming a real FFI code upstream and adding a typed arm here. Not yet raised upstream.

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/platform/sdk/SdkL1SendService.kt
+++ b/wallet/src/de/schildbach/wallet/service/platform/sdk/SdkL1SendService.kt
@@ -108,6 +108,16 @@ internal const val ADDRESS_POOL_TAG_EXTERNAL = 0
  * network", blocking the safe dashj fallback). The message-prefix rules
  * below stay for the transition window.
  *
+ * [DashSdkError.PlatformWallet.StaleReservationToken] (code 34) joins them
+ * for the same reason (dashpay/platform#4309). It used to reach only the
+ * DEFERRED surface, so its arm lived only in
+ * [classifyDeferredBroadcastFailure]; #4309 age-guards the atomic
+ * finalize → broadcast path against the same reservation bound and reuses
+ * the same code, so it now reaches the plain send and the drain as well.
+ * Both classifiers keep an arm: this one is the general case, and the
+ * deferred one stays ahead of it only to name the deferred surface in the
+ * reason string.
+ *
  * Everything else defers to [classifyBroadcastFailure]; notably
  * [DashSdkError.PlatformWallet.TransactionBroadcastUnconfirmed] (code 20 —
  * the SDK's own "may already be on the network, inputs stay reserved, do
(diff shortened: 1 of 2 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 390630ac) · triage: low · Phase 2 only (queue backlog)

thepastaclaw (approved):

The PR correctly adds a typed StaleReservationToken arm to classifyCoreSendFailure, ensuring stale funding reservations on atomic send paths are classified as definitively pre-network rather than Ambiguous. The accompanying test verifies both the new NotBroadcast behavior and preservation of Ambiguous classification for TransactionBroadcastUnconfirmed; no in-scope correctness, security, or regression issues were found.


<details>

Source: reviewer 1: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `low` by `gpt-6-astra` (effort low) — This is a small, well-contained classifier change with focused unit-test coverage, and it does not directly alter funds movement, transaction construction, signing, or broadcast behavior.
- Phase 1 reviewers: **not run (skipped for throughput: 12 PRs queued, above the 10 limit)**
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort medium); agent `phase2-reviewer`

</details>

PR #1554, opened 2026-08-29
feat(dashpay): add contact via QR code (My QR / Scan QR)

PastaPastaPasta:
## What

Ports the iOS "add contact via QR" feature to Android: a user can show a QR code identifying their DashPay identity, and scan another user's code to land on the send-contact-request sheet.

- **Wire contract parser** — `DashPayUserLink` implements the shared `dashpay://user?id=<base58-identity-id>&username=<label>` payload, mirroring iOS `DashPayUserLink.swift` exactly (case-insensitive scheme/host/param names; rejects userinfo/port/path/fragment/duplicate/unknown params; `id` must decode to 32 bytes; trailing `.dash` tolerated and stripped). `DashPayUserLinkTest` ports the iOS acceptance/rejection vectors byte-for-byte — these tests are the cross-platform contract and should stay in sync with `DashPayUserLinkTests.swift`.
- **UI entry points** — a "My QR" / "Scan QR" button row under the search field on the Add a New Contact screen (both ConstraintSet layout variants). "My QR" shows only with a registered identity + username; "Scan QR" whenever an identity exists.
- **My QR sheet** — Compose bottom sheet with display name, username, the QR on a forced-white card (deliberately white in dark mode too, for camera scanner contrast — same as iOS), and a caption.
- **Scan → verify → sheet** — scanned payloads are parsed strictly (payment URIs / invitation links get a targeted "isn't a DashPay user QR" dialog rather than falling through to `InputParser`); own-QR shows a notice; known contacts skip straight to the sheet; everyone else is verified against Platform via `PlatformRepo.updateDashPayProfile()` — only Platform-resolved profile/DPNS data is displayed, with the QR's `username` param used solely as a mismatch check. Loading state + retryable error dialog on failure. On success the existing `DashPayUserBottomSheet` handles send-request and refresh.
- Strings for Transifex, `AnalyticsConstants.UsersContacts` events for show-my-QR / scan / scan-success.

Deferred (as planned): the OS-level `dashpay://user` manifest deep link — iOS only ever emits this URI inside a QR, and in-app scanning covers the feature.

Also includes a standalone commit repairing the master build (current master HEAD fails CI): a duplicate `Configuration` import in `MenuItem.kt` and `subtitleMaxLines` arguments passed to `ActionItem`, which has no such parameter (its subtitle is already single-line).

## Verification

- `:wallet:compile_testNet3DebugKotlin` passes
- `DashPayUserLinkTest` 6/6 passes (`:wallet:test_testNet3DebugUnitTest`)
- `ktlintCheck` passes

## QA notes

Suggested two-device testnet smoke:
- Android shows / iOS scans, and iOS shows / Android scans
- Scan your own QR (expect "This is your QR code" notice)
- Scan a payment QR and an invitation QR from this screen (expect the targeted error, not the payment flow)
- Scan a QR whose username no longer matches the identity's DPNS name, and a testnet QR on mainnet (both expect the "couldn't be verified" retry dialog)
- Scan an existing contact (expect the sheet in established-contact state, no Platform round-trip)


HashEngineering:
This was tested and it works.

thepastaclaw:
🕓 Review not started yet because this PR is a draft.

- [ ] **Request normal review** — click when the PR is ready for review.
- [ ] **Request priority review** — click to move this review to the front of the queue.

_Commit 86661e3d. Normal review starts when eligible; priority review starts as soon as a slot is available._

PR #1555, opened 2026-09-01
fix: restore L1 sync after service teardown, and never leave the wallet without an engine

HashEngineering:
Fixes the MO-995 sync failures reported on the 12.0.0 QA builds, plus several defects found while retesting them. Base is `feat/kotlin-sdk-phase-1` (36 commits — see the second-round section below).

## The reported bug

**No L1 sync after a memory trim** — `b994bb6a1`. `platformSyncService.resume()`, the only in-process path that restarts the SDK's L1 engine, sat below the `!dashjEngineMayStart` early return in `checkService()`. Post-cutover that return always fires, so the engine only ever started once per process from `WalletApplication`. Any teardown that stopped it — `onTrimMemory`, the idle detector, the Android 15 FGS timeout — killed L1 sync permanently. Andrei's mainnet log: engine stopped 13:37:38, service recreated five times, **2h10m with no sync**, two Coinbase deposits invisible. Only reproduces on release builds, because `shutdown()` deliberately keeps the engines warm on debug.

**Crash tapping the fiat option on the Coinbase transfer screen** — `ce0bc1551`. `setSpan(12 … 7)`: the span was bounded by `inputValue.length`, but `fiatBalance` is the 2-dp formatted figure and is shorter whenever the input has more than two decimals.

## Found while retesting — walletB had *no* L1 engine at all

A HONOR PTP-N49 upgraded, ran fine on dashj all morning, then at 18:20 lost sync entirely. Root cause chain: an Android Keystore denial on the lock-bound master alias (MO-1022) meant the SDK could not bind — and the cutover committed anyway, holding dashj and leaving zero engines.

- `5cbffa897` / `2f574b8a1` — no path into `CUT_OVER` without durable SDK bind evidence. Both write sites (`commitLocked` **and** `transition`) now consult it, because the readiness-driven auto-commit was bypassing the seam gate entirely — reproduced committing four times with every bind failing.
- `93795401b` — the bind marker is written on every successful bind, not only a newly created one.
- `1b25d6e45` — an app-foreground visit now *drives* a bind retry instead of only resetting a backoff that nothing read.
- `984ca67ee` / `5d89950bc` — **`ProcessLifecycleOwner` has never fired in this app.** `AndroidManifest.xml` removes `androidx.startup.InitializationProvider`, so `lifecycle-process`'s initializer never runs. Across a 27,000-line log with 36 service `onCreate`s, "App moved to foreground" appears **zero** times — so `isAppInBackground` was frozen and the foreground hook was dead code. Signal now comes from `WalletActivityTracker`, keeping the cold-start optimisation.

## Other defects found

- `ad66c6718` — **Uphold portal crashes on release builds.** `integrations/uphold` was the only module minifying its own library output; R8 strips `hilt_aggregated_deps/` (5 entries in debug, 0 in release), so the app component never implements `UpholdPortalFragment_GeneratedInjector`. Present on master too, so likely shipping broken.
- `091c2bae3` — `onStoppedLast()` never ran (override missing `super`), so auto-logout was never told the app backgrounded.
- `c41b7a134` — diagnostic for the CoinJoin mislabel, plus a latent `ClassCastException` from `tx.raw as Transaction` on an `Any?` documented as opaque.

## Verification

Emulator harness (`scripts/cutover-emulator-test.sh`), release build, real Keystore denials:

| scenario | result |
|---|---|
| S1 upgrade, working bind | 5/5 |
| S2 upgrade, denied bind | 5/5 — declines, dashj live, never engine-less |
| S3 in-session recovery | 7/7 — heals on foreground, same process |
| S3b next launch commits | 2/2 |
| S4 trim → SPV restart | pass |

Uphold fix measured on the generated component: `_testNet3Release` uphold injectors **0 → 2**.

## Not fixed / known

- **The Keystore denial itself (MO-1022) is untouched** — an OEM defect (Google Issue Tracker 506989112). Of walletB's 16 denials, 9 were genuinely-locked and heal on a foreground retry; **7 were FALSE-LOCKED** and cannot. Those devices run on dashj and never activate the SDK.
- **This blocks dashj removal.** Every gate here degrades to "keep dashj". Once dashj is gone that has no second half, so MO-992/MO-993 need the SDK keystore fix first, and `SDK_BIND_EVER_SUCCEEDED` is the natural per-install gate for retiring dashj.
- FALSE-LOCKED cannot be reproduced on an emulator; that branch is unit-tested only.
- Real upgrade mechanics (Room migrations, protobuf) untested — only gate logic.
- `e4993d240` was already on the branch and is included here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transfer-screen crashes caused by invalid currency positions.
  * Corrected received-coin notifications for self-authored sends and non-positive amounts.
  * Improved CoinJoin handling for unsupported transaction data.
  * Prevented release-build networking crashes and premature service shutdowns.
  * Preserved accurate synchronization progress during temporary engine interruptions.

* **Improvements**
  * Foregrounding the app now actively retries wallet connection and recovery.
  * Improved wallet upgrade safeguards, cutover reliability, and service restarts.
  * Enhanced diagnostics and error messages for identity registration and transaction failures.

* **Testing**
  * Added regression coverage across transfers, notifications, upgrades, synchronization, retries, and CoinJoin handling.

---

# Second round — defects found retesting 12000001 → 12000006

The 20 commits after `ad66c6718`. Every one came from a QA field log; three of them are regressions from earlier commits **in this PR**, called out as such.

## MO-973 — username creation crash-looped on release builds

One crash explains three separate QA reports: the crash, the "cannot create a username" failure, and "the lock screen appears after clicking Continue".

`NettyChannelBuilder`'s static initializer wires its transport entirely by **reflection**, so R8 sees no reference to the members it needs and strips them. Each strip becomes `ExceptionInInitializerError` the first time anything constructs a `DAPIGrpcMasternode`:

1. `accbd2997` — `Utils.isEpollAvailable()` does `getDeclaredMethod("isAvailable")`; class kept, method stripped, so the probe threw instead of returning false (12000005).
2. `e9f178e15` — with epoll then correctly false, `<clinit>` falls through to `new ReflectiveChannelFactory<>(NioSocketChannel.class)` on the **next** line, whose no-arg constructor had been stripped (12000006). Targeted rules only reveal the next stripped member, so the whole shaded package is now kept: 26,542 methods / 1.9 MB of dex against a 233 MB APK. Verified by building the release variant and reading the dex back with `apkanalyzer`, not by reasoning about R8.
3. `962d4de5d` — `PlatformHealthProbe.probe()` caught only `Exception`, and this fails with an **Error**. It sailed past that catch, past the caller's bare `try/finally`, out of the coroutine, and killed the process — 11 times on one HONOR PTP-N49. An advisory row that feeds one warning must never be able to do that. `CancellationException` is rethrown so structured concurrency still works, and a non-Exception logs at ERROR **with the full stack** — degrading silently would trade a loud crash for an invisible one.

The lock screen was Android relaunching the killed process: seven times, `CreateUsernameActivity created` → crash 3–4 s later → `WalletApplication.onCreate()` in the same second → `show lock screen true`.

Netty is present only because the legacy Java dapi-client uses it for transport; the Kotlin SDK does not. It stays reachable from `TopUpRepository.getTransaction`, `IdentityRepository`'s credit balance and `PlatformSyncService.reportNetworkStatus`, which is why this is a keep rule rather than a fix to the probe alone.

## MO-995 — the L1 engine died whenever the user left the app

`0b31e7980`. `onTrimMemory` used `level >= TRIM_MEMORY_BACKGROUND`, which is wrong in **both** directions. 40 does not mean low memory — it means the process went onto the LRU background list, i.e. the user left the app. And the signals that do mean running-low, `RUNNING_LOW` (10) and `RUNNING_CRITICAL` (15), are numerically *below* 40 and were ignored entirely. The scale is not ordered by severity, so `>=` cannot express the policy at all; the log line called it "low memory detected" throughout.

Field log, HONOR PTP-N49 (384/512 MB heap) mid from-genesis scan: three engine restarts in five minutes, `scanCaughtUpToTip` never held, the cutover never committed, and Buy Credits failed for want of an SDK-owned L1. Now stops only on `TRIM_MEMORY_COMPLETE` (80) and `RUNNING_CRITICAL` (15).

## Regressions from this PR's own earlier commits

- `21748def0` — `2f574b8a1` put the bind-evidence guard inside `commitLocked`, which the **fresh-wallet** commit shares with the upgrade seam. A new install has no bind evidence by construction: the commit is what routes the launch, the bind runs after it. So every newly created or restored wallet silently fell back to dashj and the SDK's fast initial sync never ran. The bind succeeding three seconds *after* the refusal is the whole bug in one log line. `commitLocked` now takes `requireBindEvidence` explicitly — true for the seam, false for fresh setup, where `rollbackForFailedBind` was always the designed escape hatch.
- `497dec7df` — the explainer was never armed on a real upgrade. The seam cannot commit on the upgrade launch, and the path that *does* commit (the readiness auto-commit in `transition`) had no arming logic. Arming now belongs to the commit, gated on the durable boundary latch. This also supersedes the framing of `cdb17f3a2`, which fixed a double-arm on a path that in the normal flow never fires.
- `4f74977a2` — `fba24cbfb`'s filter backstop was held **in memory**, so it was empty on a cold start, which is exactly when the Network Monitor gets opened. Now floors on the engine's committed wallet cursor, seeded from the SDK's durable watermark. ChainLock deliberately left alone: the provider already max-guards it and QA's own earlier screenshot shows it populated while only filters read "—".

## Also

- `5d5e59d92` — `startForeground()` sat inside `onStartCommand`'s coroutine behind `onCreateCompleted.await()`, conflating the framework's `onCreate` (which has returned by then) with this service's async init (which the FGS deadline does not wait for). An alarm-triggered start raced the idle detector's shutdown and the process was killed with `ForegroundServiceDidNotStartInTimeException`. Now synchronous.
- `ea513d474` — `startIfEnabled`/`stop` tear down four long-lived loops and nothing recorded how long the engine was down; reconstructing a 5h28m outage meant diffing timestamps 12,000 log lines apart. Reports the gap on the start that ends it — no timer needed, and it outlives the dashj retirement, where these become *the* engine's loops.
- `fddb0e244` — the emulator harness printed FAIL and exited 0, and kept a second copy of the cutover boundary that disagreed with the app's.

## Review follow-ups

`82969ccc6` — two CodeRabbit findings, both confirmed against the code.

- **The upgrade explainer could be lost to a single dropped write.** The cutover boundary crossing is computable on exactly one launch (`Configuration.lastVersionCode` is overwritten at every startup), so when GATE 1's `set(CUTOVER_UPGRADE_BOUNDARY_CROSSED)` threw, the crossing existed only as a local `val`. `armUpgradeNoticeIfUpgraded` reads the *store*, not that value, so it saw `false` and returned — and every later launch computed `crossedNow == false` with nothing on disk to recover from. The commit still happened, so the symptom is MO-1022's (no one-time sync explanation) reached by a different route. Now held in memory for the launch that observed it, with the persist retried at commit time; if the store never accepts the key, arming proceeds from memory, which is all that can be salvaged — the explainer's own flags need that same store.
- **`filterTarget` could trail `filterHeight`.** The IDLE/CONNECTING target took every floor the height took *except* the height's own raw value, so an inverted engine snapshot rendered as `2533400/2533349` — a row past 100%. Minor, and the non-idle branch passes such a snapshot through unchanged regardless, but the comment directly above asserted the pair "can never render as height > target", which was not true. Worth noting the review's stated mechanism is wrong: both fields are read off the same `data.filters` snapshot in `toShadowSyncProgress`, so they cannot drift apart — this needs an already-inverted snapshot, not a partial update. Comment corrected to say what actually holds.

Three regression tests, each verified to fail against the unfixed code.

`517f201bd` — KDoc for `armUpgradeNoticeOnce`, the one function this PR adds without one. CodeRabbit's docstring-coverage check flags 54% against an 80% threshold, but it scores every function *touched* by the diff; the files changed here sit at ~57% against a 28% baseline for `wallet/src` on the base branch, so the gap is pre-existing code this PR barely edits, not new code.

`39d188d94` — S4 in the emulator harness was left behind by two commits in this same PR and would have reported FAIL against correct code: it fired `TRIM_MEMORY_BACKGROUND`, which `0b31e7980` deliberately stopped treating as pressure, and grepped for `low memory detected, stopping service` and `L1 shadow sync stopped` — renamed to `memory pressure (onTrimMemory level N), stopping service` and `L1ShadowLifecycle STOPPED`. S4 now drives both halves of the policy: BACKGROUND must be ignored (refuted explicitly, since a regression to `level >= TRIM_MEMORY_BACKGROUND` would otherwise still pass on the COMPLETE leg alone), and COMPLETE must stop the service. Found by thepastaclaw. `45c83c8e7` then pinned that leg's assertion to level 80: `refute_log` does not advance `LOG_MARK`, so a level-40 teardown arriving just after the refute sampled would have satisfied the COMPLETE assertion meant to catch exactly that regression. Re-marking the log would also close it but would drop the cold-launch start out of the window and break the `>= 2` restart count. Found by CodeRabbit.

`21ce82628` — the boundary-latch retry in `82969ccc6` was on a path the upgrade launch never takes. `armUpgradeNoticeIfUpgraded` is reached only via a successful commit, but on a real upgrade the seam declines (the bind runs after it), so the retry never ran on the one launch that can observe the crossing — the defect was moved, not closed. `persistBoundaryCrossingIfPending()` now runs on the decline path and on every readiness evaluation. Same review also fixed the `log` diagnostic helper's stale filters and made the `CUTOVER_UPGRADE_BOUNDARY_CROSSED` KDoc refer to `FIRST_CUTOVER_VERSION_CODE` instead of a release name. Found by thepastaclaw (Phase 1 + Phase 2).

`215fd6691` — reattaches three KDoc blocks this branch had separated from the members they document: inserting a new member directly after an existing KDoc leaves that KDoc describing the newcomer. `armUpgradeNoticeIfUpgraded` lost its ~40-line KDoc to `persistBoundaryCrossingIfPending` one commit earlier; `L1SyncStatusService.details` and `L1ShadowSyncService.logCompletion` had the same shape; `commitLocked`'s description and `@param` were split into two adjacent blocks. Two orphaned KDocs that predate the branch are left alone, since reattaching them means guessing the original author's intent. Note this does NOT clear CodeRabbit's 80% docstring threshold — coverage on changed files moves 40.2% to 41.5%, and the remainder is pre-existing functions in files this PR barely touches.

`c77b5e86e` — **rotating a screen could lock the wallet**, a regression this PR introduced. `091c2bae3` re-enabled `onStoppedLast()`, dead code since before the class existed (the override never calls `super`, so `ActivitiesTracker.numStarted` never advanced). Waking it woke a hazard that could not bite while it was dead: Android stops the outgoing activity before starting its replacement on a configuration change, so `visibleActivityCount` touches zero during a rotation and ran the "app went away" path — `setAppWentBackground(true)`, which `AutoLogout.shouldLogout()` returns outright on a zero-minute timeout, plus a `FORCE_FINISH_ACTION` broadcast. Nine activities neither pin portrait nor handle orientation (`MainActivity`, `AddressBookActivity`, `NetworkMonitorActivity`, `BlockInfoActivity` among them), so rotating any of them dropped the user on the lock screen without leaving the wallet. Guarded on `isChangingConfigurations()`, covering `AppForegroundMonitor.noteBackground()` too. Three lifecycle tests, mutation-verified. Found by thepastaclaw.

## Not fixed here

- The HONOR cutover is still blocked by a legacy identity parked at `USERNAME_REGISTERING`, which makes `IDENTITY_OPERATION_IN_FLIGHT` permanent. Not a keystore problem — the keystore fix is verified working on that device.
- The cutover still commits on a launch *after* the upgrade, so the explainer can appear late (MO-1022 comment 91177).
- `ProbeWatchdogDecider` allows one parity-loop restart per **process** while the shadow restarts several times per process.
- The `stopSelf()`/`onStartCommand` race (survivable now, but `stopSelfResult(startId)` is the answer), and the ~4 h from-genesis scan when `birthHeight` clamps to 0.

## Testing

`:wallet` and `:common` unit suites green. Eight test files added on the branch — `PlatformSyncEngineRestartTest`, `TrimMemoryPolicyTest`, `ForegroundStartPromiseTest`, `PlatformHealthProbeContainmentTest`, `CoinsReceivedNotificationGateTest`, `CoinJoinMixingTxSetTest`, `AmountCurrencySpanTest`, plus `scripts/cutover-emulator-test.sh` — and six existing suites extended: `CutoverCoordinatorTest`, `SdkBindRetryServiceTest`, `L1SyncStatusServiceTest`, `L1ShadowSyncServiceTest`, `SdkDashPayWritesTest`, `InviteCreationFailureTest`.

```bash
./gradlew :wallet:test_testNet3DebugUnitTest :common:test
```

Every fix in this round is mutation-verified: the change is reverted and the suite must fail the specific pinning test. Two caveats worth stating plainly — the R8 keep rule has **no** unit test, because R8 does not run for the JVM test variant, so the release-build dex read is the only check; and `PlatformHealthProbeContainmentTest` forces `Constants.SUPPORTS_PLATFORM` true, without which `probe()` returns on its first line and every assertion passes vacuously.

Device-verified on the emulator against real Keystore denials (release build), and the HONOR keystore fix confirmed on hardware: 23 denials and no successful bind before, zero denials and a bind in ~1 s after. Still unproven there: the **decrypt** path on an existing lock-bound alias.

### DIFF
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -218,8 +218,24 @@ require_state() {
 }
 
 lock_screen()   { adbs input keyevent 26; sleep 2; }   # power -> screen off, keyguard on
-wake_unlock()   { adbs input keyevent 26; sleep 1; adbs input keyevent 82; sleep 1
-                  adbs input text "$PIN"; adbs input keyevent 66; sleep 3; }
+# WAKE ONLY — never `keyevent 26` (power), which TOGGLES: on an already-unlocked
+# device it turns the screen OFF and re-locks it, which silently re-created the
+# Keystore device-locked denial in the middle of a scenario that needs a working
+# bind. Also note `input text` does NOT reach the keyguard PIN pad (digits need
+# keyevents 8-11); scripted PIN entry proved unreliable, so an actually-locked
+# device must be unlocked by hand before running the unlocked scenarios.
+# Keep the screen on for the whole run: the scenarios poll for up to 45s and a
+# screen-off re-locks the device, which re-creates the Keystore denial in the
+# middle of a scenario that needs a working bind.
+keep_awake()    { adbs svc power stayon true >/dev/null 2>&1
+                  adbs settings put system screen_off_timeout 1800000 >/dev/null 2>&1; }
+
+wake_unlock()   { keep_awake; adbs input keyevent 224; sleep 2
+                  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
+                    note "WARNING: device still shows a keyguard — unlock it by hand, then re-run."
+                    note "         a locked device denies the master-alias keystore op and this"
+                    note "         scenario needs a WORKING bind."
+                  fi; }
 launch_app()    { adbs monkey -p "$PKG" -c android.intent.category.LAUNCHER 1 >/dev/null 2>&1; sleep 12; }
 start_service() { adbs am start-foreground-service -n "$SVC" >/dev/null 2>&1; sleep 12; }
 
@@ -326,6 +342,8 @@ s3)
   say "S3 — denied, then healed (recovery path; expected to expose the noteAppForeground gap)"
   require_device; require_root
   note "assumes S2 just ran: state DUAL_RUNNING, bind marker unset"
+  show_state
+  mark_log          # S3 had NO mark, so its assertions matched the whole log history
   wake_unlock
   launch_app
   assert_log "bind succeeded after unlock"       "app wallet (bound to new|already bound to) SDK wallet"
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -36,7 +36,7 @@ set -uo pipefail
 PKG=hashengineering.darkcoin.wallet_test
 SVC=$PKG/de.schildbach.wallet.service.BlockchainServiceImpl
 AVD=Pixel_9_API_36_AOPS
-PIN=1234
+PIN=REDACTED
 PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
 BT=~/Library/Android/sdk/build-tools/35.0.1
 APK_DIR=wallet/build/outputs/apk/_testNet3/release
@@ -230,12 +230,37 @@ lock_screen()   { adbs input keyevent 26; sleep 2; }   # power -> screen off, ke
 keep_awake()    { adbs svc power stayon true >/dev/null 2>&1
                   adbs settings put system screen_off_timeout 1800000 >/dev/null 2>&1; }
 
-wake_unlock()   { keep_awake; adbs input keyevent 224; sleep 2
-                  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
-                    note "WARNING: device still shows a keyguard — unlock it by hand, then re-run."
-                    note "         a locked device denies the master-alias keystore op and this"
-                    note "         scenario needs a WORKING bind."
-                  fi; }
+# Enter $PIN on the keyguard using DIGIT KEYEVENTS. `input text` does not reach
+# the keyguard PIN pad — that is why earlier scripted unlocks silently failed and
+# left the device locked, which then denied the master-alias keystore op inside
+# scenarios that needed a working bind. KEYCODE_0 is 7, so digit d -> 7+d.
+pin_keyevents() {
+  local i ch
+  for (( i=0; i<${#PIN}; i++ )); do
+    ch=${PIN:$i:1}
+    adbs input keyevent $((7 + ch))
+  done
+  adbs input keyevent 66   # ENTER
+}
+
+# Wake and, if a keyguard is up, dismiss it. Never `keyevent 26` (power), which
+# TOGGLES and would turn the screen OFF on an already-unlocked device.
+wake_unlock() {
+  keep_awake
+  adbs input keyevent 224; sleep 2
+  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
+    adbs input swipe 540 2000 540 600 300; sleep 2
+    pin_keyevents; sleep 4
+  fi
+  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
+    note "WARNING: keyguard still up after entering \$PIN — unlock by hand, then re-run."
+    note "         a locked device denies the master-alias keystore op and this"
+    note "         scenario needs a WORKING bind."
+  else
+    note "device unlocked"
+  fi
+}
+
 launch_app()    { adbs monkey -p "$PKG" -c android.intent.category.LAUNCHER 1 >/dev/null 2>&1; sleep 12; }
 start_service() { adbs am start-foreground-service -n "$SVC" >/dev/null 2>&1; sleep 12; }
 
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -41,11 +41,47 @@ AVD=Pixel_9_API_36_AOPS
 # NEVER hardcode it — supply it per run:  EMULATOR_PIN=... ./this-script s3
 # Only the scenarios that lock the screen need it (s2, and s3's unlock).
 PIN="${EMULATOR_PIN:-}"
-PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
+# MUST match CutoverCoordinator.FIRST_CUTOVER_VERSION_CODE. The cutover ships
+# in 12.0.0, so the boundary is 12000000 — it MOVED here from 11100000 when the
+# release slipped a line. Two copies of a boundary is one copy too many: this
+# used to be hardcoded again below, where a device last run on 11.10–11.25 (which
+# the APP counts as pre-cutover) was judged post-cutover by the script, so the
+# script overwrote a genuine last_version instead of leaving it alone.
+FIRST_CUTOVER_VERSION_CODE=12000000
+PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE
 BT=~/Library/Android/sdk/build-tools/35.0.1
 APK_DIR=wallet/build/outputs/apk/_testNet3/release
 OUT=/tmp/mo995-emulator
 mkdir -p "$OUT"
+
+# Assertion ledger. Every FAIL bumps this; the EXIT trap turns a non-zero
+# count into a non-zero exit status.
+#
+# WHY NOT `exit 1` AT THE FAILING ASSERTION: each scenario is a SEQUENCE of
+# assertions and the diagnostic value is in seeing all of them. S3 asserting
+# "bind failed" then "dashj came back" then "state not committed" tells you
+# WHERE the chain broke; aborting at the first FAIL reports only that it broke.
+# So assertions still print and continue — but the script no longer exits 0
+# after printing FAIL, which would let an automated run call a failed cutover
+# scenario a success.
+FAILURES=0
+
+fail() {
+  FAILURES=$((FAILURES + 1))
+  printf '   \033[31mFAIL\033[0m %s\n' "$1"
+  [ $# -gt 1 ] && printf '        %s\n' "$2"
+  return 0
+}
+
+on_exit() {
+  local rc=$?
+  if [ "$FAILURES" -gt 0 ]; then
+    printf '\n\033[31m%s assertion(s) FAILED\033[0m\n' "$FAILURES"
+    exit 1
+  fi
+  exit "$rc"
+}
+trap on_exit EXIT
 
 say()  { printf '\n\033[1m== %s\033[0m\n' "$*"; }
 note() { printf '   %s\n' "$*"; }
@@ -129,7 +165,7 @@ assert_log() {
     fi
     sleep 5; waited=$((waited + 5))
   done
-  printf '   \033[31mFAIL\033[0m %s\n        (no match in %ss for: %s)\n' "$label" "$budget" "$pat"
+  fail "$label" "(no match in ${budget}s for: $pat)"
 }
 
 # refute_log <label> <grep-pattern>   — PASS if absent from the new lines
@@ -137,7 +173,7 @@ refute_log() {
   local label="$1" pat="$2"
   local w; w=$(since_mark)
   if grep -qE "$pat" "$w"; then
-    printf '   \033[31mFAIL\033[0m %s\n        (unexpected: %s)\n' "$label" "$pat"
+    fail "$label" "(unexpected: $pat)"
     grep -E "$pat" "$w" | head -2 | sed 's/^/          /'
   else printf '   \033[32mPASS\033[0m %s\n' "$label"; fi
 }
@@ -152,7 +188,7 @@ assert_not_committed() {
   local label="$1" raw
   raw=$(adbs "strings /data/data/$PKG/files/datastore/dashpay.preferences_pb 2>/dev/null" | tr -d '\r')
   if echo "$raw" | grep -qE "CUT_OVER|SETTLED"; then
-    printf '   \033[31mFAIL\033[0m %s\n        (persisted state is committed)\n' "$label"
+    fail "$label" "(persisted state is committed)"
   else
     printf '   \033[32mPASS\033[0m %s\n' "$label"
   fi
@@ -161,7 +197,7 @@ assert_not_committed() {
 fake_pre_cutover_previous_launch() {
   # Configuration.lastVersionCode is read from the default SharedPreferences
   # ("last_version", Configuration.java:61) at construction. Writing it here is
-  # exactly equivalent to "the previous launch ran a pre-11.10 build", which is
+  # exactly equivalent to "the previous launch ran a pre-cutover build", which is
   # what CutoverCoordinator's GATE 1 tests — without needing a real v11.9.0 APK
   # signed with matching keys.
   #
@@ -175,13 +211,13 @@ fake_pre_cutover_previous_launch() {
     note "prefs file not present yet — launch the app once, complete onboarding, then re-run"
     return 1
   fi
-  # If a REAL upgrade was just installed over a pre-11.10 build, lastVersionCode
+  # If a REAL upgrade was just installed over a pre-cutover build, lastVersionCode
   # already holds the genuine value — leave it alone. Faking it would only
   # overwrite the truth with the same thing, and hide a mismatch if the real
   # upgrade did not record what we expect.
   local cur
   cur=$(adbs "grep -o 'last_version\" value=\"[0-9]*' $f" 2>/dev/null | grep -o '[0-9]*$' | tr -d '\r')
-  if [ -n "$cur" ] && [ "$cur" -gt 0 ] && [ "$cur" -lt 11100000 ]; then
+  if [ -n "$cur" ] && [ "$cur" -gt 0 ] && [ "$cur" -lt "$FIRST_CUTOVER_VERSION_CODE" ]; then
     note "REAL pre-cutover upgrade detected (last_version=$cur) — not faking it"
     return 0
   fi
@@ -461,7 +497,7 @@ s4)
   if [ "${COUNT:-0}" -ge 2 ]; then
     printf '   \033[32mPASS\033[0m engine restarted after the teardown (%s starts)\n' "$COUNT"
   else
-    printf '   \033[31mFAIL\033[0m engine did NOT restart — MO-995 latch still present (%s starts)\n' "$COUNT"
+    fail "engine did NOT restart — MO-995 latch still present (${COUNT:-0} starts)"
   fi
   ;;
 
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -28,7 +28,8 @@
 #         ./scripts/cutover-emulator-test.sh setup      # build+sign+upgrade-install
 #         ./scripts/cutover-emulator-test.sh s1        # upgrade, bind works
 #         ./scripts/cutover-emulator-test.sh s2        # upgrade, bind denied
-#         ./scripts/cutover-emulator-test.sh s3        # denied then healed
+#         ./scripts/cutover-emulator-test.sh s3        # denied then healed in-session
+#         ./scripts/cutover-emulator-test.sh s3b       # ...and the next launch commits
 #         ./scripts/cutover-emulator-test.sh s4        # trim -> SPV restart
 #         ./scripts/cutover-emulator-test.sh log       # pull + tail wallet.log
 set -uo pipefail
@@ -136,6 +137,22 @@ refute_log() {
     printf '   \033[31mFAIL\033[0m %s\n        (unexpected: %s)\n' "$label" "$pat"
     grep -E "$pat" "$w" | head -2 | sed 's/^/          /'
   else printf '   \033[32mPASS\033[0m %s\n' "$label"; fi
+}
+
+# assert_not_committed <label> — reads the PERSISTED state, not the log.
+# The engine-gate lines (`dashjEngineMayStart=`, `starting peergroup`) are only
+# emitted at service onCreate. A scenario that deliberately keeps ONE process
+# alive has no new onCreate, so those lines cannot appear in its window and
+# asserting on them is a guaranteed false FAIL. The DataStore is the durable
+# source of truth for who owns L1.
+assert_not_committed() {
+  local label="$1" raw
+  raw=$(adbs "strings /data/data/$PKG/files/datastore/dashpay.preferences_pb 2>/dev/null" | tr -d '\r')
+  if echo "$raw" | grep -qE "CUT_OVER|SETTLED"; then
+    printf '   \033[31mFAIL\033[0m %s\n        (persisted state is committed)\n' "$label"
+  else
+    printf '   \033[32mPASS\033[0m %s\n' "$label"
+  fi
 }
 
 fake_pre_cutover_previous_launch() {
@@ -364,18 +381,49 @@ s2)
   ;;
 
 s3)
-  say "S3 — denied, then healed (recovery path; expected to expose the noteAppForeground gap)"
+  say "S3 — denied, then healed IN-SESSION (the walletB recovery path)"
   require_device; require_root
-  note "assumes S2 just ran: state DUAL_RUNNING, bind marker unset"
+  note "assumes S2 just ran: state DUAL_RUNNING, bind marker unset, screen locked"
   show_state
-  mark_log          # S3 had NO mark, so its assertions matched the whole log history
+  mark_log
   wake_unlock
   launch_app
-  assert_log "bind succeeded after unlock"       "app wallet (bound to new|already bound to) SDK wallet"
-  note "does it recover WITHOUT a restart? (this is the open question)"
-  assert_log "committed in-session"              "cutover state DUAL_RUNNING -> CUT_OVER"
-  note "if the line above FAILED, restart the app and re-check — needing a restart confirms"
-  note "that noteAppForeground() only resets a backoff nothing reads once the wait loop ended"
+
+  # The recovery chain, in order. Each of these was broken until b49ef25b0:
+  #   - ProcessLifecycleOwner never fired, so the foreground signal never existed
+  #   - noteAppForeground was state-only, so nothing drove a retry
+  #   - nothing else calls maybeRetry once the cutover correctly declines to commit
+  assert_log "app-foreground signal fired"       "App moved to foreground"
+  assert_log "foreground drove a bind retry"     "app foregrounded with an SDK bind retry pending"
+  assert_log "the retry ran a pass"              "SDK bind retry [0-9]+ \(app foreground\)"
+  assert_log "bind healed after the unlock"      "app wallet (bound to new|already bound to) SDK wallet"
+  assert_log "retry pressure cleared"            "bind established after [0-9]+ failed pass"
+
+  # …and it must be the SAME process: a restart would heal it trivially and prove
+  # nothing. walletB restarted repeatedly and never recovered.
+  refute_log "healed WITHOUT a process restart"  "WalletApplication.onCreate\(\)"
+
+  # dashj must still own L1 for this launch. The commit is NOT expected here and
+  # asserting it was my error: the upgrade seam only runs at process start, so
+  # in-session the only route is the readiness-gated auto-commit, which needs
+  # MIN_PARITY_STREAK readings at a 10s throttle AND the SDK scan caught up to
+  # tip. 45s cannot satisfy that, so a missing commit here is correct deferral,
+  # not a defect. The commit is checked on the NEXT launch instead — see s3b.
+  assert_not_committed "dashj still owns L1 (state not committed)"
+  note "commit is deliberately NOT asserted here — run s3b to check the next launch"
+  ;;
+
+s3b)
+  say "S3b — the launch AFTER an in-session heal should commit"
+  require_device; require_root
+  show_state
+  note "the bind marker must be set by now (s3 healed it); the seam can act at process start"
+  adbs am force-stop "$PKG"; sleep 3
+  mark_log
+  wake_unlock
+  launch_app
+  assert_log "committed on the next launch"      "cutover state DUAL_RUNNING -> CUT_OVER|READY_OBSERVED -> CUT_OVER"
+  assert_log "SDK L1 engine started"             "L1 shadow SPV started"
   ;;
 
 s4)
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -37,7 +37,10 @@ set -uo pipefail
 PKG=hashengineering.darkcoin.wallet_test
 SVC=$PKG/de.schildbach.wallet.service.BlockchainServiceImpl
 AVD=Pixel_9_API_36_AOPS
-PIN=REDACTED
+# Device lock PIN, for the scenarios that must lock and unlock the emulator.
+# NEVER hardcode it — supply it per run:  EMULATOR_PIN=... ./this-script s3
+# Only the scenarios that lock the screen need it (s2, and s3's unlock).
+PIN="${EMULATOR_PIN:-}"
 PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
 BT=~/Library/Android/sdk/build-tools/35.0.1
 APK_DIR=wallet/build/outputs/apk/_testNet3/release
@@ -253,6 +256,11 @@ keep_awake()    { adbs svc power stayon true >/dev/null 2>&1
 # scenarios that needed a working bind. KEYCODE_0 is 7, so digit d -> 7+d.
 pin_keyevents() {
   local i ch
+  if [ -z "$PIN" ]; then
+    note "no PIN supplied — set EMULATOR_PIN=<pin> to let the script unlock the device,"
+    note "or unlock it by hand before running this scenario."
+    return 1
+  fi
   for (( i=0; i<${#PIN}; i++ )); do
     ch=${PIN:$i:1}
     adbs input keyevent $((7 + ch))
@@ -337,8 +345,13 @@ setup)
   fi
   note "installed: $(adbs "dumpsys package $PKG | grep -m1 versionName" | tr -d ' \r')"
   say "Ensure a secure lock screen (this is what makes the master alias lock-bound)"
-  adbs locksettings set-pin "$PIN" 2>/dev/null && note "PIN set to $PIN" \
-    || note "locksettings unavailable — set a PIN via Settings before running s2"
+  if [ -n "$PIN" ]; then
+    adbs locksettings set-pin "$PIN" >/dev/null 2>&1 \
+      && note "device PIN set from EMULATOR_PIN" \
+      || note "locksettings unavailable — set a device PIN via Settings before running s2"
+  else
+    note "no EMULATOR_PIN given — set a device PIN yourself; s2/s3 need one to lock+unlock"
+  fi
   note "then run: $0 s1"
   ;;
 
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -1,0 +1,373 @@
+#!/bin/bash
+#
+# MO-995 cutover / SPV-restart test harness for the Android emulator.
+#
+# WHY A RELEASE BUILD: PlatformSyncService.shutdown() skips stopSdkEngines()
+# when BuildConfig.DEBUG (PlatformSyncService.kt:656) — a deliberate warm-SPV
+# battery trade-off. On a debug build the SDK engine never stops on teardown,
+# so neither MO-995 nor its fix can be observed. Everything below needs the
+# _testNet3Release variant.
+#
+# WHY THE AOSP EMULATOR IMAGE: steps 1-3 fake the previous-launch versionCode
+# by writing shared_prefs directly, which needs `adb root`. The
+# android-36/google_apis_playstore image blocks root; android-36/default
+# allows it. Use Pixel_9_API_36_AOPS (Android 16, arm64-v8a — matches all four
+# field devices).
+#
+# WHAT THIS CANNOT REPRODUCE: walletB's FALSE-LOCKED classification
+# (KeyguardManager reports unlocked while Keystore denies). That is the HONOR
+# PTP-N49 OEM defect; on AOSP every denial classifies as "genuinely locked".
+# The false-locked branch is covered by SdkBindRetryServiceTest instead. A
+# green run here does NOT clear the HONOR case.
+#
+# ORDER OF OPERATIONS: you install the PRE-CUTOVER build yourself and onboard
+# /sync it. Then `setup` builds, signs and installs the fix build OVER it —
+# that install is the upgrade under test. The scenarios run after that.
+#
+# Usage:  ./scripts/cutover-emulator-test.sh <step>
+#         ./scripts/cutover-emulator-test.sh setup      # build+sign+upgrade-install
+#         ./scripts/cutover-emulator-test.sh s1        # upgrade, bind works
+#         ./scripts/cutover-emulator-test.sh s2        # upgrade, bind denied
+#         ./scripts/cutover-emulator-test.sh s3        # denied then healed
+#         ./scripts/cutover-emulator-test.sh s4        # trim -> SPV restart
+#         ./scripts/cutover-emulator-test.sh log       # pull + tail wallet.log
+set -uo pipefail
+
+PKG=hashengineering.darkcoin.wallet_test
+SVC=$PKG/de.schildbach.wallet.service.BlockchainServiceImpl
+AVD=Pixel_9_API_36_AOPS
+PIN=1234
+PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
+BT=~/Library/Android/sdk/build-tools/35.0.1
+APK_DIR=wallet/build/outputs/apk/_testNet3/release
+OUT=/tmp/mo995-emulator
+mkdir -p "$OUT"
+
+say()  { printf '\n\033[1m== %s\033[0m\n' "$*"; }
+note() { printf '   %s\n' "$*"; }
+adbs() { adb shell "$@"; }
+
+require_device() {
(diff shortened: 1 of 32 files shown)


HashEngineering:
@coderabbitai review.

scripts/cutover-emulator-test.sh
```diff
@@ -0,0 +1,477 @@
...
+  local label="$1" pat="$2" budget="${3:-45}" waited=0
+  while [ "$waited" -lt "$budget" ]; do
+    if grep -qE "$pat" "$(since_mark)"; then
+      printf '   \033[32mPASS\033[0m %s%s\n' "$label" "$([ "$waited" -gt 0 ] && echo " (after ${waited}s)")"
+      return 0
+    fi
+    sleep 5; waited=$((waited + 5))
+  done
+  printf '   \033[31mFAIL\033[0m %s\n        (no match in %ss for: %s)\n' "$label" "$budget" "$pat"
+}
```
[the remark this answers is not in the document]
HashEngineering: Valid — fixed in d66bfe0a3. All three helpers ended in a `printf`, so they returned 0 and the script exited 0 after printing FAIL.

Implemented differently than suggested: a `FAILURES` ledger plus an `EXIT` trap, rather than `exit 1` at each failing assertion. Each scenario is a *sequence* — S3 asserts "bind failed" → "dashj came back" → "state not committed", and knowing *where* the chain broke is the point of the harness. Aborting at the first FAIL would only report that it broke. Assertions still print and continue; only the exit status changed. Verified: clean run exits 0, any failure exits 1.

wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverCoordinator.kt
```diff
@@ -459,7 +547,7 @@ class CutoverCoordinator @Inject constructor(
          * one-time upgrade sync explainer arms only for upgrades crossing
          * this boundary.
          */
-        const val FIRST_CUTOVER_VERSION_CODE = 11100000
+        const val FIRST_CUTOVER_VERSION_CODE = 12000000
```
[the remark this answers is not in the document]
HashEngineering: Not reverting the constant — `12000000` is correct. The cutover ships in 12.0.0 (`wallet/build.gradle` versionName `12.0.0`); the boundary moved from 11100000 when the release slipped from 11.10. Setting it back would classify every 11.10–11.25 build as *already cut over* and silently deny those users the one-time sync explainer, which is the opposite of the intent.

That said, the inconsistency you flagged was real, and one instance was a genuine bug — fixed in d66bfe0a3:

`fake_pre_cutover_previous_launch` in `scripts/cutover-emulator-test.sh` kept its own hardcoded copy of the boundary (`[ "$cur" -lt 11100000 ]`). A device last run on 11.10–11.25 — pre-cutover as far as the app is concerned — was judged post-cutover by the script, so it overwrote a genuine `last_version` with a faked one, defeating the "leave the truth alone" branch its own comment describes. It now reads a single variable documented to track the constant.

The remaining hits were stale prose: KDoc and comments naming "11.10" in eight places. Version numbers now appear only where the constant is defined, with a note that this boundary has already moved once. The test fixture was already derived from the constant (it had been hardcoded to 11100100 and silently inverted its own assertion when the boundary moved).

wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverCoordinator.kt
```diff
@@ -273,7 +277,51 @@ class CutoverCoordinator @Inject constructor(
...
+                runCatching {
+                    dashPayConfig.set(DashPayConfig.CUTOVER_UPGRADE_BOUNDARY_CROSSED, true)
+                }.onFailure {
+                    if (it is CancellationException) throw it
+                    log.warn("failed to latch the cutover boundary crossing", it)
+                }
+            }
+            val crossedEver = crossedNow || runCatching {
+                dashPayConfig.get(DashPayConfig.CUTOVER_UPGRADE_BOUNDARY_CROSSED) == true
+            }.getOrDefault(false)
```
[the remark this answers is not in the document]
HashEngineering: Confirmed and fixed in `82969ccc6`.

The analysis holds exactly as written. `armUpgradeNoticeIfUpgraded` reads the persisted latch rather than the in-memory `crossedNow`, so a failed `set` skipped the explainer on the launch that observed the crossing — and `Configuration.lastVersionCode` has advanced by the next launch, so `crossedNow` is false thereafter with nothing on disk to recover from. This write site was indeed the only writer of the key.

Scope worth recording: the cutover itself is unaffected. The readiness-gated auto-commit path does not consult this latch — only the explainer does — so the failure costs the one-time sync explanation, not sync or cutover correctness. That is MO-1022's symptom reached by a different route, which is why it was worth closing.

The fix keeps the crossing in memory for the launch that observed it and retries the persist when the cutover commits, so a transient DataStore failure now costs nothing. If the store never accepts the key, arming still proceeds from memory — that is the limit of what is salvageable, since the explainer's own flags (`CUTOVER_UPGRADE_NOTICE_EVER_ARMED` / `_PENDING`) need the same store, and `armUpgradeNoticeOnce` already logs when those fail.

Two tests as requested, both verified to fail against the unfixed code: a write that fails once and is re-latched on retry, and one that never persists.

wallet/src/de/schildbach/wallet/service/L1SyncStatusService.kt
```diff
@@ -384,8 +386,61 @@ internal fun mergeL1SyncDetail(
...
+        } else {
+            progress.filterHeight
+        },
+        filterTarget = if (idleOrConnecting) {
+            maxOf(
+                progress.filterTarget,
+                lastKnownFilterTarget,
+                progress.walletSyncedHeight,
+                lastKnownFilterHeight
+            )
```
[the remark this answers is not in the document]
HashEngineering: Taken in `82969ccc6`, though the stated mechanism does not hold.

`L1ShadowSyncService` does not map the two fields independently in the sense of them drifting apart. `toShadowSyncProgress` reads both off the *same* `data.filters` snapshot:

```kotlin
filterHeight = data.filters?.currentHeight ?: 0,
filterTarget = data.filters?.targetHeight ?: 0,
```

so there is no partial-update window where one is fresh and the other stale, and a null `filters` gives both `0`. Inverting the pair therefore requires the engine to report `currentHeight > targetHeight` within one snapshot — and if it ever does, the non-IDLE/CONNECTING branch passes both through raw, so the same inverted pair renders with or without these floors. The floors are not the cause.

The change is still right, for a different reason: the comment directly above the block asserts that the target "takes the SAME floors as the height, so the pair can never render as height > target", and that was false while `progress.filterHeight` was missing from the target's `maxOf`. The code did not match its own documented invariant. Both are now fixed — the term is added and the comment says what actually holds.

Regression test added (`detail_filterTargetNeverTrailsTheHeight_onAnInvertedEngineSnapshot`), verified to fail against the unfixed code with `target 2533349 must not trail height 2533400`.

thepastaclaw:
✅ Final review complete — no blockers (commit c77b5e86) · triage: normal · Phase 2 only (queue backlog)

scripts/cutover-emulator-test.sh
```diff
@@ -0,0 +1,513 @@
...
+  # Marking while the app is already up leaves nothing to observe and reports a
+  # bogus FAIL for a healthy engine.
+  adbs am force-stop "$PKG"; sleep 3
+  mark_log
+  launch_app
+  assert_log "engine started on a cold launch"   "L1 shadow SPV started"
+  note "firing TRIM_MEMORY_BACKGROUND deterministically"
+  adbs am send-trim-memory "$PKG" BACKGROUND; sleep 6
+  assert_log "service tore down"                 "low memory detected, stopping service"
+  assert_log "engine stopped (release build)"    "L1 shadow sync stopped"
```
thepastaclaw, line 493: **🔴 Blocking: Update S4 to exercise the new trim-memory policy and log messages**

S4 still sends `TRIM_MEMORY_BACKGROUND`, but `BlockchainServiceImpl` now stops only for `TRIM_MEMORY_COMPLETE` or `TRIM_MEMORY_RUNNING_CRITICAL`; `TRIM_MEMORY_BACKGROUND` is intentionally ignored. Its assertions also search for `low memory detected, stopping service` and `L1 shadow sync stopped`, while the implementation now logs `memory pressure (onTrimMemory level ...), stopping service` and `L1ShadowLifecycle STOPPED ...`. As written, S4 will report failure instead of validating the claimed trim-to-SPV-restart path.

```suggestion
  adbs am send-trim-memory "$PKG" COMPLETE; sleep 6
  assert_log "service tore down"                 "memory pressure \\(onTrimMemory level .*\\), stopping service"
  assert_log "engine stopped (release build)"    "L1ShadowLifecycle STOPPED"
```

<sub>source: `gpt-6-astra` (phase2-reviewer: general, security-auditor)</sub>

thepastaclaw (changes requested):

The production trim-memory policy correctly avoids tearing down the service for ordinary backgrounding, but the S4 emulator scenario was not updated with that policy. It still sends TRIM_MEMORY_BACKGROUND and searches for log messages removed by this PR, so the documented trim-to-SPV-restart verification cannot pass against the current implementation.

🔴 1 blocking

<details>

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a large, intricate, cross-cutting behavioral change spanning service lifecycle, SDK cutover and binding, foreground handling, build configuration, and extensive tests, but it does not directly modify consensus, funds movement, cryptography, key handling, peer-facing deserialization, or storage migrations.
- Phase 1 reviewers: **not run (skipped for throughput: 25 PRs queued, above the 10 limit)**

</details>

<details>

```

In `scripts/cutover-emulator-test.sh`:
- [BLOCKING] scripts/cutover-emulator-test.sh:490-493: Update S4 to exercise the new trim-memory policy and log messages
  S4 still sends `TRIM_MEMORY_BACKGROUND`, but `BlockchainServiceImpl` now stops only for `TRIM_MEMORY_COMPLETE` or `TRIM_MEMORY_RUNNING_CRITICAL`; `TRIM_MEMORY_BACKGROUND` is intentionally ignored. Its assertions also search for `low memory detected, stopping service` and `L1 shadow sync stopped`, while the implementation now logs `memory pressure (onTrimMemory level ...), stopping service` and `L1ShadowLifecycle STOPPED ...`. As written, S4 will report failure instead of validating the claimed trim-to-SPV-restart path.
```
</details>

scripts/cutover-emulator-test.sh
HashEngineering: Confirmed and fixed in `39d188d94`.

Verified all three points against the source before changing anything:

- `shouldStopForMemoryPressure(40)` is false (`40 >= TRIM_MEMORY_COMPLETE` false, `40 == TRIM_MEMORY_RUNNING_CRITICAL` false), so `TRIM_MEMORY_BACKGROUND` is indeed ignored and the teardown assertions could never be satisfied.
- `low memory detected, stopping service` survives only inside a KDoc example at `BlockchainServiceImpl.kt:283,293`; the live call at :2861 is `memory pressure (onTrimMemory level {}), stopping service`.
- `L1 shadow sync stopped` is gone entirely; `L1ShadowSyncService.kt:2154` now logs `L1ShadowLifecycle STOPPED after {} up`.

So S4 would have reported FAIL against correct code — and since `fddb0e244` made the harness exit non-zero, that FAIL breaks the run.

I went one step beyond the suggestion. Swapping BACKGROUND for COMPLETE fixes the false failure but drops coverage of the actual MO-995 bug, which was that BACKGROUND tore the service down. S4 now drives both halves: `refute_log` that level 40 produces no teardown, then `assert_log` that COMPLETE does. Without the refute, restoring `level >= TRIM_MEMORY_BACKGROUND` would reintroduce the bug and S4 would still pass on the COMPLETE leg alone.

RUNNING_CRITICAL is the other stop level, but `am send-trim-memory` delivers it while the app is foreground-visible where the framework may re-raise it, so COMPLETE is the deterministic one to assert.

Also repointed two comment SHAs the rebase orphaned: `ea506f978` -> `b994bb6a1`, `b49ef25b0` -> `5d89950bc`.

Not executed — S4 needs a rooted emulator with a cut-over wallet. Checked with `bash -n`, and the new patterns verified against the real log format while the old ones match nothing.

scripts/cutover-emulator-test.sh
```diff
@@ -0,0 +1,527 @@
...
+  note "firing TRIM_MEMORY_BACKGROUND — the service must IGNORE it"
+  adbs am send-trim-memory "$PKG" BACKGROUND; sleep 6
+  refute_log "background trim ignored (no teardown)" \
+      "memory pressure \(onTrimMemory level 40\), stopping service"
+  # COMPLETE (80) is genuine pressure and MUST stop the service. RUNNING_CRITICAL
+  # (15) is the other stop level, but `am send-trim-memory` delivers it while the
+  # app is foreground-visible, where the framework may re-raise it; COMPLETE is
+  # the deterministic one.
+  note "firing TRIM_MEMORY_COMPLETE — genuine pressure, the service must stop"
+  adbs am send-trim-memory "$PKG" COMPLETE; sleep 6
```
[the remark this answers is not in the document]
HashEngineering: Confirmed and fixed in `45c83c8e7`, though by a different route than suggested.

The flaw is real: `refute_log` does not advance `LOG_MARK`, so both legs share one window, and `memory pressure \(onTrimMemory level [0-9]+\)` is satisfied by a level-40 line. A regression to `level >= TRIM_MEMORY_BACKGROUND` whose teardown landed just after the refute sampled would have been reported as a PASS by the assertion meant to catch it.

I did not take the `mark_log` though, for the reason your own comment identifies: re-marking before the COMPLETE leg drops the cold-launch `L1 shadow SPV started` out of the window, so the `>= 2` restart count at the end of the scenario sees one start and fails. That trades a false pass for a false failure.

Pinning the level instead fixes it without touching the window:

```
assert_log "service tore down (level 80)"  "memory pressure \(onTrimMemory level 80\), stopping service"
```

`am send-trim-memory COMPLETE` delivers exactly `TRIM_MEMORY_COMPLETE` (80), so this is precise, and a level-40 line can no longer satisfy it no matter when it arrives. Verified both directions against sample logs: the new pattern does not match a level-40-only log, and does match a real level-80 teardown. The restart count keeps its full window.

The neighbouring `L1ShadowLifecycle STOPPED` assertion stays unpinned deliberately — a stop is a stop, and the reason it happened is now pinned by the level-80 line above it.

wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverCoordinator.kt
```diff
@@ -273,7 +297,59 @@ class CutoverCoordinator @Inject constructor(
...
+                }.onFailure {
+                    if (it is CancellationException) throw it
+                    // Remember it in memory: the crossing is observable on this
+                    // launch only, so dropping it here costs the explainer for
+                    // good. Retried in armUpgradeNoticeIfUpgraded.
+                    boundaryCrossingUnpersisted = true
+                    log.warn(
+                        "failed to latch the cutover boundary crossing — holding it in memory " +
+                            "for this launch and retrying when the cutover commits",
+                        it
```
thepastaclaw, line 330: **🔴 Blocking: Retry the boundary latch even when this launch cannot commit**

When the first upgrade launch observes the boundary but `set(CUTOVER_UPGRADE_BOUNDARY_CROSSED, true)` fails, this code records the crossing only in `boundaryCrossingUnpersisted`. The subsequent `commitLocked(..., requireBindEvidence = true)` can decline because SDK bind evidence is not yet available, and the only retry is inside `armUpgradeNoticeIfUpgraded()`, which runs after a successful cutover. If the process ends before readiness-driven auto-commit, the next coordinator instance loses the in-memory flag, while `Configuration.lastVersionCode` has already advanced and `crossedNow` is false. The boundary can then never be recovered, so the eventual commit skips the upgrade explainer. Retry the latch independently of successful cutover, and add a test where the first write fails with bind evidence initially absent, followed by a new coordinator instance that successfully persists the latch.

scripts/cutover-emulator-test.sh
```diff
@@ -0,0 +1,535 @@
...
+    printf '   \033[32mPASS\033[0m engine restarted after the teardown (%s starts)\n' "$COUNT"
+  else
+    fail "engine did NOT restart — MO-995 latch still present (${COUNT:-0} starts)"
+  fi
+  ;;
+
+log)
+  require_device
+  f=$(pull_log); echo "$f"
+  grep -nE "cutover state|declining to commit|Keystore denied|L1 shadow SPV started|L1 shadow sync stopped|low memory detected|idling detected|starting peergroup|holding the dashj L1 engine|bind has never succeeded|explainer armed" "$f" | tail -40
```
thepastaclaw: **🟡 Suggestion: Update the diagnostic log helper for renamed lifecycle messages**

The `log` command still filters for `L1 shadow sync stopped` and `low memory detected`, but the live implementation now emits `L1ShadowLifecycle STOPPED` and `memory pressure (onTrimMemory level ...), stopping service`. Consequently, running the diagnostic helper after an outage omits the engine stop and memory-pressure events that are central to diagnosing this PR's MO-995 behavior. This does not affect scenario pass/fail results, but it makes the documented troubleshooting command misleading.

```suggestion
grep -nE "cutover state|declining to commit|Keystore denied|L1 shadow SPV started|L1ShadowLifecycle (STOPPED|RESUMING)|memory pressure \\(onTrimMemory level|idling detected|starting peergroup|holding the dashj L1 engine|bind has never succeeded|explainer armed" "$f" | tail -40
```

<sub>source: `glm-5.3-flash` (phase1-reviewer: general, security-auditor)</sub>

wallet/src/de/schildbach/wallet/ui/dashpay/utils/DashPayConfig.kt
```diff
@@ -420,12 +420,87 @@ open class DashPayConfig @Inject constructor(
...
+         * has ever bound successfully still commits promptly, while a device
+         * whose keystore denies the bind never commits at all — the difference
+         * between walletC/D (bind fine, cut over) and walletB (16 keystore
+         * denials, cut over anyway, left with NO L1 engine).
+         */
+        val SDK_BIND_EVER_SUCCEEDED = booleanPreferencesKey("sdk_bind_ever_succeeded")
+
+        /**
+         * MO-995: set the first time a launch is seen to have crossed the
+         * cutover boundary (the previous launch ran a pre-11.10 build), and
```
thepastaclaw: **💬 Nitpick: Use the current cutover boundary in the new KDoc**

The KDoc for `CUTOVER_UPGRADE_BOUNDARY_CROSSED` says the previous launch ran a pre-11.10 build, but `FIRST_CUTOVER_VERSION_CODE` is now 12000000 for the 12.0.0 cutover. The stale release name is misleading: 11.10 through 11.25 are below the actual boundary and should qualify as pre-cutover. Keep the explanation version-independent and refer to the coordinator constant instead, so future boundary moves do not leave contradictory documentation.

```suggestion
         * MO-995: set the first time a launch is seen to have crossed the
         * cutover boundary (the previous launch ran a pre-cutover build, i.e.
         * [CutoverCoordinator.FIRST_CUTOVER_VERSION_CODE] or below), and
         * never cleared except by a wallet wipe.
```

<sub>source: `glm-5.3-flash` (phase1-reviewer: general, security-auditor)</sub>

thepastaclaw (changes requested):
## Re-review — Final validation — Phase 1 + Phase 2

The PR fixes the previously reported S4 harness defect, but the boundary-latch recovery remains incomplete. A transient failure on the first upgrade launch can leave the crossing only in the coordinator instance's memory; if that launch cannot commit and the process ends, a later instance permanently loses upgrade-explainer eligibility. The diagnostic log helper and newly added boundary KDoc also retain stale log/version references.

🔴 1 blocking | 🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

Source: reviewer 1: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `security-auditor`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `normal` by `gpt-6-astra` (effort low) — The diff is large and intricate but changes engine lifecycle, cutover eligibility, bind-retry orchestration, UI formatting, diagnostics and build configuration rather than cryptography, key handling, funds movement or storage migrations; SdkWalletBinder.kt only adds persistence of a successful-bind marker.
- Phase 1 reviewers: `glm-5.3-flash` — general (completed, effort max); agent `phase1-reviewer`, `glm-5.3-flash` — security-auditor (completed, effort max); agent `phase1-reviewer`
- Phase 1 model: `glm-5.3-flash` — zai quota: 5h 98% left, weekly 46% left; passed over `gemini-3.8-flash-high` (antigravity below 15% reserve: weekly 11% left, 5h 100% left)

</details>

<details>

```

In `wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverCoordinator.kt`:
- [BLOCKING] wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverCoordinator.kt:321-328: Retry the boundary latch even when this launch cannot commit
  When the first upgrade launch observes the boundary but `set(CUTOVER_UPGRADE_BOUNDARY_CROSSED, true)` fails, this code records the crossing only in `boundaryCrossingUnpersisted`. The subsequent `commitLocked(..., requireBindEvidence = true)` can decline because SDK bind evidence is not yet available, and the only retry is inside `armUpgradeNoticeIfUpgraded()`, which runs after a successful cutover. If the process ends before readiness-driven auto-commit, the next coordinator instance loses the in-memory flag, while `Configuration.lastVersionCode` has already advanced and `crossedNow` is false. The boundary can then never be recovered, so the eventual commit skips the upgrade explainer. Retry the latch independently of successful cutover, and add a test where the first write fails with bind evidence initially absent, followed by a new coordinator instance that successfully persists the latch.

In `scripts/cutover-emulator-test.sh`:
- [SUGGESTION] scripts/cutover-emulator-test.sh:529: Update the diagnostic log helper for renamed lifecycle messages
  The `log` command still filters for `L1 shadow sync stopped` and `low memory detected`, but the live implementation now emits `L1ShadowLifecycle STOPPED` and `memory pressure (onTrimMemory level ...), stopping service`. Consequently, running the diagnostic helper after an outage omits the engine stop and memory-pressure events that are central to diagnosing this PR's MO-995 behavior. This does not affect scenario pass/fail results, but it makes the documented troubleshooting command misleading.

In `wallet/src/de/schildbach/wallet/ui/dashpay/utils/DashPayConfig.kt`:
- [NITPICK] wallet/src/de/schildbach/wallet/ui/dashpay/utils/DashPayConfig.kt:483-484: Use the current cutover boundary in the new KDoc
  The KDoc for `CUTOVER_UPGRADE_BOUNDARY_CROSSED` says the previous launch ran a pre-11.10 build, but `FIRST_CUTOVER_VERSION_CODE` is now 12000000 for the 12.0.0 cutover. The stale release name is misleading: 11.10 through 11.25 are below the actual boundary and should qualify as pre-cutover. Keep the explanation version-independent and refer to the coordinator constant instead, so future boundary moves do not leave contradictory documentation.
```
</details>

wallet/src/de/schildbach/wallet/service/platform/sdk/CutoverCoordinator.kt
HashEngineering: Confirmed and fixed in `21ce82628`. The analysis is exactly right, and this was my bug — `82969ccc6` moved the defect rather than closing it.

`armUpgradeNoticeIfUpgraded` is reached only through a successful commit, and on a real upgrade the seam declines because the bind runs after it, so `commitForUpgradedWalletAsync` returns at `if (!justCutOver)` first. The KDoc two lines above GATE 2 states that outright. So the retry I added sat behind the one branch that the launch which can observe the crossing never takes: the flag died with the process, `Configuration.lastVersionCode` had advanced, and the explainer was gone permanently.

`persistBoundaryCrossingIfPending()` now runs where that launch actually reaches it:

- the decline path, immediately before `return@launch` — the last code to run on the upgrade launch;
- the top of `transition()`, which the readiness observer drives repeatedly for as long as the process lives, so a transient failure gets every remaining chance rather than one;
- arm time, as before.

Test added as requested, with two coordinator instances over one store — the only construction that can demonstrate it, since the in-memory flag cannot cross a process boundary, so the second instance commits only if the latch genuinely reached disk. Launch 1: latch write fails, bind evidence absent, commit correctly refused, latch must still be persisted. Launch 2: new instance, bind now succeeded, commits off the persisted latch and arms the explainer.

Mutation-verified: with the two new retry sites reverted, it fails on `the declining launch must still get the boundary crossing onto disk`. Full wallet suite green (1965 tests).

scripts/cutover-emulator-test.sh
HashEngineering: Confirmed and fixed in `21ce82628`. `39d188d94` fixed S4's assertions but missed this filter, so the two events most worth seeing after an MO-995 outage were the two the helper dropped. Now greps `L1ShadowLifecycle STOPPED` and `memory pressure \(onTrimMemory` alongside the rest.

wallet/src/de/schildbach/wallet/ui/dashpay/utils/DashPayConfig.kt
HashEngineering: Confirmed and fixed in `21ce82628`. The KDoc said "pre-11.10" while `FIRST_CUTOVER_VERSION_CODE` is `12000000`, so it read 11.10-11.25 as already cut over when they are below the boundary and do qualify.

Taken the version-independent framing you suggested: it now refers to `CutoverCoordinator.FIRST_CUTOVER_VERSION_CODE` rather than a release name, and records why — the boundary has already moved once (11100000 -> 12000000 when the cutover slipped from 11.10 to 12.0.0), which is precisely what turned this sentence false. A constant cannot go stale the next time it moves.

wallet/src/de/schildbach/wallet/WalletActivityTracker.java
```diff
@@ -111,12 +124,34 @@ public void onActivityPaused(@NonNull Activity activity) {
...
+            AppForegroundMonitor.INSTANCE.noteBackground();
+            // MO-995: invoke the last-stopped hook explicitly. The base class's
+            // onStoppedLast() is NEVER reached, because this class overrides
+            // onActivityStarted/onActivityStopped without calling super, so
+            // ActivitiesTracker.numStarted never advances. That has been true
+            // since before this class existed — the anonymous tracker it
+            // replaced (WalletApplication, pre-fae81fefb) had the identical
+            // bug — so autoLogout has not been told the app backgrounded, and
+            // the zero-minute force-finish has not fired, for a long time.
+            onStoppedLast();
```
thepastaclaw: **🔴 Blocking: Exclude configuration changes from the newly enabled logout hook**

Recreating the only visible activity for a configuration change stops the old instance before starting its replacement, temporarily reducing visibleActivityCount to zero. This newly added call then sets appWentBackground=true and, when automatic logout is enabled with a zero-minute timeout, broadcasts FORCE_FINISH_ACTION. AddressBookActivity is a concrete affected screen: its manifest permits rotation and does not handle orientation changes, and it inherits LockScreenActivity through AbstractBindServiceActivity. Rotating that screen now makes AutoLogout.shouldLogout() return true, so the replacement activity displays the lock screen even though the user never left the wallet. Keep the activity counter balanced, but exclude activity.isChangingConfigurations() from the logout decision. Add lifecycle regression coverage for both configuration recreation and genuine backgrounding.

thepastaclaw (changes requested):
## Re-review — Final validation — Phase 2 only (queue backlog)

Verified the supplied Phase-2 finding and all four prior findings against head 215fd66910a8b717a216c5a93ed19b03aca4f824. All four prior issues are fixed, but the newly activated logout hook still treats activity recreation as genuine backgrounding and can lock the wallet during rotation. Shell syntax and diff whitespace checks passed; Android/JVM tests and emulator scenarios were not run during this verification.

🔴 1 blocking

<details>

- Triage: `normal` by `gpt-6-astra` (effort low) — The diff is large and intricate but changes engine lifecycle, cutover eligibility, bind-retry orchestration, UI error handling, and release configuration rather than consensus, funds movement, cryptography, key handling, peer-facing deserialization, or storage migrations.
- Phase 1 reviewers: **not run (skipped for throughput: 29 PRs queued, above the 10 limit)**

</details>

<details>

```

In `wallet/src/de/schildbach/wallet/WalletActivityTracker.java`:
- [BLOCKING] wallet/src/de/schildbach/wallet/WalletActivityTracker.java:153: Exclude configuration changes from the newly enabled logout hook
  Recreating the only visible activity for a configuration change stops the old instance before starting its replacement, temporarily reducing visibleActivityCount to zero. This newly added call then sets appWentBackground=true and, when automatic logout is enabled with a zero-minute timeout, broadcasts FORCE_FINISH_ACTION. AddressBookActivity is a concrete affected screen: its manifest permits rotation and does not handle orientation changes, and it inherits LockScreenActivity through AbstractBindServiceActivity. Rotating that screen now makes AutoLogout.shouldLogout() return true, so the replacement activity displays the lock screen even though the user never left the wallet. Keep the activity counter balanced, but exclude activity.isChangingConfigurations() from the logout decision. Add lifecycle regression coverage for both configuration recreation and genuine backgrounding.
```
</details>

wallet/src/de/schildbach/wallet/WalletActivityTracker.java
HashEngineering: Confirmed and fixed in `c77b5e86e`. Verified each link before changing anything:

- `AddressBookActivity` declares `configChanges="keyboard|keyboardHidden"` — not orientation — and sets no `screenOrientation`, so a rotation does recreate it.
- Chain confirmed: `AbstractBindServiceActivity : LockScreenActivity()`.
- `AutoLogout.shouldLogout()` (AutoLogout.java:106) returns `appWentBackground` outright when `autoLogoutMinutes == 0`, so no timer needs to elapse.

Worth recording that the hazard arrived WITH the fix: `onStoppedLast()` had been dead code since before this class existed — the override of `onActivityStarted`/`onActivityStopped` never calls `super`, so `ActivitiesTracker.numStarted` never advanced and the base never reached it. `091c2bae3` woke the hook, and that woke this.

It is also broader than one screen: nine activities in `AndroidManifest.xml` neither pin portrait nor list orientation in `configChanges` — `MainActivity`, `AddressBookActivity`, `NetworkMonitorActivity` and `BlockInfoActivity` among them.

Guarded on `activity.isChangingConfigurations()`, with the counter still decrementing so counting stays balanced. One deliberate step beyond the suggestion: the guard also covers `AppForegroundMonitor.noteBackground()` on the same line. A configuration change is not backgrounding for that consumer either, and letting the signal flap fires a spurious foreground edge into `SdkBindRetryService.noteAppForeground()` — harmless but wrong, and it would have outlived the logout symptom.

Three lifecycle tests as requested: genuine backgrounding fires the hook and clears the foreground signal; configuration recreation fires neither, and a later genuine background still fires exactly once; auto-logout disabled still runs the hook but broadcasts nothing. Mutation-verified — removing the guard fails `configurationChange_doesNotFireTheLastStoppedHook` on `setAppWentBackground(any()) should not be called`.

Suites green: wallet 1968, common 90, coinbase 9. Not device-verified — the rotation itself is not covered by an emulator scenario.

thepastaclaw (approved):
## Re-review — Final validation — Phase 2 only (queue backlog)

The complete PR diff contains no substantiated in-scope defects requiring changes, and all five prior findings are addressed at the current head. The fixes preserve configuration-change lifecycle behavior, retry boundary persistence independently of cutover success, and align the emulator assertions and diagnostics with the current implementation. Shell syntax validation and git diff --check passed; Android builds, unit suites, and device scenarios were not rerun during this verification.

🔴 0 blocking | 🟡 0 suggestion(s) | 💬 0 nitpick(s)

<details>

Source: reviewer 1: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 2: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `normal` by `gpt-6-astra` (effort low) — The diff is large and intricate but changes engine lifecycle, cutover readiness gates, foreground tracking, diagnostics, UI formatting, and build behavior rather than directly changing funds movement, key handling, cryptography, consensus, network deserialization, or storage migrations.
- Phase 1 reviewers: **not run (skipped for throughput: 43 PRs queued, above the 10 limit)**
- Fresh final gate: an independent Phase-2 review ran after iterative findings were reconciled
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort high); agent `phase2-reviewer`

</details>

<details>

```
These findings are from an automated code review. Verify the current code and confirm that no unresolved issues remain.

No unresolved findings remain from the prior review on this head.
```
</details>

abaranouski (approved):
Approve as well

PR #1557, opened 2026-09-08
fix(dash-dex): Max enters the sendable amount, not the gross balance

HashEngineering:
Fixes [MO-1036](https://dashpay.atlassian.net/browse/MO-1036). Reported by QA on MO-997 (comment 91128): *"sometimes max show unreal value, that exceed the real balance"*.

## Problem

On the Dash DEX **sell** enter-amount screen, **Max** enters the wallet's **gross balance**. A max sell is carried out as a sweep and the miner fee comes out of the sweep's single output, so what the swap can actually deliver is `balance − fee`. Max entered a figure the swap would never use, and the quote and preview then showed a different, lower number than the one the user just entered.

The buy flow isn't involved — `DEXEnterAmountScreen` passes `showMaxButton = false`.

## Root cause

Three Max paths in `ConvertViewViewModel` each read `wallet.getBalance(ESTIMATED)` independently — `getMaxAmount()`, `selectMaxAmount()`, and the hand-typed-max fallback in `continueSwap()` — and `checkEnteredAmountValue()` bounded entry by the same gross figure, so an unsendable typed amount was only discovered at the preview.

The wallet's own send screen never makes this mistake: it caps against `observeBalance(ESTIMATED, MaxOutputAmountCoinSelector())` (`SendCoinsViewModel.kt:160`). Dash DEX bypassed that entirely.

**The backends already had it right.** `SwapKitApiAggregator.getSwapInfo` ignores the passed amount for a `maximum` request and re-derives the post-fee sweep output — it has to, since NEAR Intents refuses any deposit below the quoted amount (#1519 / MO-984). Only the UI disagreed.

## Fix

One max-sendable figure in `ConvertViewViewModel`, driving every Max path:

- `maxSendableAmount()` computes it from `estimateNetworkFee(currentReceiveAddress(), balance, emptyWallet = true).amountToSend` — deliberately the same call `getSwapInfo` makes, so entry, quote and deposit finally agree on one number. Cached against the balance it was computed from and mutex-serialized, since the estimate builds and signs a candidate sweep (which derives the wallet key).
- `getMaxAmount()` / `selectMaxAmount()` take that figure instead of reading the balance themselves; a failed estimate stays distinct from a genuine zero so a transient failure can't wipe a pinned Max.
- `checkEnteredAmountValue()` bounds entry by it.
- `continueSwap()`'s typed-max fallback compares against it, so a max typed by hand still sets `SwapRequest.maximum`.
- The sell direction pre-warms the estimate (so Max is instant) and treats a wallet whose balance the fee eats as empty for this flow.

On screen: Max is async against the existing busy state rather than a dead tap, a restored Max re-pins a freshly recomputed value, and the "more than max" message names the sendable limit instead of the balance the check just rejected.

**Sweep semantics are unchanged** — the send still empties the wallet; only the entered and displayed amount changes.

## Worth a look in review

- **`estimateNetworkFee` runs on the caller's dispatcher.** Unlike its neighbours in `SendCoinsTaskRunner`, it has no `withContext` of its own (`SendCoinsTaskRunner.kt:151`) and it derives the wallet key via scrypt. Every call site here is a Main-dispatched coroutine, so the estimate is wrapped in `Dispatchers.IO`. Other callers may have the same exposure — out of scope here, but worth someone's time.
- **The preview needed no change.** `MayaConversionPreviewFragment` reads `SwapTradeUIModel.amount`, not the entered amount. On the SwapKit/NEAR route that was already the sweep output; the inflated Total came only from the Maya-native route, where `MayaWebApi.getSwapInfo` passes `swapRequest.amount` straight through. Correcting entry corrects that Total too.
- **CoinJoin needs no special-casing** — with CoinJoin on, `createSendRequest` selects with `CoinJoinCoinSelector` (mixed-only), so the estimate already follows the same rules the real send will.
- **`MaxOutputAmountCoinSelector` would not have matched the quote** — it prices at `Transaction.DEFAULT_TX_FEE` (10000 duff/kB) vs `Constants.ECONOMIC_FEE` (1000), 10× apart. That's why this uses the `SendPaymentService` route.
- **The exact intermittency isn't reproduced.** Both the "Balance:" row and Max read ESTIMATED, so Max equalled the displayed balance rather than visibly exceeding it. The gap is always present but small (~0.00002 DASH for a one-input sweep), growing with UTXO count — the leading explanation for "sometimes", but unconfirmed.

## Testing

- `:integrations:maya:compileDebugKotlin`, `:wallet:compile_testNet3DebugKotlin`, `:wallet:compile_testNet3DebugJavaWithJavac` (forces Dagger component generation, confirming the new `SendPaymentService` injection resolves) — all pass.
- `:integrations:maya:testDebugUnitTest` — 94 tests, 0 failures. New `MaxSendableTest` adds 13: fee below / equal to / above balance, empty wallet, negative estimator output, a Max entered in fiat and read back as DASH, the pin restoring the exact figure, and a hand-typed max still flagged as a maximum.
- `:integrations:maya:ktlintCheck` — clean.

**Not yet done, needs a device and testnet funds:** tap Max with a fully confirmed balance, and again with an unconfirmed incoming tx pending — in both cases the entered amount must not exceed what the wallet can send. Then complete one NEAR-route sell with Max and confirm it isn't refunded (MO-984 regression check).

To be merged into the kotlin-sdk branches later.

## Summary by CodeRabbit

* **Bug Fixes**
  * Max conversion amounts now account for network fees, helping prevent attempts to send more than the wallet can deliver.
  * Maximum amounts are estimated asynchronously and remain accurate after balance or quote recalculations.
  * Improved handling for insufficient balances and cases where fees consume the entire available balance.
  * Maximum-amount errors now display the wallet’s sendable amount rather than its total balance.

* **Tests**
  * Added coverage for fee calculations, zero or negative sendable amounts, currency precision, and maximum-amount selection.

### DIFF
--- a/integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/MayaConvertCryptoFragment.kt
+++ b/integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/MayaConvertCryptoFragment.kt
@@ -94,6 +94,12 @@ class MayaConvertCryptoFragment : Fragment() {
         }
     private var canContinue: Boolean = false
 
+    // The screen's busy state has two independent sources — a quote in flight and the Max
+    // sendable estimate — so each is tracked separately and combined by [updateProcessing];
+    // mirroring one straight into uiState let the other clear it.
+    private var quoteInFlight: Boolean = false
+    private var maxEstimateInFlight: Boolean = false
+
     // Hard gate on Get quote independent of the entered value — false when the wallet has no
     // DASH to convert, so the button stays disabled no matter what the user types. Derived
     // from the ViewModel so the gate survives a configuration change.
@@ -232,8 +238,8 @@ class MayaConvertCryptoFragment : Fragment() {
         // While the quote is being fetched, block all amount input so a late key press
         // can't alter the value carried to the preview.
         viewModel.showLoading.observe(viewLifecycleOwner) { loading ->
-            uiState = uiState.copy(isProcessing = loading == true)
-            updateContinueEnabled()
+            quoteInFlight = loading == true
+            updateProcessing()
         }
 
         viewModel.swapTradeOrder.observe(viewLifecycleOwner) { swapTrade ->
@@ -243,11 +249,11 @@ class MayaConvertCryptoFragment : Fragment() {
                 // navigating. Keep Get quote disabled for that whole stretch (it happens in the
                 // same frame as the showLoading=false observer, so the button never flashes
                 // enabled); re-enable only on the error paths that keep the user on this screen.
-                uiState = uiState.copy(isProcessing = true)
-                updateContinueEnabled()
+                quoteInFlight = true
+                updateProcessing()
                 fun failToRetry() {
-                    uiState = uiState.copy(isProcessing = false)
-                    updateContinueEnabled()
+                    quoteInFlight = false
+                    updateProcessing()
                 }
 
                 val dashInbound = try {
@@ -394,10 +400,15 @@ class MayaConvertCryptoFragment : Fragment() {
             )
             applyNewValue(convertViewModel.enteredConvertAmount, pickedCurrencyType, isLocalized = true)
             // The value re-applied above comes from the formatted display string, which for fiat
-            // is rounded to the currency's 2 decimals — so a restored Max needs the exact balance
-            // pinned back on, taken from the balance as it stands now.
+            // is rounded to the currency's 2 decimals — so a restored Max needs the exact figure
+            // pinned back on. selectMaxAmount re-derives it from the balance as it stands now, so
+            // a balance that moved while this screen was away is picked up rather than restored.
             if (maxAmountSelected) {
-                convertViewModel.selectMaxAmount(pickedCurrencyType)
+                val type = pickedCurrencyType
+                lifecycleScope.launch {
+                    convertViewModel.selectMaxAmount(type)
+                    showPinnedMaxAmount(type)
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


thepastaclaw:
⛔ Final review complete — 2 blocking finding(s) (commit 1c5f160c) · triage: critical · Phase 2 only (queue backlog)

integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/MayaConvertCryptoFragment.kt
```diff
@@ -421,19 +432,47 @@ class MayaConvertCryptoFragment : Fragment() {
...
+        lifecycleScope.launch {
+            val maxAmount = try {
+                convertViewModel.getMaxAmount()
+            } finally {
+                maxEstimateInFlight = false
+                updateProcessing()
+            }
+
+            if (maxAmount == null) {
+                // Nothing left once the fee is taken out — say so rather than entering zero.
```
thepastaclaw, line 453: **🔴 Blocking: Keep Continue disabled until asynchronous Max selection finishes**

The `finally` block clears `maxEstimateInFlight` before `applyNewValue()` and the subsequent suspending `selectMaxAmount()` call execute. `applyNewValue()` can make `canContinue` true, so Continue may become enabled while `maxAmountSelected` is still false and the displayed, rounded amount has not yet been pinned to the exact sendable DASH value. A tap during this window can create a normal partial `SwapRequest` rather than a maximum request, particularly affecting the sweep-sensitive NEAR route. The restored-Max path in `resetViewSelection()` has the same race because it launches `selectMaxAmount()` without setting the busy flag. Keep the in-flight state set through amount application and exact pinning, and clear it only after `selectMaxAmount()` completes.

integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/convert_currency/ConvertViewViewModel.kt
```diff
@@ -451,32 +478,114 @@ class ConvertViewViewModel @Inject constructor(
...
+     * deliverable amount is `balance − fee`, not the balance. This is deliberately the same call
+     * `SwapKitApiAggregator.getSwapInfo` makes to quote a maximum, so entry, quote and deposit all
+     * agree on one number; reading the gross `Wallet.BalanceType.ESTIMATED` here instead is what
+     * made Max show an amount the swap would never use (MO-997).
+     *
+     * Estimation is not cheap — it builds and signs a candidate sweep, which derives the wallet
+     * key — so the result is cached against the balance it was computed from and recomputed only
+     * once that balance moves. Serialized so a pre-warm and a Max tap can't both pay for it.
+     */
+    suspend fun maxSendableAmount(): Coin = estimateMaxSendable() ?: Coin.ZERO
```
thepastaclaw, line 496: **🔴 Blocking: Do not turn fee-estimation failures into a zero maximum**

`estimateMaxSendable()` returns `null` when wallet loading or sweep construction fails, but `maxSendableAmount()` converts that failure to `Coin.ZERO`. As a result, `checkEnteredAmountValue()` treats any positive DASH entry as exceeding a zero sendable balance and reports `NotEnoughBalance` or `MoreThanMax`. The Max action also maps the same transient failure to the insufficient-balance message, while `lastMaxSendableAmount` can still contain the previous successful estimate. Preserve estimation-unavailable separately from a confirmed zero result, and make normal validation and Max handling retryable when the estimate is unavailable rather than validating against a synthetic zero.

thepastaclaw (changes requested):

The PR correctly changes sell Max to use the fee-adjusted sweep amount, but two asynchronous failure paths remain. Continue can be enabled before the exact maximum is pinned, and estimation failures are converted into a synthetic zero balance, causing valid amounts to be rejected and obscuring retryable failures.

🔴 2 blocking

<details>

- Triage: `critical` by `gpt-6-astra` (effort low) — The substantial change alters ConvertViewViewModel's max-sendable calculation and amount validation for DEX sweeps, directly affecting coin selection and funds movement, including fee estimation, wallet-key-derived transaction construction, and CoinJoin handling.
- Phase 1 reviewers: **not run (skipped for throughput: 14 PRs queued, above the 10 limit)**
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort xhigh); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort xhigh); agent `phase2-reviewer`

</details>

<details>

```

In `integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/MayaConvertCryptoFragment.kt`:
- [BLOCKING] integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/MayaConvertCryptoFragment.kt:447-453: Keep Continue disabled until asynchronous Max selection finishes
  The `finally` block clears `maxEstimateInFlight` before `applyNewValue()` and the subsequent suspending `selectMaxAmount()` call execute. `applyNewValue()` can make `canContinue` true, so Continue may become enabled while `maxAmountSelected` is still false and the displayed, rounded amount has not yet been pinned to the exact sendable DASH value. A tap during this window can create a normal partial `SwapRequest` rather than a maximum request, particularly affecting the sweep-sensitive NEAR route. The restored-Max path in `resetViewSelection()` has the same race because it launches `selectMaxAmount()` without setting the busy flag. Keep the in-flight state set through amount application and exact pinning, and clear it only after `selectMaxAmount()` completes.

In `integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/convert_currency/ConvertViewViewModel.kt`:
- [BLOCKING] integrations/maya/src/main/java/org/dash/wallet/integrations/maya/ui/convert_currency/ConvertViewViewModel.kt:496: Do not turn fee-estimation failures into a zero maximum
  `estimateMaxSendable()` returns `null` when wallet loading or sweep construction fails, but `maxSendableAmount()` converts that failure to `Coin.ZERO`. As a result, `checkEnteredAmountValue()` treats any positive DASH entry as exceeding a zero sendable balance and reports `NotEnoughBalance` or `MoreThanMax`. The Max action also maps the same transient failure to the insufficient-balance message, while `lastMaxSendableAmount` can still contain the previous successful estimate. Preserve estimation-unavailable separately from a confirmed zero result, and make normal validation and Max handling retryable when the estimate is unavailable rather than validating against a synthetic zero.
```
</details>

PR #1558, opened 2026-09-08
fix(coinbase): correct amount entry, auth persistence and API amount format

HashEngineering:
Jira: [MO-1028](https://dashpay.atlassian.net/browse/MO-1028) — split from MO-995 comment 91129.

Everything here is **pre-existing on `master`**, not a regression from the v11.9.0 merge. QA found it on a release build of a feature branch; the responsible code was identical on master.

## 1. Transfer screen — four QA symptoms, one root cause

| # | Reported | Symptom |
| --- | --- | --- |
| A | MO-995 c91098 #1 | Tapping the fiat option crashes the screen |
| B | MO-995 c91129 #3 | After fiat → dash → fiat the amount can't be changed; number keys dead, backspace works |
| C | MO-995 c91129 #4 | With a non-USD local currency (BYN) the fiat option shows USD on first open |
| D | MO-995 c91129 #5 | Small "USD" label next to an amount actually in the local currency |

`EnterAmountToTransferFragment` built the currency picker once in `onViewCreated` from a synchronous read of `localCurrencyCode`, which is only ever assigned asynchronously from `WalletUIConfig` — so it kept the seeded `USD` default (**C**). That stale *label* was then passed to `applyNewValue` as the formatting mode, which branched on `localCurrencyCode == monetaryCode`. A stale `"USD"` never matched a real `"BYN"`, so fiat amounts took the DASH branch: DASH-scale digits under a USD label (**D**), a keypad gated at 2 decimals that refused every digit (**B**), and span bounds from the other branch giving `from > to`, so `Spannable.setSpan` threw (**A**).

Fix — address the cause, not just the crash:
- `applyNewValue` branches on `isFiatSelected`, never on a currency label.
- The selected currency is observed; the picker and rendered amount rebuild on every emission.
- The text, its currency-label span, and the keypad's decimal limit are produced together by one pure function (`TransferAmountFormat.kt`), so they cannot describe different amounts. A bounds check degrades to an unstyled amount rather than crashing — as a belt, not the fix.

## 2. "I need to link Coinbase every time" (c91129 #2)

- `TokenAuthenticator` cleared **both** tokens on any non-success refresh, so one timeout or 5xx silently de-authenticated the user. Now clears only on a definitive OAuth2 rejection of the grant (400/401), never on transport failure. Also drops two `println` calls that printed tokens to logcat.
- `CoinBaseRepository.isAuthenticated` was a non-volatile `var` seeded `false`, written only from a `Dispatchers.IO` collector. Its four synchronous readers could read `false` before the first emission and without a memory barrier might never observe the write. Now `StateFlow`-backed, with a flow for reactive consumers and a suspending authoritative read; all four readers react to the emission.

## 3. Found while fixing

- **Locale-formatted amounts sent to the Coinbase API — live, money-moving.** `buyDash()` formatted the deposit and buy-order amounts with `Constants.SEND_PAYMENT_LOCAL_FORMAT` (`withLocale(deviceLocale)`), rendering `"12,34"` on every comma-decimal locale — including **be_BY, the tester's own** — straight into a bank deposit body and the buy-order `quote_size`. The buy button is not hidden, so this ships. Now uses a locale-independent `CoinbaseConstants.API_AMOUNT_FORMAT`. Audited every other wire amount: `SendTransactionToWalletParams.amount` comes from `Coin.toPlainString()`, verified locale-independent.
- **Decimal separator mixing.** The keypad inserted `'.'` while the gate counted with the locale separator, so on comma-decimal locales the limit never fired, `"0.0"` collapsed to `"0"`, and MAX in DASH mode produced `"1,5"` leaving the transfer button permanently disabled. `'.'` is now canonical for everything stored, counted, gated or parsed. Resolves the two standing `// TODO`s in `applyCoinbaseExchangeRate`.
- **`GenericUtils.formatFiatWithoutComma` mis-read grouped amounts.** Its rule was "contains a dot and a comma → strip the commas", turning `"1.234,56"` into `"1.23456"` — a 1000× error. The decimal mark is now decided, not guessed: it occurs at most once, so the last separator is the decimal mark precisely when that character appears once. Ambiguous single-separator input keeps its existing reading, so output is byte-identical for every currently reachable input.
- **Convert screen** (unreachable behind `convertBtn.isVisible = false`) gets the same span fix, and `toDashValue` no longer has identical `if`/`else` branches that scaled a local-currency amount by the crypto rate.

## Verification

- `:integrations:coinbase:compileDebugKotlin`, `:integrations:coinbase:testDebugUnitTest`, `:common:testDebugUnitTest`, `:wallet:compile_testNet3DebugKotlin` — all pass, re-run after rebasing onto `520555d37`.
- **56 unit tests, up from 7.** Covers span bounds, mode selection, decimal gating, BYN with a stale/absent currency emission, the fiat → dash → fiat round trip, conversions finer than 2 decimals, comma-decimal locales, separator normalisation, exchange-rate semantics, and wire-format locale independence.
- ktlint is not applied to `coinbase`/`common`/`wallet` (only exploredash, maya, uphold, crowdnode), so `ktlintFormat` is a no-op here; the 120-char limit was enforced by hand.

**Not verified on device.** QA should re-check A–D and the re-link behaviour on a release build with a non-USD, comma-decimal locale — BYN reproduces both classes.

## Reviewer note

The first commit, `fix(common): repair two compile errors that block every module`, is **unrelated** to MO-1028 — `master` did not compile (duplicate `Configuration` import in `MenuItem.kt`; `ActionItem.kt` missing the `subtitleMaxLines` three maya call sites pass). It's kept separate so it can be dropped or picked independently. An equivalent `ActionItem`/`MenuItem` fix is currently sitting uncommitted on `master` from parallel work — functionally identical, differing only in parameter position and KDoc. **If that lands on master first, drop commit 1 from this branch.**

## Summary by CodeRabbit

- **New Features**
  - Action items can now display subtitles across multiple lines before truncating.

- **Bug Fixes**
  - Improved Coinbase amount formatting across regional locales.
  - Corrected fiat and cryptocurrency conversion calculations.
  - Prevented amount-display crashes when converted values use different decimal lengths.
  - Coinbase authentication status now updates reliably after startup and connection changes.
  - Preserved credentials during temporary network failures while clearing them after rejected authentication.
  - Improved handling of decimal limits and partial amount entry.

### DIFF
--- a/common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
+++ b/common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
@@ -72,28 +72,55 @@ object GenericUtils {
         return Locale.US
     }
 
+    /** Characters any supported locale may use as a decimal or grouping mark. */
+    private const val SEPARATORS = ".,٫"
+
     /**
-     * To perform some operations on our fiat values (ex: parse to double, convert fiat to Coin), it needs to be properly formatted
-     * In case our fiat value is in a currency that has a comma, we need to strip it away so as to have our value as a decimal
-     * @param fiatValue
-     * @return
+     * Normalises a human-formatted amount into one BigDecimal, Coin.parseCoin and
+     * Fiat.parseFiat will accept: a single '.' decimal mark and no grouping marks.
+     *
+     * Which mark is the decimal one is decided, never guessed. The decimal mark can occur
+     * at most once in a number, so the LAST separator is the decimal mark precisely when
+     * that character appears exactly once; every other separator is grouping and is
+     * dropped. When a character repeats it cannot be a decimal mark, so all its
+     * occurrences are grouping.
+     *
+     * This used to test only "contains a dot AND a comma" and then strip the commas, which
+     * silently mangled every European-formatted amount: "1.234,56" became "1.23456", a
+     * thousandfold error in a value on its way to being parsed and sent. A single
+     * separator is genuinely ambiguous ("1,234" is 1.234 in one locale and 1234 in
+     * another), so that case is left reading as a decimal mark, exactly as before.
      */
     fun formatFiatWithoutComma(fiatValue: String): String {
-        val fiatValueContainsCommaWithDecimal = fiatValue.contains(".") && fiatValue.contains(",")
+        val cleaned = normalizeSeparators(fiatValue)
 
-        val cleaned = if (fiatValueContainsCommaWithDecimal) {
-            fiatValue.replace(",", "")
-        } else {
-            fiatValue.replace(",", ".")
-                .replace("٫", ".")
-        }
-        
         // Limit to 8 decimal places to prevent rounding errors in Coin.parseCoin
         val decimalIndex = cleaned.indexOf('.')
         return if (decimalIndex != -1 && cleaned.length > decimalIndex + 9) {
             cleaned.substring(0, decimalIndex + 9)
         } else {
             cleaned
+        }
+    }
+
+    private fun normalizeSeparators(value: String): String {
+        val lastSeparator = value.indexOfLast { it in SEPARATORS }
+
+        if (lastSeparator < 0) {
+            return value
+        }
+
+        // A repeated character cannot be the decimal mark, so it is grouping throughout.
+        val decimalIndex = if (value.count { it == value[lastSeparator] } == 1) lastSeparator else -1
+
+        return buildString(value.length) {
+            value.forEachIndexed { index, character ->
+                when {
+                    index == decimalIndex -> append('.')
+                    character in SEPARATORS -> Unit // grouping mark: drop it
+                    else -> append(character)
+                }
+            }
         }
     }
 
--- a/common/src/main/java/org/dash/wallet/common/ui/components/ActionItem.kt
+++ b/common/src/main/java/org/dash/wallet/common/ui/components/ActionItem.kt
@@ -92,6 +92,11 @@ fun ActionItem(
     /** One-line secondary text under [title] (`Action item / Var 3`); null renders `Var 1`. */
     subtitle: String? = null,
     /**
+     * Lines [subtitle] may wrap to before it ellipsizes. Defaults to the one-line contract
+     * above; ignored when [subtitleMiddleEllipsis] is set, which is single-line by nature.
+     */
+    subtitleMaxLines: Int = 1,
+    /**
      * Truncates [subtitle] from the middle to fit the available width (e.g. for addresses,
      * where both the start and end need to stay checkable) instead of the standard end-ellipsis.
      * Width-measured so it stays correct at any font scale, unlike a fixed character count.
@@ -167,7 +172,7 @@ fun ActionItem(
                         text = it,
                         style = MyTheme.Typography.Footnote,
                         color = colors.textSecondary,
-                        maxLines = 1,
+                        maxLines = subtitleMaxLines,
                         overflow = TextOverflow.Ellipsis,
                         modifier = Modifier.fillMaxWidth()
                     )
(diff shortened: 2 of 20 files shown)


thepastaclaw:
⛔ Final review complete — 2 blocking finding(s) (commit c25f5a86) · triage: normal · Phase 2 only (queue backlog)

common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt
```diff
@@ -72,22 +72,28 @@ object GenericUtils {
         return Locale.US
     }
 
+    /** Characters any supported locale may use as a decimal or grouping mark. */
+    private const val SEPARATORS = ".,٫"
```
thepastaclaw, line 76: **🔴 Blocking: Arabic thousands separator is not normalized**

`SEPARATORS` includes the Arabic decimal separator `٫` (U+066B) but omits the Arabic grouping separator `٬` (U+066C). An Arabic-formatted value such as `1٬234٫56` is consequently normalized to `1٬234.56`; the grouping mark remains in the result, so subsequent `BigDecimal` or `Fiat.parseFiat` parsing fails or falls back to zero. Treat U+066C as a grouping separator and add a regression test for grouped Arabic input.

```suggestion
private const val SEPARATORS = ".,٫٬"
```

thepastaclaw (changes requested):

The PR fixes the reported Coinbase amount-entry, authentication persistence, and locale-independent API formatting issues, but two correctness bugs remain in the new implementation. Balance updates can overwrite a newer authentication state, and Arabic-formatted grouped amounts are not fully normalized before parsing.

🔴 2 blocking

### 1 finding(s) not shown inline (the lines are not part of this PR's diff)

**🔴 Blocking: Balance updates can restore stale authentication state**
`integrations/coinbase/src/main/java/org/dash/wallet/integrations/coinbase/viewmodels/CoinbaseServicesViewModel.kt:84-97`

The balance pipeline copies `uiState.value` before `flatMapLatest` waits for the selected-currency exchange rate, then assigns that captured state wholesale in the terminal collector. If authentication changes while the rate request is pending, the authentication collector updates `_uiState` to the new value, but this collector later executes `_uiState.value = state` and restores the old `isLoggedIn` value. The Coinbase UI can therefore show the user as logged out again after a successful authentication update. Update only the balance fields from the balance pipeline using the current state.

```suggestion
_uiState.update { current ->
    current.copy(
        balance = state.balance,
        balanceFiat = state.balanceFiat
    )
}
```

<details>

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a large, cross-cutting Coinbase UI, authentication, formatting, and repository logic change with substantial tests, but it does not directly modify consensus, cryptography, key handling, storage migrations, or funds-movement execution.
- Phase 1 reviewers: **not run (skipped for throughput: 15 PRs queued, above the 10 limit)**

</details>

<details>

```

In `integrations/coinbase/src/main/java/org/dash/wallet/integrations/coinbase/viewmodels/CoinbaseServicesViewModel.kt`:
- [BLOCKING] integrations/coinbase/src/main/java/org/dash/wallet/integrations/coinbase/viewmodels/CoinbaseServicesViewModel.kt:84-97: Balance updates can restore stale authentication state
  The balance pipeline copies `uiState.value` before `flatMapLatest` waits for the selected-currency exchange rate, then assigns that captured state wholesale in the terminal collector. If authentication changes while the rate request is pending, the authentication collector updates `_uiState` to the new value, but this collector later executes `_uiState.value = state` and restores the old `isLoggedIn` value. The Coinbase UI can therefore show the user as logged out again after a successful authentication update. Update only the balance fields from the balance pipeline using the current state.

In `common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt`:
- [BLOCKING] common/src/main/java/org/dash/wallet/common/util/GenericUtils.kt:75-76: Arabic thousands separator is not normalized
  `SEPARATORS` includes the Arabic decimal separator `٫` (U+066B) but omits the Arabic grouping separator `٬` (U+066C). An Arabic-formatted value such as `1٬234٫56` is consequently normalized to `1٬234.56`; the grouping mark remains in the result, so subsequent `BigDecimal` or `Fiat.parseFiat` parsing fails or falls back to zero. Treat U+066C as a grouping separator and add a regression test for grouped Arabic input.
```
</details>

PR #1559, opened 2026-09-08
fix: restored-wallet balance and history — ordered DashPay bring-up before SPV, walker mirror guard, onboarding crash fixes

HashEngineering:
## Issue being fixed
Restored wallets with DashPay contacts lost contact-payment coins and history: the DIP-15 receival accounts registered only after the filter scan had passed their funding heights, and the host-side backfill gate then forced a redundant rewind (321k blocks on a large CoinJoin testnet wallet). The transaction-list reattribution walker could also stamp corrections from an incomplete TXO mirror.

Stacked on #1555 (`fix/sdk-spv-restart-on-service-restart`).

## What was done
- `L1ShadowSyncService` calls the SDK's ordered bring-up (`PlatformWalletManager.startWalletSubsystems`: identity → contacts → contact-account drain) immediately before `startSpv`, best-effort. The app-side `DashPayBackfillGate` is provided as a no-op — the SDK marks contacts covered at `synced_height = 0` (dashpay/platform#4475).
- `SdkTxStoreWalker`: guard durable reattribution stamps behind an address-level TXO-completeness check.
- Onboarding crash fixes.

## SDK dependency
Requires a Kotlin SDK AAR that exposes `startWalletSubsystems` and carries the platform-side fixes: dashpay/platform `integration/v42int11` (mirrored at `HashEngineering/platform`, tip e31cc30fed) = v4.2-dev + #4439 + the #4474/#4475 restore commits + two record-fold fixes, on the rust-dashcore#979 engine. Until that AAR is published, `dashSdkVersion` in `build.gradle` must point at a local `-SNAPSHOT` build — hence **draft**.

## Verified
Large CoinJoin testnet wallet (7,300 txs, 1,062 coins) fresh restore against AAR `0.1.0-v42int14-SNAPSHOT`: balance exact to the duff vs dashj, 0 missing coins, 0 CoinJoin mixing records mis-netted, sync 4.5 min from birth height, no rescan rewind. 181 wallet unit tests green (sync service, cutover, bind retry, walker).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved DashPay wallet startup by initializing identities, contacts, and contact accounts before synchronization.
  - Synchronization now continues if DashPay initialization encounters an error.
  - Prevented transaction corrections from being saved while wallet output data is incomplete; corrections retry when data becomes available.
  - Preserved correct handling of outgoing transactions involving external addresses and wallets.
[...]

PR #1564, opened 2026-09-10
fix(dashpay): contested-username duplicate registration, CBOR identity crash, and five live-QA defects (MO-973)

HashEngineering:
Seven fixes to the DashPay username flow, from MO-973 field reports and live device testing. They share the username/identity path but are independent defects — the summary below is ordered by severity.

## 1. A contested name was registered twice, paying the vote-poll prefund again

`recoverUsernames` resolves a name through the DPNS **unique** index, and for a contested label that index is not awarded until masternodes resolve the vote. So the name is on chain but invisible, `RestoreIdentityWorker` reads `currentUsername == null` as "never registered", and re-registers — paying the **0.20 DASH prefunded specialized balance a second time**.

Reproduced on demand (emulator, testnet). The explorer accounts for the money exactly:

```
25,000,000,000 funded − 343,199,040 gas − 4,656,800,960 balance
= 20,000,000,000 credits = 0.20 DASH — the vote-poll prefund, paid once
```

The FFI then rejected it with `Insufficient identity balance 4680452100 required 20000100000`. **That error is the symptom of the duplicate, not an under-funding problem.** The retry can never fail cheaply either: DPNS is two documents, and the preorder is keyed on a hash of a *fresh random salt*, so it never collides and always succeeds (~23.8M credits) before the domain step dies. Three triggers burned three preorders for a name already owned.

Fixed by asking before registering — one targeted `getVoteContenders`, using `getVoteContendersOrThrow` so a failed read cannot masquerade as "never requested" and re-trigger the duplicate.

## 2. A CBOR crash aborted every identity read

`IdentityPublicKey.toObject()` in dashj-platform 4.0.0 put the raw `ContractBounds` object into the map handed to `Cbor`, which has no converter for it:

```
4.0.0  objMap.put("contractBounds", this)
4.0.1  objMap.put("contractBounds", this.toObject())
```

`Cbor.kt` is byte-identical across versions — the encoder was never the bug. Taking `4.0.1-SNAPSHOT` cures **every** call path, including `PlatformRepo.recoverIdentityAsync`, which a wallet-side tolerance does not cover.

> **Note on the snapshot:** review correctly flagged that `prod` should use an immutable release. `4.0.1` is not published to Maven Central, so there is currently no alternative. `prod` should be pinned as soon as it publishes.

## 3. The instant username could be the contested one again

The secondary screen pre-fills the input with the contested primary and expects a suffix, but only enforced `startsWith(primary)` — nothing required anything after it, the suffix-character rule beneath it is commented out, and the clear button restored the bare primary.

The availability query cannot catch it either, for the same reason as #1: a contested name is absent from the unique index while voting, so the primary reads as "not taken". Compared on the **DPNS-normalized** label, so `asdo`/`asd0` collide as one name.

Not a contestability-predicate change — there is one predicate across all call sites, and the denomination charged was correct.

## 4. "Insufficient funds" after a successful transfer

`insufficientFunds` was the first branch of the hint/error `when`, gated on nothing, while every sibling branch was keyed on `submitState` and `canContinue` already greyed the button out with exactly that condition. After a transfer broadcasts, the spent output leaves the CONFIRMED balance while the change is unconfirmed, so the comparison flips true for a transfer that already succeeded. Both shielded screens had it.

## 5. "Network syncing 100%" while Tools says synced

`percentageSync` was preserved on phases with no trustworthy scan position but `syncStage` was written unconditionally, so the two disagreed. The engine never latches SYNCED — each new block bumps the targets — so a synced wallet flips SYNCED → FILTERS → SYNCED, the stage regressed to BLOCKS while the percent stayed 100, and the pair rendered as "syncing 100%". 59 such blips in one field log.

Deliberately narrow: a fault and a cold reconnect must still read OFFLINE, and an established stage is held only when the scan has genuinely caught up — never seeded as OFFLINE at process start.

## 6. No completion card when a contested name wins

A contested-only creation showed no card at any point: none during voting (deliberate — nothing is usable yet), and none on winning, because both win paths wrote `DONE_AND_DISMISS` and skipped `DONE`. So the user waits out the vote — 90 min on testnet, two weeks on mainnet — wins, and is never told.

## 7. The confirm dialog hijacked the instant-username screen

`checkViewConfirmDialog()` had a fourth trigger that fired off **observed state** rather than user action, whenever a `usernameVerified` flag was seen true. That flag is sticky, lives on an `activityViewModels` view model, and `Flow.observe` is `repeatOnLifecycle(STARTED) { collect }` — which replays the StateFlow on every re-subscribe. So a later screen in the flow received a stale true and raised the dialog by itself, on the instant-username step, unbidden.

It was also redundant: the verify path it existed to resume already opens the dialog from `VerifyIdentityFragment` before navigating back. Removing it leaves three triggers, all user actions — the Continue button, the skip-verify branch, and verify itself.

The screen also latched `isFocusable = false` with nothing anywhere setting it back, so once it fired the field was dead and back was the only way out. Now derived from the state each emission.

Pre-existing, and not from the contested-username work.

> The first attempt at this (`28e740fc0`, was 3ab86557a) made the flag one-shot instead of removing the trigger. That only moved the symptom — dialog after typing rather than on arrival — and QA caught it. `b7e09a95a` supersedes it.

## Verification

- Full `:wallet` suite green — **2011 tests**, ktlint clean
- Device (emulator, testnet): dual contested creation end to end — identity, both DPNS names, restore worker, TARGETED scan issuing **one** query, `VOTING`, `Worker result SUCCESS` (previously always `FAILED`), vote resolved in our favour ~91 min later, identity balance unchanged throughout
- Mutation-verified where a test asserts something *did not* happen

## Rebased onto `feat/kotlin-sdk-phase-1`

#1559 and #1555 merged (squashed) into `feat/kotlin-sdk-phase-1` on 2026-09-19 and their branches were deleted, so GitHub retargeted this PR there. Because those merges were squashes, this branch briefly showed 51 commits / 55 files of base-line history that phase-1 already holds.

Rebased onto `1e7b2c37e` the same day. The old base `c3948c75c` and the phase-1 head have byte-identical trees, so the replay was conflict-free and the result is byte-identical to the pre-rebase head: **8 commits, 20 files, +876 / −35**.

The earlier CI blocker (`Unresolved reference 'startWalletSubsystems'`) is gone — it was the base pinning an SDK older than its own source, fixed upstream by committing `0.1.0-v42int19-SNAPSHOT`. This branch compiles with **no local SDK override**.

## Summary by CodeRabbit

- **New Features**
  - Added validation preventing instant usernames from matching contested usernames.
  - Completion cards now remain visible after winning a contested username vote.
  - Added clearer messaging when an instant username matches the contested name.

- **Bug Fixes**
  - Prevented identity restoration failures caused by incompatible cached data or cancellation.
  - Insufficient-funds messages no longer appear after a transfer is broadcast.
  - Prevented duplicate contested-username registration attempts.
  - Improved blockchain sync status during catch-up and restarts.
  - Prevented outdated username availability results from overwriting current input.
  - Corrected username input locking and confirmation behavior during submission.

### DIFF
--- a/wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
+++ b/wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
@@ -278,10 +278,20 @@ class RestoreIdentityWorker @AssistedInject constructor(
                         // catch below stamps a generic error — NOT "missing domain
                         // document", so the tile shows the retry card, and restoring
                         // stays true so a retry re-runs THIS worker, never a re-fund).
+                        // MO-973: `error(...)` throws WITHOUT a cause, so the
+                        // engine's own message never reached the log — the
+                        // field report showed only our classifier's label and
+                        // the real reason was unrecoverable. Carry the cause.
                         is SdkWriteResult.NotBroadcast ->
-                            error("username registration did not complete (retryable): ${result.reason}")
+                            throw IllegalStateException(
+                                "username registration did not complete (retryable): ${result.reason}",
+                                result.cause
+                            )
                         is SdkWriteResult.Ambiguous ->
-                            error("username registration outcome unconfirmed (retryable): ${result.cause.message}")
+                            throw IllegalStateException(
+                                "username registration outcome unconfirmed (retryable): ${result.cause.message}",
+                                result.cause
+                            )
                     }
                 }
             }
--- a/integrations/coinbase/src/main/java/org/dash/wallet/integrations/coinbase/ui/EnterAmountToTransferFragment.kt
+++ b/integrations/coinbase/src/main/java/org/dash/wallet/integrations/coinbase/ui/EnterAmountToTransferFragment.kt
@@ -58,11 +58,13 @@ import androidx.compose.ui.unit.dp
 import org.dash.wallet.common.ui.components.DashWalletTheme
 import org.dash.wallet.common.ui.components.LocalDashColors
 import org.dash.wallet.common.ui.segmented_picker.SegmentedOption
+import org.slf4j.LoggerFactory
 
 @AndroidEntryPoint
 class EnterAmountToTransferFragment : Fragment(R.layout.enter_amount_to_transfer_fragment) {
 
     companion object {
+        private val log = LoggerFactory.getLogger(EnterAmountToTransferFragment::class.java)
         private const val DECIMAL_SEPARATOR = '.'
         fun newInstance() = EnterAmountToTransferFragment()
     }
@@ -162,15 +164,18 @@ class EnterAmountToTransferFragment : Fragment(R.layout.enter_amount_to_transfer
 
     private fun formatTransferredAmount(value: String) {
         val text = viewModel.applyNewValue(value, pickedCurrencyOption)
+        val span = amountCurrencySpan(
+            text = text,
+            isDashSelected = pickedCurrencyIndex == 0,
+            formattedValue = viewModel.formattedValue,
+            fiatBalance = viewModel.fiatBalance,
+            isCurrencyFirst = viewModel.fiatAmount?.isCurrencyFirst() == true
+        )
         val spannableString = SpannableString(text).apply {
-            if (pickedCurrencyIndex == 0) {
-                spanAmount(this, viewModel.formattedValue.length, text.length)
+            if (span != null) {
+                spanAmount(this, span.from, span.to)
             } else {
-                if (viewModel.fiatAmount?.isCurrencyFirst() == true && text.length - viewModel.fiatBalance.length > 0) {
-                    spanAmount(this, 0, text.length - viewModel.fiatBalance.length)
-                } else {
-                    spanAmount(this, viewModel.inputValue.length, text.length)
-                }
+                log.warn("no valid currency span for amount text of length {}", text.length)
             }
         }
 
@@ -285,4 +290,50 @@ class EnterAmountToTransferFragment : Fragment(R.layout.enter_amount_to_transfer
             }
         }
     }
+}
+
+/** The slice of the amount text that renders in the smaller currency style. */
+internal data class AmountCurrencySpan(val from: Int, val to: Int)
+
+/**
+ * Which slice of [text] is the currency label (rendered smaller) rather than the
+ * amount itself, or `null` when there is nothing valid to style.
+ *
+ * Pure and host-testable on purpose. The three branches take their bounds from
+ * separate mutable ViewModel fields, and `applyNewValue` refreshes only the one
+ * belonging to the branch it took — `formattedValue` for DASH, `fiatBalance` for
+ * fiat. Mixing a field from the other branch produces bounds that do not describe
+ * [text] at all, and `Spannable.setSpan` throws on those rather than ignoring them.
+ *
+ * MO-995: the fiat currency-last branch used the raw `inputValue` length. The fiat
+ * branch of `applyNewValue` sets `fiatBalance = inputValue` only when the input has
+ * at most 2 decimals; otherwise `fiatBalance` is the 2-dp formatted figure, which is
+ * SHORTER. Tapping USD converts the DASH amount through `formatInput`, yielding many
+ * decimals — so `from` (12) overshot `to` (7) and setSpan crashed the screen.
+ */
+internal fun amountCurrencySpan(
+    text: String,
+    isDashSelected: Boolean,
+    formattedValue: String,
+    fiatBalance: String,
+    isCurrencyFirst: Boolean
+): AmountCurrencySpan? {
+    val from: Int
+    val to: Int
+    if (isDashSelected) {
+        // text == "$formattedValue $DASH" — the trailing code.
+        from = formattedValue.length
+        to = text.length
+    } else if (isCurrencyFirst && text.length - fiatBalance.length > 0) {
+        // text == "$symbol $fiatBalance" — the leading symbol.
+        from = 0
+        to = text.length - fiatBalance.length
+    } else {
+        // text == "$fiatBalance $symbol" — the trailing symbol.
+        from = fiatBalance.length
+        to = text.length
+    }
+    // A stale or mismatched field must degrade to an unstyled amount, never crash.
+    if (from < 0 || to > text.length || from >= to) return null
+    return AmountCurrencySpan(from, to)
 }
--- a/integrations/coinbase/src/test/java/org/dash/wallet/integrations/coinbase/AmountCurrencySpanTest.kt
+++ b/integrations/coinbase/src/test/java/org/dash/wallet/integrations/coinbase/AmountCurrencySpanTest.kt
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2026 Dash Core Group.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.dash.wallet.integrations.coinbase
+
+import org.dash.wallet.integrations.coinbase.ui.AmountCurrencySpan
+import org.dash.wallet.integrations.coinbase.ui.amountCurrencySpan
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Test
+
+/**
+ * MO-995: tapping the fiat option on the Coinbase transfer screen crashed with
+ * `IndexOutOfBoundsException: setSpan (12 ... 7) has end before start`.
+ *
+ * The span bounds are computed from separate mutable ViewModel fields and handed
+ * straight to `Spannable.setSpan`, which throws on a reversed or out-of-bounds
+ * range. These tests pin the bounds for each branch and the degrade-to-null
+ * behaviour that keeps a field mismatch off the crash path.
+ */
+class AmountCurrencySpanTest {
+
+    @Test
+    fun dashSelected_spansTheTrailingCurrencyCode() {
+        // applyNewValue returns "$formattedValue $monetaryCode".
+        val span = amountCurrencySpan(
+            text = "1.5 DASH",
+            isDashSelected = true,
+            formattedValue = "1.5",
+            fiatBalance = "irrelevant on this branch",
+            isCurrencyFirst = false
+        )
+
+        assertEquals(AmountCurrencySpan(3, 8), span)
+    }
+
+    @Test
+    fun fiatCurrencyFirst_spansTheLeadingSymbol() {
+        // applyNewValue returns "$symbol $fiatBalance".
+        val span = amountCurrencySpan(
+            text = "$ 4.52",
+            isDashSelected = false,
+            formattedValue = "stale",
+            fiatBalance = "4.52",
+            isCurrencyFirst = true
+        )
+
+        assertEquals(AmountCurrencySpan(0, 2), span)
+    }
+
+    @Test
+    fun fiatCurrencyLast_spansTheTrailingSymbol() {
+        // applyNewValue returns "$fiatBalance $symbol".
+        val span = amountCurrencySpan(
+            text = "4.52 Kč",
+            isDashSelected = false,
+            formattedValue = "stale",
+            fiatBalance = "4.52",
+            isCurrencyFirst = false
+        )
+
+        assertEquals(AmountCurrencySpan(4, 7), span)
+    }
+
+    /**
+     * The regression itself, with the field lengths from the crash log.
+     *
+     * Switching DASH -> USD converts the amount via `formatInput`, so `inputValue`
+     * becomes a many-decimal figure (12 chars) while `fiatBalance` is the 2-dp
+     * formatted one. Using `inputValue.length` gave from=12 against a 7-char text.
+     * The span must be bounded by `fiatBalance` instead, and must stay in range.
+     */
+    @Test
+    fun fiatCurrencyLast_manyDecimalInput_staysInRange() {
+        val text = "4.52 Kč" // length 7, as in the crash
+        val span = amountCurrencySpan(
+            text = text,
+            isDashSelected = false,
+            formattedValue = "stale",
+            fiatBalance = "4.52",
+            isCurrencyFirst = false
+        )
+
+        requireNotNull(span)
+        assertEquals(4, span.from)
+        assertEquals(7, span.to)
+        // The exact invariant Spannable.setSpan enforces.
+        assert(span.from in 0..span.to && span.to <= text.length)
+    }
+
+    @Test
+    fun reversedBounds_degradeToNullInsteadOfCrashing() {
+        // fiatBalance longer than the whole text — the shape that used to throw.
+        val span = amountCurrencySpan(
+            text = "4.52 Kč",
+            isDashSelected = false,
+            formattedValue = "stale",
+            fiatBalance = "123456789.12",
+            isCurrencyFirst = false
+        )
+
+        assertNull(span)
+    }
+
+    @Test
+    fun emptyRange_degradesToNull() {
+        // Nothing to style: the amount fills the whole text.
+        val span = amountCurrencySpan(
+            text = "1.5",
+            isDashSelected = true,
+            formattedValue = "1.5",
+            fiatBalance = "",
+            isCurrencyFirst = false
+        )
+
+        assertNull(span)
+    }
+
+    @Test
+    fun currencyFirstWithNoRoomForASymbol_fallsBackAndDegradesToNull() {
+        // text.length - fiatBalance.length == 0 sends this down the currency-last
+        // branch, where from == to; that must be null, not a zero-width span.
+        val span = amountCurrencySpan(
+            text = "4.52",
+            isDashSelected = false,
+            formattedValue = "stale",
+            fiatBalance = "4.52",
+            isCurrencyFirst = true
+        )
+
+        assertNull(span)
+    }
+}
--- a/integrations/uphold/build.gradle
+++ b/integrations/uphold/build.gradle
@@ -19,11 +19,28 @@ android {
     buildTypes {
         configureEach {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            // Shipped to consumers (the app) and applied when THEY minify — this
+            // is where this module's keep rules belong.
             consumerProguardFiles 'proguard-rules.pro'
         }
-        release {
-            minifyEnabled true
-        }
+        // NO minifyEnabled here. MO-995: this was the only module in the project
+        // that minified its own library output, and R8 stripped Hilt's
+        // `hilt_aggregated_deps.*` metadata out of the release classes jar
+        // (measured: 5 entries in debug, 0 in release). That metadata is how the
+        // APP module's Hilt processor discovers @AndroidEntryPoint classes in
+        // library modules, so the generated FragmentCImpl did not implement
+        // UpholdPortalFragment_GeneratedInjector and attaching the fragment
+        // crashed with a ClassCastException — release builds only, which is why
+        // it never appeared in development.
+        //
+        // The existing `-keep class org.dash.wallet.integrations.uphold.**` rule
+        // does not help: hilt_aggregated_deps is a different package.
+        //
+        // Dropping it costs nothing. The app module already sets
+        // minifyEnabled true (wallet/build.gradle, release), so every class here
+        // is still shrunk and obfuscated in the final APK — once, at the app
+        // level, with the whole program visible. Minifying a library separately
+        // only removes metadata its consumers still need.
     }
 
     compileOptions {
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -218,8 +218,24 @@ require_state() {
 }
 
 lock_screen()   { adbs input keyevent 26; sleep 2; }   # power -> screen off, keyguard on
-wake_unlock()   { adbs input keyevent 26; sleep 1; adbs input keyevent 82; sleep 1
-                  adbs input text "$PIN"; adbs input keyevent 66; sleep 3; }
+# WAKE ONLY — never `keyevent 26` (power), which TOGGLES: on an already-unlocked
+# device it turns the screen OFF and re-locks it, which silently re-created the
+# Keystore device-locked denial in the middle of a scenario that needs a working
+# bind. Also note `input text` does NOT reach the keyguard PIN pad (digits need
+# keyevents 8-11); scripted PIN entry proved unreliable, so an actually-locked
+# device must be unlocked by hand before running the unlocked scenarios.
+# Keep the screen on for the whole run: the scenarios poll for up to 45s and a
+# screen-off re-locks the device, which re-creates the Keystore denial in the
+# middle of a scenario that needs a working bind.
+keep_awake()    { adbs svc power stayon true >/dev/null 2>&1
+                  adbs settings put system screen_off_timeout 1800000 >/dev/null 2>&1; }
+
+wake_unlock()   { keep_awake; adbs input keyevent 224; sleep 2
+                  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
+                    note "WARNING: device still shows a keyguard — unlock it by hand, then re-run."
+                    note "         a locked device denies the master-alias keystore op and this"
+                    note "         scenario needs a WORKING bind."
+                  fi; }
 launch_app()    { adbs monkey -p "$PKG" -c android.intent.category.LAUNCHER 1 >/dev/null 2>&1; sleep 12; }
 start_service() { adbs am start-foreground-service -n "$SVC" >/dev/null 2>&1; sleep 12; }
 
@@ -326,6 +342,8 @@ s3)
   say "S3 — denied, then healed (recovery path; expected to expose the noteAppForeground gap)"
   require_device; require_root
   note "assumes S2 just ran: state DUAL_RUNNING, bind marker unset"
+  show_state
+  mark_log          # S3 had NO mark, so its assertions matched the whole log history
   wake_unlock
   launch_app
   assert_log "bind succeeded after unlock"       "app wallet (bound to new|already bound to) SDK wallet"
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -36,7 +36,7 @@ set -uo pipefail
 PKG=hashengineering.darkcoin.wallet_test
 SVC=$PKG/de.schildbach.wallet.service.BlockchainServiceImpl
 AVD=Pixel_9_API_36_AOPS
-PIN=1234
+PIN=REDACTED
 PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
 BT=~/Library/Android/sdk/build-tools/35.0.1
 APK_DIR=wallet/build/outputs/apk/_testNet3/release
@@ -230,12 +230,37 @@ lock_screen()   { adbs input keyevent 26; sleep 2; }   # power -> screen off, ke
 keep_awake()    { adbs svc power stayon true >/dev/null 2>&1
                   adbs settings put system screen_off_timeout 1800000 >/dev/null 2>&1; }
 
-wake_unlock()   { keep_awake; adbs input keyevent 224; sleep 2
-                  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
-                    note "WARNING: device still shows a keyguard — unlock it by hand, then re-run."
-                    note "         a locked device denies the master-alias keystore op and this"
-                    note "         scenario needs a WORKING bind."
-                  fi; }
+# Enter $PIN on the keyguard using DIGIT KEYEVENTS. `input text` does not reach
+# the keyguard PIN pad — that is why earlier scripted unlocks silently failed and
+# left the device locked, which then denied the master-alias keystore op inside
+# scenarios that needed a working bind. KEYCODE_0 is 7, so digit d -> 7+d.
+pin_keyevents() {
+  local i ch
+  for (( i=0; i<${#PIN}; i++ )); do
+    ch=${PIN:$i:1}
+    adbs input keyevent $((7 + ch))
+  done
+  adbs input keyevent 66   # ENTER
+}
+
+# Wake and, if a keyguard is up, dismiss it. Never `keyevent 26` (power), which
+# TOGGLES and would turn the screen OFF on an already-unlocked device.
+wake_unlock() {
+  keep_awake
+  adbs input keyevent 224; sleep 2
+  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
+    adbs input swipe 540 2000 540 600 300; sleep 2
+    pin_keyevents; sleep 4
+  fi
+  if adbs dumpsys window 2>/dev/null | grep -q "mDreamingLockscreen=true"; then
+    note "WARNING: keyguard still up after entering \$PIN — unlock by hand, then re-run."
+    note "         a locked device denies the master-alias keystore op and this"
+    note "         scenario needs a WORKING bind."
+  else
+    note "device unlocked"
+  fi
+}
+
 launch_app()    { adbs monkey -p "$PKG" -c android.intent.category.LAUNCHER 1 >/dev/null 2>&1; sleep 12; }
 start_service() { adbs am start-foreground-service -n "$SVC" >/dev/null 2>&1; sleep 12; }
 
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -41,11 +41,47 @@ AVD=Pixel_9_API_36_AOPS
 # NEVER hardcode it — supply it per run:  EMULATOR_PIN=... ./this-script s3
 # Only the scenarios that lock the screen need it (s2, and s3's unlock).
 PIN="${EMULATOR_PIN:-}"
-PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
+# MUST match CutoverCoordinator.FIRST_CUTOVER_VERSION_CODE. The cutover ships
+# in 12.0.0, so the boundary is 12000000 — it MOVED here from 11100000 when the
+# release slipped a line. Two copies of a boundary is one copy too many: this
+# used to be hardcoded again below, where a device last run on 11.10–11.25 (which
+# the APP counts as pre-cutover) was judged post-cutover by the script, so the
+# script overwrote a genuine last_version instead of leaving it alone.
+FIRST_CUTOVER_VERSION_CODE=12000000
+PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE
 BT=~/Library/Android/sdk/build-tools/35.0.1
 APK_DIR=wallet/build/outputs/apk/_testNet3/release
 OUT=/tmp/mo995-emulator
 mkdir -p "$OUT"
+
+# Assertion ledger. Every FAIL bumps this; the EXIT trap turns a non-zero
+# count into a non-zero exit status.
+#
+# WHY NOT `exit 1` AT THE FAILING ASSERTION: each scenario is a SEQUENCE of
+# assertions and the diagnostic value is in seeing all of them. S3 asserting
+# "bind failed" then "dashj came back" then "state not committed" tells you
+# WHERE the chain broke; aborting at the first FAIL reports only that it broke.
+# So assertions still print and continue — but the script no longer exits 0
+# after printing FAIL, which would let an automated run call a failed cutover
+# scenario a success.
+FAILURES=0
+
+fail() {
+  FAILURES=$((FAILURES + 1))
+  printf '   \033[31mFAIL\033[0m %s\n' "$1"
+  [ $# -gt 1 ] && printf '        %s\n' "$2"
+  return 0
+}
+
+on_exit() {
+  local rc=$?
+  if [ "$FAILURES" -gt 0 ]; then
+    printf '\n\033[31m%s assertion(s) FAILED\033[0m\n' "$FAILURES"
+    exit 1
+  fi
+  exit "$rc"
+}
+trap on_exit EXIT
 
 say()  { printf '\n\033[1m== %s\033[0m\n' "$*"; }
 note() { printf '   %s\n' "$*"; }
@@ -129,7 +165,7 @@ assert_log() {
     fi
     sleep 5; waited=$((waited + 5))
   done
-  printf '   \033[31mFAIL\033[0m %s\n        (no match in %ss for: %s)\n' "$label" "$budget" "$pat"
+  fail "$label" "(no match in ${budget}s for: $pat)"
 }
 
 # refute_log <label> <grep-pattern>   — PASS if absent from the new lines
@@ -137,7 +173,7 @@ refute_log() {
   local label="$1" pat="$2"
   local w; w=$(since_mark)
   if grep -qE "$pat" "$w"; then
-    printf '   \033[31mFAIL\033[0m %s\n        (unexpected: %s)\n' "$label" "$pat"
+    fail "$label" "(unexpected: $pat)"
     grep -E "$pat" "$w" | head -2 | sed 's/^/          /'
   else printf '   \033[32mPASS\033[0m %s\n' "$label"; fi
 }
@@ -152,7 +188,7 @@ assert_not_committed() {
   local label="$1" raw
   raw=$(adbs "strings /data/data/$PKG/files/datastore/dashpay.preferences_pb 2>/dev/null" | tr -d '\r')
   if echo "$raw" | grep -qE "CUT_OVER|SETTLED"; then
-    printf '   \033[31mFAIL\033[0m %s\n        (persisted state is committed)\n' "$label"
+    fail "$label" "(persisted state is committed)"
   else
     printf '   \033[32mPASS\033[0m %s\n' "$label"
   fi
@@ -161,7 +197,7 @@ assert_not_committed() {
 fake_pre_cutover_previous_launch() {
   # Configuration.lastVersionCode is read from the default SharedPreferences
   # ("last_version", Configuration.java:61) at construction. Writing it here is
-  # exactly equivalent to "the previous launch ran a pre-11.10 build", which is
+  # exactly equivalent to "the previous launch ran a pre-cutover build", which is
   # what CutoverCoordinator's GATE 1 tests — without needing a real v11.9.0 APK
   # signed with matching keys.
   #
@@ -175,13 +211,13 @@ fake_pre_cutover_previous_launch() {
     note "prefs file not present yet — launch the app once, complete onboarding, then re-run"
     return 1
   fi
-  # If a REAL upgrade was just installed over a pre-11.10 build, lastVersionCode
+  # If a REAL upgrade was just installed over a pre-cutover build, lastVersionCode
   # already holds the genuine value — leave it alone. Faking it would only
   # overwrite the truth with the same thing, and hide a mismatch if the real
   # upgrade did not record what we expect.
   local cur
   cur=$(adbs "grep -o 'last_version\" value=\"[0-9]*' $f" 2>/dev/null | grep -o '[0-9]*$' | tr -d '\r')
-  if [ -n "$cur" ] && [ "$cur" -gt 0 ] && [ "$cur" -lt 11100000 ]; then
+  if [ -n "$cur" ] && [ "$cur" -gt 0 ] && [ "$cur" -lt "$FIRST_CUTOVER_VERSION_CODE" ]; then
     note "REAL pre-cutover upgrade detected (last_version=$cur) — not faking it"
     return 0
   fi
@@ -461,7 +497,7 @@ s4)
   if [ "${COUNT:-0}" -ge 2 ]; then
     printf '   \033[32mPASS\033[0m engine restarted after the teardown (%s starts)\n' "$COUNT"
   else
-    printf '   \033[31mFAIL\033[0m engine did NOT restart — MO-995 latch still present (%s starts)\n' "$COUNT"
+    fail "engine did NOT restart — MO-995 latch still present (${COUNT:-0} starts)"
   fi
   ;;
 
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -28,7 +28,8 @@
 #         ./scripts/cutover-emulator-test.sh setup      # build+sign+upgrade-install
 #         ./scripts/cutover-emulator-test.sh s1        # upgrade, bind works
 #         ./scripts/cutover-emulator-test.sh s2        # upgrade, bind denied
-#         ./scripts/cutover-emulator-test.sh s3        # denied then healed
+#         ./scripts/cutover-emulator-test.sh s3        # denied then healed in-session
+#         ./scripts/cutover-emulator-test.sh s3b       # ...and the next launch commits
 #         ./scripts/cutover-emulator-test.sh s4        # trim -> SPV restart
 #         ./scripts/cutover-emulator-test.sh log       # pull + tail wallet.log
 set -uo pipefail
@@ -136,6 +137,22 @@ refute_log() {
     printf '   \033[31mFAIL\033[0m %s\n        (unexpected: %s)\n' "$label" "$pat"
     grep -E "$pat" "$w" | head -2 | sed 's/^/          /'
   else printf '   \033[32mPASS\033[0m %s\n' "$label"; fi
+}
+
+# assert_not_committed <label> — reads the PERSISTED state, not the log.
+# The engine-gate lines (`dashjEngineMayStart=`, `starting peergroup`) are only
+# emitted at service onCreate. A scenario that deliberately keeps ONE process
+# alive has no new onCreate, so those lines cannot appear in its window and
+# asserting on them is a guaranteed false FAIL. The DataStore is the durable
+# source of truth for who owns L1.
+assert_not_committed() {
+  local label="$1" raw
+  raw=$(adbs "strings /data/data/$PKG/files/datastore/dashpay.preferences_pb 2>/dev/null" | tr -d '\r')
+  if echo "$raw" | grep -qE "CUT_OVER|SETTLED"; then
+    printf '   \033[31mFAIL\033[0m %s\n        (persisted state is committed)\n' "$label"
+  else
+    printf '   \033[32mPASS\033[0m %s\n' "$label"
+  fi
 }
 
 fake_pre_cutover_previous_launch() {
@@ -364,18 +381,49 @@ s2)
   ;;
 
 s3)
-  say "S3 — denied, then healed (recovery path; expected to expose the noteAppForeground gap)"
+  say "S3 — denied, then healed IN-SESSION (the walletB recovery path)"
   require_device; require_root
-  note "assumes S2 just ran: state DUAL_RUNNING, bind marker unset"
+  note "assumes S2 just ran: state DUAL_RUNNING, bind marker unset, screen locked"
   show_state
-  mark_log          # S3 had NO mark, so its assertions matched the whole log history
+  mark_log
   wake_unlock
   launch_app
-  assert_log "bind succeeded after unlock"       "app wallet (bound to new|already bound to) SDK wallet"
-  note "does it recover WITHOUT a restart? (this is the open question)"
-  assert_log "committed in-session"              "cutover state DUAL_RUNNING -> CUT_OVER"
-  note "if the line above FAILED, restart the app and re-check — needing a restart confirms"
-  note "that noteAppForeground() only resets a backoff nothing reads once the wait loop ended"
+
+  # The recovery chain, in order. Each of these was broken until b49ef25b0:
+  #   - ProcessLifecycleOwner never fired, so the foreground signal never existed
+  #   - noteAppForeground was state-only, so nothing drove a retry
+  #   - nothing else calls maybeRetry once the cutover correctly declines to commit
+  assert_log "app-foreground signal fired"       "App moved to foreground"
+  assert_log "foreground drove a bind retry"     "app foregrounded with an SDK bind retry pending"
+  assert_log "the retry ran a pass"              "SDK bind retry [0-9]+ \(app foreground\)"
+  assert_log "bind healed after the unlock"      "app wallet (bound to new|already bound to) SDK wallet"
+  assert_log "retry pressure cleared"            "bind established after [0-9]+ failed pass"
+
+  # …and it must be the SAME process: a restart would heal it trivially and prove
+  # nothing. walletB restarted repeatedly and never recovered.
+  refute_log "healed WITHOUT a process restart"  "WalletApplication.onCreate\(\)"
+
+  # dashj must still own L1 for this launch. The commit is NOT expected here and
+  # asserting it was my error: the upgrade seam only runs at process start, so
+  # in-session the only route is the readiness-gated auto-commit, which needs
+  # MIN_PARITY_STREAK readings at a 10s throttle AND the SDK scan caught up to
+  # tip. 45s cannot satisfy that, so a missing commit here is correct deferral,
+  # not a defect. The commit is checked on the NEXT launch instead — see s3b.
+  assert_not_committed "dashj still owns L1 (state not committed)"
+  note "commit is deliberately NOT asserted here — run s3b to check the next launch"
+  ;;
+
+s3b)
+  say "S3b — the launch AFTER an in-session heal should commit"
+  require_device; require_root
+  show_state
+  note "the bind marker must be set by now (s3 healed it); the seam can act at process start"
+  adbs am force-stop "$PKG"; sleep 3
+  mark_log
+  wake_unlock
+  launch_app
+  assert_log "committed on the next launch"      "cutover state DUAL_RUNNING -> CUT_OVER|READY_OBSERVED -> CUT_OVER"
+  assert_log "SDK L1 engine started"             "L1 shadow SPV started"
   ;;
 
 s4)
--- a/scripts/cutover-emulator-test.sh
+++ b/scripts/cutover-emulator-test.sh
@@ -37,7 +37,10 @@ set -uo pipefail
 PKG=hashengineering.darkcoin.wallet_test
 SVC=$PKG/de.schildbach.wallet.service.BlockchainServiceImpl
 AVD=Pixel_9_API_36_AOPS
-PIN=REDACTED
+# Device lock PIN, for the scenarios that must lock and unlock the emulator.
+# NEVER hardcode it — supply it per run:  EMULATOR_PIN=... ./this-script s3
+# Only the scenarios that lock the screen need it (s2, and s3's unlock).
+PIN="${EMULATOR_PIN:-}"
 PREV_VERSION_PRE_CUTOVER=11090000   # any value < FIRST_CUTOVER_VERSION_CODE (11100000)
 BT=~/Library/Android/sdk/build-tools/35.0.1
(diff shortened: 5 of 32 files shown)


HashEngineering:
@coderabbitai review

wallet/src/de/schildbach/wallet/service/platform/sdk/SdkBlockchainStateMapper.kt
```diff
@@ -177,6 +182,42 @@ internal fun sdkSyncStage(phase: ShadowSyncPhase): SyncStage = when (phase) {
...
+ *   `derive_noHeightKnowledgePreservesHeightAndDate`).
+ * - a genuine re-scan caps at 99 and still moves the stage normally.
+ * - SYNCED always reports COMPLETE outright; reaching the terminal state is never
+ *   suppressed.
+ *
+ * Pure — host-testable.
+ */
+internal fun sdkSyncStageOrPreserve(phase: ShadowSyncPhase, scanPercent: Int): SyncStage? = when {
+    phase == ShadowSyncPhase.SYNCED -> SyncStage.COMPLETE
+    scanPercent >= 100 -> null
```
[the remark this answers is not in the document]
HashEngineering: Valid — and a real regression I introduced in 363c5adfa, not a style point. Fixed in ec33150e1.

Confirmed the mechanism before changing anything: `syncStageFlow` is `MutableStateFlow<SyncStage?>(null)` and `getSyncStage()` returns `syncStageFlow.value ?: SyncStage.OFFLINE`, so a first update that preserved left the flow null and the app read OFFLINE. A cold start on an already-synced wallet — first snapshot caught up, engine not yet latched to SYNCED — hit exactly that, and the reset at `:212` sets the flow back to null so it could recur mid-session.

Root cause was my design: a pure mapper cannot know whether a prior stage exists, so "hold the established stage" and "there is nothing to hold" collapsed into the same `null`. That decision moved to the writer:

```kotlin
if (!update.preserveEstablishedSyncStage || syncStageFlow.value == null) {
    syncStageFlow.value = update.syncStage
}
```

`syncStage` is non-null again and always carries the phase's own stage — the value to write, and the seed when nothing is established. `sdkSyncStageOrPreserve` became `mayPreserveEstablishedSyncStage`, which says only that an established stage *may* be held. The narrowing is unchanged: `shadowSyncPercent` reads 0 for IDLE/CONNECTING/ERROR so they can never be held, and SYNCED is never held.

Added the provider-level regression test you asked for, plus its opposite so both directions are pinned: one asserts a caught-up blip does not regress an established stage, one asserts a caught-up *first* update seeds the stage instead of reporting OFFLINE.

wallet/src/de/schildbach/wallet/ui/username/request/RequestUsernameFragment.kt
```diff
@@ -395,8 +395,17 @@ open class RequestUsernameFragment : Fragment(R.layout.fragment_request_username
...
                 // check was made fail-closed.
-                binding.usernameAvailableContainer.isVisible = it.usernameCheckFailed
-                if (it.usernameCheckFailed) {
+                // Same fail-closed surface for the instant name that is really
+                // the contested one again (MO-973 report 3): the button is already
+                // disabled, but silence would read as "still typing" when the
+                // screen pre-filled the primary and the user pressed on.
+                binding.usernameAvailableContainer.isVisible =
+                    it.usernameCheckFailed || it.secondaryNameSameAsPrimary
+                if (it.secondaryNameSameAsPrimary) {
```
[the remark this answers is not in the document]
HashEngineering: Valid, and fixed in ec33150e1 — this one was reproduced independently by QA while the review was open, which is about as strong a confirmation as a race gets.

MO-973 comment 91244, testing 12000012: *"I was still able to request an instance username when the request was contested. When I quickly added 2-9 digit and removed it? No information that any instance name was created."* The "quickly" is the timing window you identified; the missing information is the consequence — the request proceeds, DPNS rejects the duplicate, and nothing surfaces it.

Confirmed the path: `usernameSubmitButtonState` takes no collision input and for a `Secondary` enables on `!usernameExists && !usernameContestable`, so a stale in-flight result for the *suffixed* name enables submit while the field holds the bare primary.

Fixed by making the flag part of the success branch's ENTRY condition rather than gating inside it, so no stale result can route around it and the else-path holds the button disabled with the reason:

```kotlin
if (it.usernameCharactersValid && it.usernameLengthValid && it.usernameCheckSuccess &&
    !it.secondaryNameSameAsPrimary
) {
```

Regression test added. One caveat stated plainly: the device build now carries the fix and the sequence no longer reproduces there, but a timing window failing to reproduce is weaker evidence than the unit test, so I am leaning on the latter.

build.gradle
```diff
@@ -9,7 +9,12 @@ buildscript {
         // DASH_SDK_VERSION BuildConfig field shown on the About screen once the
         // cutover hands L1 ownership to the SDK.
         dashSdkVersion = '0.1.0-v42int5'
-        dppVersion = "4.0.0"
+        // 4.0.1 fixes the CBOR crash that aborted every identity read:
+        // IdentityPublicKey.toObject() put the raw ContractBounds object
+        // (SingleContractDocumentType) into the map handed to Cbor, which has
+        // no converter for it -> IllegalArgumentException("No converter for ...").
+        // 4.0.1 calls this.toObject() first. Cbor.kt itself is unchanged.
+        dppVersion = "4.0.1-SNAPSHOT"
```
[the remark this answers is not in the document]
HashEngineering: Right in principle, but not actionable today — and it caught a wrong claim of mine, which is worth recording.

`4.0.1` is **not published to Maven Central**. The snapshot is currently the only way to get the CBOR fix, so there is no immutable alternative to switch `prod` to. Keeping `dppVersion = "4.0.1-SNAPSHOT"`.

The part you were right to probe: my commit message on 49b7c0995 justified the snapshot by claiming the seams need `NativeStateTransition` and the native `toNative()` plumbing that only the snapshot carries. That was wrong — no wallet source references either symbol, which I verified when checking this finding. Availability is the real and only reason, and ec33150e1's message corrects the record.

Worth being explicit about what is actually being taken on here, since your concern about unconstrained snapshot contents stands: the fix itself is one line in `IdentityPublicKey.toObject()` —

```
4.0.0  objMap.put("contractBounds", this)
4.0.1  objMap.put("contractBounds", this.toObject())
```

4.0.0 put the raw `ContractBounds` object into the map handed to `Cbor`, which has no converter for it. `Cbor.kt` is byte-identical across 4.0.0, 4.0.1 and 4.0.1-SNAPSHOT — the encoder was never the bug. The snapshot additionally differs from the 4.0.1 release in `DapiClient`, `ContactRequests`, `Profiles`, `Document`, `Identity`, `IdentityVerify`, `TxMetadata` and an added `NativeStateTransition.kt`, none of which this change needs.

Pinning `prod` to an immutable release is the right end state; it should happen as soon as 4.0.1 is on Maven Central.

HashEngineering:
@coderabbitai review

All three findings from the previous review are addressed in ec33150e1, with replies on each thread.

Two were real defects in my own earlier commits — the stage-seeding regression (a caught-up first update left `syncStageFlow` null, so `getSyncStage()` reported OFFLINE) and the collision race, which QA reproduced independently in MO-973 comment 91244 while the review was open.

The third (prod on a SNAPSHOT) is correct in principle but has no immutable alternative: 4.0.1 is not published to Maven Central. It also caught a wrong justification in my 49b7c0995 commit message, now corrected.

Since that review, the branch also gained the won-vote completion card fix (7144410ab). Worth a look at the new `preserveEstablishedSyncStage` seam between `SdkBlockchainStateMapper` and `BlockchainStateDataProvider` in particular — the pure/stateful split there is the part I would most want a second opinion on.

wallet/test/de/schildbach/wallet/service/BlockchainStateDataProviderTest.kt
```diff
@@ -232,4 +234,42 @@ class BlockchainStateDataProviderTest {
...
+    @Test
+    fun sdkUpdate_caughtUpBlipDoesNotRegressAnEstablishedStage() {
+        provider.updateSdkBlockchainState(sdkUpdate(syncStage = SyncStage.COMPLETE))
+        awaitUntil("stage established") { provider.getSyncStage() == SyncStage.COMPLETE }
+
+        // The blip: phase says BLOCKS, but the scan has caught up.
+        provider.updateSdkBlockchainState(
+            sdkUpdate(syncStage = SyncStage.BLOCKS, preserveEstablishedSyncStage = true)
+        )
+        Thread.sleep(100)
```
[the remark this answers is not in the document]
HashEngineering: Correct, and a flaw in the test I added — fixed in 2ed95da2d.

You were right that `Thread.sleep(100)` proves nothing: the assertion could be satisfied by the FIRST update's `COMPLETE` before the blip ever ran, so the test could pass without exercising the preservation path at all.

One refinement on the suggested fix. Awaiting the blip's *own* row value is still not quite sufficient — `saveState` happens **before** the stage decision in the same body, so the row can be visible while that decision is pending. Since the scope is `Executors.newSingleThreadExecutor()` running jobs in order, I awaited a **trailing** update's row value instead: once `percentageSync 101` is visible, the blip's body has completed, stage line included.

Mutation-verified both ways — removing the preservation guard from `BlockchainStateDataProvider` fails exactly this test now, which the sleeping version would not reliably have caught.

Full suite green (1992), ktlint clean.

thepastaclaw:
⚠️ DEGRADED — Blockers found — Phase 2 deferred (commit 6cc5108c) · triage: normal · stand-in models (primary models out of quota)
_Canonical validated blockers: 1_

build.gradle
thepastaclaw, line 17: **🔴 Blocking: Production builds resolve a mutable DPP SNAPSHOT dependency**

The `prod` flavor is mapped to `dppVersion` in `wallet/build.gradle`, and this line sets that value to `4.0.1-SNAPSHOT`. As a result, production builds incorporate a mutable artifact whose contents can change between builds; the repository also has no Gradle dependency verification metadata or lockfile constraining the resolved artifact. A compromised Maven repository, publishing credential, or release environment could therefore deliver code inside the production wallet capable of exposing wallet data or altering transaction behavior. Use an immutable released DPP artifact for `prod`, or pin and verify the exact snapshot artifact contents before shipping production builds. The PR's note that 4.0.1 is not yet available explains why the change was made, but does not remove the release-integrity risk.

<sub>source: `gpt-6-astra` (phase2-reviewer: security-auditor)</sub>

thepastaclaw (changes requested):

The PR fixes the reported DashPay username, identity serialization, shielded-transfer, synchronization, completion-card, and confirm-dialog defects with targeted production code and regression coverage. However, production flavor dependencies resolve the mutable 4.0.1-SNAPSHOT DPP artifact without dependency verification metadata, leaving the production APK exposed to supply-chain substitution; this remains a blocking release-integrity issue even though the PR documents that no immutable artifact is currently available.

🔴 1 blocking

<details>

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a large, cross-cutting set of username, identity, sync-state, UI, and dependency changes with substantial tests, but the diff does not directly modify consensus rules, cryptography, key handling, storage migrations, or funds-movement/coin-selection code.
- Phase 1 reviewers: **not run (skipped for throughput: 21 PRs queued, above the 10 limit)**

</details>

<details>

```

In `build.gradle`:
- [BLOCKING] build.gradle:17: Production builds resolve a mutable DPP SNAPSHOT dependency
  The `prod` flavor is mapped to `dppVersion` in `wallet/build.gradle`, and this line sets that value to `4.0.1-SNAPSHOT`. As a result, production builds incorporate a mutable artifact whose contents can change between builds; the repository also has no Gradle dependency verification metadata or lockfile constraining the resolved artifact. A compromised Maven repository, publishing credential, or release environment could therefore deliver code inside the production wallet capable of exposing wallet data or altering transaction behavior. Use an immutable released DPP artifact for `prod`, or pin and verify the exact snapshot artifact contents before shipping production builds. The PR's note that 4.0.1 is not yet available explains why the change was made, but does not remove the release-integrity risk.
```
</details>

build.gradle
HashEngineering: Confirmed on the mechanism, with two corrections to the scope, and I'm leaving this open rather than closing it.

**What is accurate.** `wallet/build.gradle:35` does map `prod : dppVersion`, so `prodImplementation org.dashj.platform:dash-sdk-{java,kotlin,android}` now resolves a mutable snapshot, and there is no `gradle/verification-metadata.xml` or lockfile in the repo constraining it.

**Correction 1 — the prod exposure is not introduced by this line.** Directly above, `wallet/build.gradle:90` declares `implementation "org.dashj:dash-sdk-android:$dashSdkVersion"` unscoped, so it applies to `prod` as well, and on the current base `dashSdkVersion = '0.1.0-v42int19-SNAPSHOT'` (`9b62c7bc0`). A prod build on this branch line already resolves a mutable snapshot artifact regardless of `dppVersion`. Reverting only `dppVersion` would not restore an immutable prod closure.

**Correction 2 — the flavor map's comment is stale, and it pre-dates this PR.** Line 33 says "prod uses the stable release; all other flavors use the latest snapshot", but all four entries point at the same `dppVersion`. `git diff <base>...HEAD -- wallet/build.gradle` is empty — this PR does not touch that file. The collapse is pre-existing.

**Where that leaves it.** This is a v12.0.0 pre-release line under testnet QA; neither snapshot is shippable to prod as-is, so this is a release-gate item for the whole line, not a merge blocker for this PR. The two ways to close it properly:

1. Publish dashj-platform `4.0.1` (and a non-snapshot SDK build) and pin both — the real fix; `4.0.1` is not on Maven Central yet, which is why the snapshot is here.
2. Split the map so `prod : "4.0.0"` and the rest take `dppVersion`, restoring the comment's stated intent.

I have not taken (2) unilaterally: it would make prod the only flavor that still carries the `IdentityPublicKey.toObject()` CBOR crash this PR fixes, so prod and testnet would diverge on identity reads and the divergence would surface late. That trade is a release-policy call for the repo owner, not something to settle in a bugfix PR.

HashEngineering: @thepastaclaw This is not going to production so we can specify -SNAPSHOT temporarily.

thepastaclaw (changes requested):
## Re-review — Final validation — Phase 2 only (queue backlog)

The PR fixes the contested-username, identity serialization, synchronization-stage, shielded-transfer, completion-card, and confirmation-dialog defects with targeted production changes and regression tests. It also introduces a mutable DPP SNAPSHOT into the production dependency path, so production builds are not reproducible or release-integrity controlled until an immutable verified artifact is available.

🔴 1 blocking

_1 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

Source: reviewer 1: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 2: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a broad, cross-cutting DashPay username, identity, sync, shielded-UI, dependency, and test change, but the diff does not directly alter consensus rules, funds movement or coin selection, cryptography, key handling, peer-facing deserialization, or storage migrations.
- Phase 1 reviewers: **not run (skipped for throughput: 29 PRs queued, above the 10 limit)**
- Fresh final gate: an independent Phase-2 review ran after iterative findings were reconciled
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort high); agent `phase2-reviewer`

</details>

<details>

```

In `build.gradle`:
- [BLOCKING] build.gradle:23: Production builds resolve a mutable DPP SNAPSHOT dependency
  (existing thread: https://github.com/dashpay/dash-wallet/pull/1564#discussion_r4010588230)
  `dppVersion` is set to `4.0.1-SNAPSHOT`, and `wallet/build.gradle` maps the `prod` flavor directly to `dppVersion` for the dashj-platform SDK artifacts. Because the repository has no Gradle dependency-verification metadata or lockfile constraining the resolved snapshot, the contents incorporated into a production build can change without a source change. A compromised snapshot repository, publishing credential, or release environment could therefore ship code with access to wallet data or transaction behavior. Use an immutable released DPP artifact for `prod` once available, or pin and cryptographically verify the exact temporary artifact before permitting a production release. The separately existing `dash-sdk-android` snapshot does not eliminate the DPP-specific release-integrity issue introduced by this change.
```
</details>

<details>
<summary>Out-of-scope follow-up suggestions (1)</summary>

These are valid observations, but they are outside this PR's scope and should be handled in separate issues or author/maintainer-requested PRs rather than blocking this review.

- **Production also resolves a mutable Dash SDK snapshot** — `wallet/build.gradle` declares `org.dashj:dash-sdk-android:$dashSdkVersion` through an unscoped `implementation`, and the current base value is also a SNAPSHOT. This is a concrete production release-integrity concern, but it predates this PR and should be tracked as separate release-hardening work rather than used to expand this change.
  - Follow-up: Create a separate release-hardening issue or maintainer-requested change to pin immutable versions and enable Gradle dependency verification for all production dependencies.

</details>

build.gradle
thepastaclaw: **Still applies** (re-reviewed at `f17003ba`): The dependency remains `4.0.1-SNAPSHOT` for `prod`, and no dependency locking or verification was added, so the mutable-artifact release-integrity risk is unchanged at this head. The unavailable immutable release makes this a release gate rather than an actionable code substitution in this PR, but it does not make the finding fixed.

wallet/src/de/schildbach/wallet/ui/username/request/RequestUsernameFragment.kt
```diff
@@ -273,7 +273,17 @@ open class RequestUsernameFragment : Fragment(R.layout.fragment_request_username
...
+            // the disabled branch below: checkUsernameValid stops NEW lookups for a
+            // colliding name, but one already in flight for the suffixed name can
+            // land afterwards and set usernameCheckSuccess. This branch would then
+            // ask usernameSubmitButtonState — which takes no collision input, and
+            // for a Secondary enables on `!usernameExists && !usernameContestable`
+            // — about the STALE name while the field holds the bare primary, and
+            // enable submit for the duplicate.
+            if (it.usernameCharactersValid && it.usernameLengthValid && it.usernameCheckSuccess &&
+                !it.secondaryNameSameAsPrimary
+            ) {
```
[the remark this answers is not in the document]
HashEngineering: Confirmed and fixed in `2d174cff3`.

Verified the mechanism before changing anything: `checkUsername` (`RequestUserNameViewModel.kt:1241`) launches a fresh `viewModelScope.launch` per call and cancels nothing, and the 600 ms debounce in `RequestUsernameFragment.kt:637-645` only calls `removeCallbacks` on a runnable that has not fired yet — once a lookup is away, further typing does not touch it. Both round-trips (`getUsername`, `getVoteContendersOrThrow`) are DAPI calls, so two can be in flight and finish out of order. You are right that the collision check does not cover two different non-colliding names.

**Fix.** A `pendingUsernameCheck` label records what the UI is entitled to show a verdict for. It is written in two places, which matters: `checkUsername` claims it synchronously before the coroutine starts (main thread, so the newest caller wins unambiguously), and `checkUsernameValid` also sets it on every keystroke — otherwise a keystroke that starts no new query, because the new name is invalid, would leave the old lookup free to land.

All three terminal writes consult it, not just the success path. A stale *failure* would otherwise raise the retry banner over a name the user had already moved past.

**On the two things you asked to preserve:** the validity checks are untouched, and `secondaryNameSameAsPrimary` stays in the entry condition. It was originally added for the stale write this change now prevents, but it independently guards a second case — `usernameSubmitButtonState` takes no collision input and enables a Secondary on `!usernameExists && !usernameContestable`, which a contested primary satisfies on its own, since it is absent from the unique index until its vote resolves. I corrected the comment there, which had claimed the staleness rationale was the only one.

**Verification.** Three ViewModel regression tests drive genuine out-of-order completion — a latch holds `alice` open until `bob` has answered, then releases it — covering a replaced name, a cleared field, and a late failure. Mutation-verified: forcing `usernameCheckResultIsCurrent` to `true` turns all three red. Six pure tests pin the predicate itself, including that matching is exact rather than prefix-based, which the instant-name screen needs since it pre-fills the primary and the user appends a suffix.

Full suite 2020 tests, 0 failures; ktlint clean.

wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
```diff
@@ -255,7 +258,57 @@ class RestoreIdentityWorker @AssistedInject constructor(
...
+                        )
+                    } catch (e: Exception) {
+                        log.warn(
+                            "contested re-registration guard: could not read the vote state for '{}' — " +
+                                "treating it as ALREADY requested and skipping re-registration (fail closed; " +
+                                "a later trigger re-drives this)",
+                            requestedLabel,
+                            e
+                        )
+                        true
```
thepastaclaw: **🔴 Blocking: Preserve a retryable restoration error when the contender check fails**

Converting a failed contender lookup to true prevents a duplicate write, but does not reliably defer registration. If the identity's requested contested name has not landed, this skips registration and the subsequent contested-name walks can legitimately find no matching contender. The worker then reaches lines 781–785, clears restoring, and records "missing domain document". HistoryHeaderAdapter presents that as an unavailable username, and WalletTransactionsFragment.openIdentityCreation routes the user to choose another name instead of retrying restoration. Later contender lookups also collapse failures into empty results, so they cannot establish that continuing is safe. Abort this attempt with a retryable restoration error while retaining restoring=true, or carry an explicit unknown outcome that cannot enter the missing-domain branch. Add a worker-level test covering a failed guard lookup with no matching contender in the subsequent lists.

wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt
```diff
@@ -186,7 +186,15 @@ class BlockchainStateDataProvider @Inject constructor(
...
             blockchainStateDao.saveState(blockchainState)
-            syncStageFlow.value = update.syncStage
+            // A caught-up snapshot must not REGRESS an established stage — that is
+            // the "syncing 100%" blip (see mayPreserveEstablishedSyncStage). But
+            // only an ESTABLISHED one: with the flow still null there is nothing to
+            // preserve, and holding the write would leave getSyncStage() reporting
+            // OFFLINE for a wallet that is actually caught up — the process-start
+            // case. Seed it then, and hold only afterwards.
+            if (!update.preserveEstablishedSyncStage || syncStageFlow.value == null) {
+                syncStageFlow.value = update.syncStage
```
thepastaclaw: **🔴 Blocking: Do not preserve OFFLINE when caught-up progress resumes**

This condition preserves every non-null stage, including OFFLINE. IDLE, CONNECTING, and ERROR snapshots explicitly write OFFLINE; if the next snapshot is FILTERS with the scan caught up, the mapper sets preserveEstablishedSyncStage=true and this branch suppresses the transition to BLOCKS. Repeated caught-up active snapshots retain OFFLINE until a SYNCED or below-100 snapshot arrives, despite the scan being active and caught up. The null-stage startup test does not cover this reconnect/recovery sequence. Restrict preservation to the completed state needed for the blip fix, or otherwise allow forward transitions out of OFFLINE. Add provider-level CONNECTING/ERROR → caught-up FILTERS tests while retaining the COMPLETE → caught-up FILTERS regression test.

wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt
```diff
@@ -847,6 +849,112 @@ class RequestUserNameViewModelTest {
...
+        viewModel.checkUsername("bob")
+        viewModel.uiState.first { it.usernameCheckSuccess }
+        assertTrue("bob is registered", viewModel.uiState.value.usernameExists)
+
+        // Now alice's answer lands. It says "free" — for a name that is no longer there.
+        aliceInFlight.countDown()
+        viewModel.uiState.first { !it.checkingUsername }
+
+        val state = viewModel.uiState.value
+        assertTrue("bob's verdict must survive the late arrival", state.usernameExists)
```
thepastaclaw: **🟡 Suggestion: Wait for the stale lookup coroutine to finish before asserting**

Bob's successful result has already set checkingUsername=false before Alice's latch is released. Consequently, first { !it.checkingUsername } can return immediately from Bob's existing state, before Alice's coroutine reaches the stale-result guard. The assertion can pass without exercising that guard, and runVmTest then cancels the ViewModel scope in its finally block. The reported mutation run does not eliminate this scheduling-dependent false pass. Join the specific lookup job or use controlled dispatchers with a completion barrier that proves the terminal result-handling code ran before asserting. Replace the fixed sleeps in the two adjacent stale-result tests with equivalent completion synchronization.

wallet/src/de/schildbach/wallet/ui/username/request/RequestUserNameViewModel.kt
```diff
@@ -1618,7 +1665,7 @@ class RequestUserNameViewModel @Inject constructor(
                 usernameNonContestedChars = validateNonContestedUsernameCharacters(username)
             )
         }
-        return validCharacters && validLength
+        return validCharacters && validLength && !sameAsPrimary
```
thepastaclaw: **🟡 Suggestion: Stale-drop can leave the availability spinner stuck after an invalid keystroke**

When a valid lookup starts, checkUsername sets checkingUsername=true. If the user then changes the field to a non-empty invalid name, checkUsernameValid advances pendingUsernameCheck but leaves checkingUsername unchanged, and processUsername schedules no replacement lookup. The old lookup correctly returns through a stale-result guard without clearing checkingUsername, leaving the primary screen's voting-period spinner visible indefinitely until another lookup completes or the state is reset. Clear the checking state when validation invalidates an outstanding lookup without scheduling a replacement; do not clear it from a stale completion that could belong to an older lookup than the current one. Add a valid-in-flight → invalid-input → old-result-completes test.

<sub>source: `muse-spark-1.3-contributor` (phase1-reviewer: general)</sub>

thepastaclaw (changes requested):
## Re-review — Final validation — Phase 1 + Phase 2

Verified the supplied findings against head 2d174cff3d8d743b3e02c0b060aac2d1e1133408 and the PR base: two blocking recovery/sync regressions and two suggestions remain. The mutable-dependency concern remains technically present but is intentionally deferred to production-release readiness under the author's explicit non-production scope. Source inspection and git diff --check completed; no builds or tests were executed during this verification.

🔴 2 blocking | 🟡 2 suggestion(s)

<details>

Source: reviewer 1: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `security-auditor`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `normal` by `gpt-6-astra` (effort low) — The PR spans several behavioral fixes and asynchronous UI states, but its funds-related change in RestoreIdentityWorker is a small fail-closed registration guard rather than a large or intricate change to funds movement, and the remaining changes concern UI, sync reporting, identity reads, tests, and a dependency bump.
- Phase 1 reviewers: `muse-spark-1.3-contributor` — general (completed, effort xhigh); agent `phase1-reviewer`, `muse-spark-1.3-contributor` — security-auditor (completed, effort xhigh); agent `phase1-reviewer`
- Phase 1 model: `muse-spark-1.3-contributor` — not quota-gated; passed over `gemini-3.8-flash-high` (lane failed), `glm-5.3-flash` (not used above high effort; tier asks max)

</details>

<details>

```

In `wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt`:
- [BLOCKING] wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt:291-299: Preserve a retryable restoration error when the contender check fails
  Converting a failed contender lookup to true prevents a duplicate write, but does not reliably defer registration. If the identity's requested contested name has not landed, this skips registration and the subsequent contested-name walks can legitimately find no matching contender. The worker then reaches lines 781–785, clears restoring, and records "missing domain document". HistoryHeaderAdapter presents that as an unavailable username, and WalletTransactionsFragment.openIdentityCreation routes the user to choose another name instead of retrying restoration. Later contender lookups also collapse failures into empty results, so they cannot establish that continuing is safe. Abort this attempt with a retryable restoration error while retaining restoring=true, or carry an explicit unknown outcome that cannot enter the missing-domain branch. Add a worker-level test covering a failed guard lookup with no matching contender in the subsequent lists.

In `wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt`:
- [BLOCKING] wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt:195-196: Do not preserve OFFLINE when caught-up progress resumes
  This condition preserves every non-null stage, including OFFLINE. IDLE, CONNECTING, and ERROR snapshots explicitly write OFFLINE; if the next snapshot is FILTERS with the scan caught up, the mapper sets preserveEstablishedSyncStage=true and this branch suppresses the transition to BLOCKS. Repeated caught-up active snapshots retain OFFLINE until a SYNCED or below-100 snapshot arrives, despite the scan being active and caught up. The null-stage startup test does not cover this reconnect/recovery sequence. Restrict preservation to the completed state needed for the blip fix, or otherwise allow forward transitions out of OFFLINE. Add provider-level CONNECTING/ERROR → caught-up FILTERS tests while retaining the COMPLETE → caught-up FILTERS regression test.

In `wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt`:
- [SUGGESTION] wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt:887-892: Wait for the stale lookup coroutine to finish before asserting
  Bob's successful result has already set checkingUsername=false before Alice's latch is released. Consequently, first { !it.checkingUsername } can return immediately from Bob's existing state, before Alice's coroutine reaches the stale-result guard. The assertion can pass without exercising that guard, and runVmTest then cancels the ViewModel scope in its finally block. The reported mutation run does not eliminate this scheduling-dependent false pass. Join the specific lookup job or use controlled dispatchers with a completion barrier that proves the terminal result-handling code ran before asserting. Replace the fixed sleeps in the two adjacent stale-result tests with equivalent completion synchronization.

In `wallet/src/de/schildbach/wallet/ui/username/request/RequestUserNameViewModel.kt`:
- [SUGGESTION] wallet/src/de/schildbach/wallet/ui/username/request/RequestUserNameViewModel.kt:1645-1668: Stale-drop can leave the availability spinner stuck after an invalid keystroke
  When a valid lookup starts, checkUsername sets checkingUsername=true. If the user then changes the field to a non-empty invalid name, checkUsernameValid advances pendingUsernameCheck but leaves checkingUsername unchanged, and processUsername schedules no replacement lookup. The old lookup correctly returns through a stale-result guard without clearing checkingUsername, leaving the primary screen's voting-period spinner visible indefinitely until another lookup completes or the state is reset. Clear the checking state when validation invalidates an outstanding lookup without scheduling a replacement; do not clear it from a stale completion that could belong to an older lookup than the current one. Add a valid-in-flight → invalid-input → old-result-completes test.
```
</details>

<details>
<summary>Out-of-scope follow-up suggestions (1)</summary>

These are valid observations, but they are outside this PR's scope and should be handled in separate issues or author/maintainer-requested PRs rather than blocking this review.

- **Track immutable SDK dependencies as a production-release gate** — The current prod dependency graph includes the newly selected DPP 4.0.1-SNAPSHOT and the separate Android SDK snapshot already present on the base, without checked-in dependency verification. This is a concrete release-integrity concern, but the author explicitly excludes production distribution from this pre-release PR's scope.
  - Follow-up: Track production readiness separately: use immutable releases or pin and verify the exact artifact contents for both SDK dependency families before production distribution.

</details>

wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
HashEngineering: Valid, and fixed in `d3f8c7886`.

I traced the path you describe and it is real. After the guard skips, the contested walk runs; if `currentUsername` is still null it falls through to `RestoreIdentityWorker.kt:781-785`, which sets `restoring = false`, stamps `missing domain document`, and routes the user to a different name. That branch's own comment (`:774-780`) asserts it is *"reached only when ... there was nothing to register here (no requested USERNAME persisted)"* — my guard broke exactly that invariant, since it reaches there with a label present. Your point about the later walks is right too: they collapse failures into empty results, so they cannot establish that continuing is safe.

The original reasoning was "a duplicate costs 0.2 DASH irrecoverably, deferring costs one worker trigger". That is correct about the duplicate and wrong about the deferral — skipping only defers registration *if the name actually landed*.

**Fix.** The guard is now three-way rather than a boolean. `getVoteContendersOrThrow` returning normally gives contend/not-contend; a failure gives null, which maps to `DEFER_UNREADABLE` and throws the same retryable `IllegalStateException` the registration paths at `:338-347` already use — so `restoring` stays true, the state stays at `USERNAME_REGISTERING`, the tile shows the retry card, and the next trigger re-drives the guard with a fresh read. No duplicate is possible, and nothing is decided on evidence the worker does not have. `CancellationException` is now rethrown rather than swallowed into the failure branch.

**On the worker-level test:** I did not add one, and I want to be straight about why rather than quietly skip it. `RestoreIdentityWorker` has no test harness in this repo — the only existing tests for it (`ContestedReregistrationGuardTest`, `ContestedNameNarrowingTest`) are pure-logic, and standing up the full dependency graph is well beyond this fix. Instead the decision is split into a pure `contestedReregistrationAction(requestedLabel, ownIdentityIsContender: Boolean?)` and pinned exhaustively (six cases), including one that asserts a skip can only ever be produced by a positive contender answer. The fall-through you describe is what that null case now prevents; if you want the end-to-end worker test as follow-up work, it needs a harness first.

wallet/src/de/schildbach/wallet/service/BlockchainStateDataProvider.kt
HashEngineering: Valid, and fixed in `d3f8c7886`. This was my regression.

Confirmed the sequence exactly as you describe: `sdkSyncStage` maps IDLE, CONNECTING and ERROR to `SyncStage.OFFLINE` (`SdkBlockchainStateMapper.kt:186`), and all three *write* it, because `shadowSyncPercent` returns 0 for them (`L1ShadowSyncService.kt:293-295`) so `mayPreserveEstablishedSyncStage` is false. The caught-up FILTERS snapshot that follows recovery gives `FILTERS != SYNCED && 100 >= 100` → true, and with OFFLINE non-null the write was suppressed. The wallet then read offline with an active scan sitting at the tip, until a SYNCED or below-100 snapshot arrived.

**Fix**, and it is the narrowing you suggested: preserve only COMPLETE.

```kotlin
if (!update.preserveEstablishedSyncStage || syncStageFlow.value != SyncStage.COMPLETE) {
    syncStageFlow.value = update.syncStage
}
```

COMPLETE is the only stage the "syncing 100%" blip fix ever needed to hold — a live shadow SPV never latches SYNCED, so a caught-up FILTERS snapshot would drag a finished row back to BLOCKS. Everything else now transitions forward normally. The null-flow seeding case you noted is not separately handled any more and does not need to be: `null != COMPLETE`, so a first update still seeds.

**Tests**, all three you asked for, at provider level: CONNECTING → caught-up FILTERS and ERROR → caught-up FILTERS both assert the forward transition out of OFFLINE, plus HEADERS → caught-up FILTERS for the intermediate case. The COMPLETE → caught-up FILTERS regression test and the null-stage startup test are untouched and still green.

wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt
HashEngineering: Correct on all counts, fixed in `d3f8c7886`.

You are right that `first { !it.checkingUsername }` could return from Bob's already-written state before Alice's coroutine reached the guard, and right that the mutation run does not rule that out — it shows the assertion *can* fail, not that it fails for the intended reason. The underlying problem is that a stale result is dropped without touching the UI state, so there is deliberately nothing to observe, and any state-based wait can be satisfied by a different lookup.

**Fix.** `checkUsername` now returns its `Job`, so a test can await the terminal handling of one *specific* lookup:

```kotlin
val aliceLookup = viewModel.checkUsername("alice")
viewModel.checkUsername("bob")?.join()   // bob's verdict is recorded
aliceInFlight.countDown()
aliceLookup?.join()                      // alice reached the guard and returned
```

Both fixed sleeps in the adjacent tests are gone, replaced by the same join.

**One more thing that fell out of this, worth recording.** Writing the new spinner test I hit a genuine hang rather than a failure: an assertion failed while a mock was blocked in `CountDownLatch.await()` on a `Dispatchers.IO` thread, the latch was never released, and the Gradle test JVM could not exit — a 10-minute hang instead of a red test. Every latch in these tests now counts down in a `finally`. Worth knowing for anyone adding to this file.

Full suite 2031 tests green.

wallet/src/de/schildbach/wallet/ui/username/request/RequestUserNameViewModel.kt
HashEngineering: Valid — this one was introduced by my stale-result guard in `2d174cff3`. Fixed in `d3f8c7886`.

Verified the mechanism: `checkUsernameValid`'s `_uiState.update` never touched `checkingUsername`, so before the guard existed the stale completion cleared it as a side effect of writing its result. Now it returns early and writes nothing, and for an invalid keystroke `processUsername` schedules no replacement — so the flag stayed true with nothing left to clear it.

**Fix**, following your constraint that it must not be cleared from a stale completion. It is cleared in the validity pass instead, where the caller's own scheduling decision is already being computed:

```kotlin
val lookupWillFollow = validCharacters && validLength && !sameAsPrimary   // == the return value
...
checkingUsername = if (lookupWillFollow) it.checkingUsername else false,
```

Keyed to the exact predicate the fragment uses to decide whether to schedule, so the two cannot drift. When a replacement *is* coming the flag is left alone, so the spinner keeps running across the 600 ms debounce instead of flickering off and on.

**Tests**, both the one you asked for and its complement: valid-in-flight → invalid input → old result completes (spinner cleared, and not resurrected by the completion), and valid-in-flight → still-valid input (spinner kept). Mutation-verified — reverting the one-line clear reds `checkUsername_invalidKeystrokeWhileALookupIsOut_doesNotStrandTheSpinner`.

Note for anyone reading the predicate: `lookupWillFollow` intentionally uses `validCharacters` rather than `usernameCharactersValid`, matching the pre-existing return value, which does not fold in the leading/trailing-hyphen rule. I kept that as-is rather than changing scheduling behaviour inside a spinner fix.

wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
```diff
@@ -255,7 +259,82 @@ class RestoreIdentityWorker @AssistedInject constructor(
...
+                // more worker trigger; a duplicate costs 0.2 DASH irrecoverably.
+                var voteStateFailure: Exception? = null
+                val contenderVerdict = if (contestedReregistrationCheckApplies(requestedLabel)) {
+                    try {
+                        isOwnIdentityAContender(
+                            platformRepo.getVoteContendersOrThrow(requestedLabel!!).map.keys,
+                            blockchainIdentity.uniqueIdentifier
+                        )
+                    } catch (e: CancellationException) {
+                        throw e
```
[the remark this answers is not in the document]
HashEngineering: Correct, fixed in `c5c94d5fe`. The rethrow I added was doing nothing.

Traced both handlers: `restoreIdentity`'s outer `catch (e: Exception)` catches `CancellationException`, stamps it onto the identity row via `updateIdentityCreationState(it, it.creationState, e)` and calls `triggerPreBlockDownloadComplete()` before rethrowing; then `doWorkInForeground`'s `catch (ex: Exception)` logs an analytics error and returns `Result.failure`. So WorkManager stopping the worker was recorded as a genuine restoration failure, with an error on the row and the pre-block stage reported complete when nothing had completed.

Explicit `CancellationException` rethrows now sit ahead of both, as you asked.

I also added one you did not flag, at the `isCutoverCommitted()` read (`:258`): its `catch (Exception)` answered a cancelled call with `false`, i.e. "cutover not committed", which would then take the dashj path. Same class of bug, one line, so it seemed wrong to leave it next to the two being fixed.

wallet/src/de/schildbach/wallet/ui/username/request/RequestUserNameViewModel.kt
```diff
@@ -1703,3 +1767,44 @@ class RequestUserNameViewModel @Inject constructor(
...
+ * The older one would otherwise write `usernameCheckSuccess = true` together with ITS
+ * `usernameExists` / `usernameContested` / `usernameBlocked`, and submit would be enabled for
+ * the name now in the box on the strength of a different name's answer.
+ *
+ * A null [currentLabel] means the field was cleared (reset), so nothing is current.
+ *
+ * Pure — host-testable.
+ */
+internal fun usernameCheckResultIsCurrent(checkedLabel: String, currentLabel: String?): Boolean =
+    currentLabel != null && checkedLabel == currentLabel
```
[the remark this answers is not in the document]
HashEngineering: You are right, and this one corrects a judgement I had explicitly made and pinned in a test. Fixed in `c5c94d5fe`.

I had reasoned that in `alice -> bob -> alice` the late first-alice result describes exactly what the field holds, so applying it is correct rather than stale — and wrote `aNameRetypedAfterDetouringIsCurrentAgain` asserting that. That reasoning covers the availability fields and nothing else: it ignores that the first lookup also writes `checkingUsername = false` while the second is still running, and that its read is simply older. As you note, the fragment removes only the delayed debounce callback, never the launched job.

**Fix.** `pendingUsernameCheck: String?` is now `usernameCheckToken: Long`, pre-incremented by every claim on the slot — each keystroke in the validity pass and each lookup — and captured per lookup. Only the newest claim can write. It also removed bookkeeping rather than adding it: one counter in place of a label that had to be kept correct in two places. The test above is now the opposite assertion, over tokens.

**On the verification, because the first attempt was not sound.** My initial probe simulated label semantics in the token bookkeeping and left every test green — the new test could not tell a stale write from a fresh one, since both alice lookups returned the same thing, so all its assertions held either way. The test now makes them DISAGREE: the stale lookup reports alice taken, the current one reports it free, and the assertion is that the field still reads available. A faithful probe — swapping `usernameCheckIsStale` back to comparing labels — reds both that test and `checkUsername_lateResultAfterTheFieldIsCleared_isDropped`.

Full suite 2031 tests green, ktlint clean.

wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
```diff
@@ -251,11 +260,88 @@ class RestoreIdentityWorker @AssistedInject constructor(
...
+                }
+                val alreadyContending =
+                    reregistrationAction == ContestedReregistrationAction.SKIP_ALREADY_CONTENDING
+
+                if (cutoverCommitted && !requestedLabel.isNullOrEmpty() && alreadyContending) {
+                    log.info(
+                        "'{}' is a contested name this identity already contends for — skipping DPNS " +
+                            "re-registration; the contested-name walk below recovers the voting state",
+                        requestedLabel
+                    )
```
thepastaclaw, line 349: **🔴 Blocking: Preserve the positive contender result through the recovery walk**

When the guard positively identifies this identity's contender, this branch skips registration but leaves currentUsername null and discards the fetched contender data. The recovery walks then fetch the contenders again at lines 485 and 650 through PlatformRepo.getVoteContenders(), which converts exceptions into empty maps. If the name-list reads succeed but these subsequent contender reads fail, neither walk populates currentUsername. Lines 814–818 consequently clear restoring and record "missing domain document" despite already having confirmed the paid contender exists. HistoryHeaderAdapter presents the username as unavailable, and WalletTransactionsFragment routes the user to choose another name rather than retry restoration. Reuse the successful contender response, or retain the positive verdict and prevent an unsuccessful recovery walk from entering the missing-domain branch; leave a retryable restoration error instead. Add a regression test for a successful guard followed by failed recovery reads.

wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt
```diff
@@ -847,6 +849,252 @@ class RequestUserNameViewModelTest {
...
+        val staleAlice = try {
+            viewModel.checkUsernameValid("alice", UsernameType.Primary)
+            val job = viewModel.checkUsername("alice")
+
+            viewModel.checkUsernameValid("bob", UsernameType.Primary)
+            viewModel.checkUsername("bob")?.join()
+
+            // Back to alice — a NEW lookup, which answers immediately and says "free".
+            viewModel.checkUsernameValid("alice", UsernameType.Primary)
+            viewModel.checkUsername("alice")?.join()
```
thepastaclaw, line 1012: **🟡 Suggestion: Await the first Alice lookup entering the mock before launching its replacement**

The mock assigns the blocked, stale response using aliceCalls++, but getUsername runs on real Dispatchers.IO threads. Starting the first lookup does not establish that it has entered the mock, and joining Bob establishes no ordering between the two Alice IO tasks. If the replacement Alice enters first, it takes firstAlice.await(), while the test joins that replacement before reaching the finally block that releases the latch. The test then times out instead of exercising stale-result rejection. Signal entry into the first Alice mock and await that signal before launching Bob and the replacement Alice, so the blocked response is deterministically associated with the older request token.

thepastaclaw (changes requested):
## Re-review — Final validation — Phase 1 + Phase 2

Verified the supplied findings and all five prior findings against head c5c94d5fe811373e35525ea80eb34db284a73a25. Four prior issues are fixed and the snapshot dependency remains an explicitly deferred production-release gate; one blocking restoration error path and one asynchronous test race remain. Source inspection and git diff --check completed successfully; tests were not executed in this verification pass.

🔴 1 blocking | 🟡 1 suggestion(s)

<details>

Source: reviewer 1: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `security-auditor`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `critical` by `gpt-6-astra` (effort low) — This large, intricate change directly alters funds-spending authorization in RestoreIdentityWorker.kt by deciding whether to execute paid DPNS re-registration, with contender lookup failures, cancellation, and recovery state determining whether another prefund and registration fees are spent.
- Phase 1 reviewers: `muse-spark-1.3-contributor` — general (completed, effort xhigh); agent `phase1-reviewer`, `muse-spark-1.3-contributor` — security-auditor (completed, effort xhigh); agent `phase1-reviewer`
- Phase 1 model: `muse-spark-1.3-contributor` — not quota-gated; passed over `gemini-3.8-flash-high` (lane failed), `glm-5.3-flash` (not used above high effort; tier asks max)
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort xhigh); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort xhigh); agent `phase2-reviewer`

</details>

<details>

```

In `wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt`:
- [BLOCKING] wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt:338-343: Preserve the positive contender result through the recovery walk
  When the guard positively identifies this identity's contender, this branch skips registration but leaves currentUsername null and discards the fetched contender data. The recovery walks then fetch the contenders again at lines 485 and 650 through PlatformRepo.getVoteContenders(), which converts exceptions into empty maps. If the name-list reads succeed but these subsequent contender reads fail, neither walk populates currentUsername. Lines 814–818 consequently clear restoring and record "missing domain document" despite already having confirmed the paid contender exists. HistoryHeaderAdapter presents the username as unavailable, and WalletTransactionsFragment routes the user to choose another name rather than retry restoration. Reuse the successful contender response, or retain the positive verdict and prevent an unsuccessful recovery walk from entering the missing-domain branch; leave a retryable restoration error instead. Add a regression test for a successful guard followed by failed recovery reads.

In `wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt`:
- [SUGGESTION] wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt:992-1001: Await the first Alice lookup entering the mock before launching its replacement
  The mock assigns the blocked, stale response using aliceCalls++, but getUsername runs on real Dispatchers.IO threads. Starting the first lookup does not establish that it has entered the mock, and joining Bob establishes no ordering between the two Alice IO tasks. If the replacement Alice enters first, it takes firstAlice.await(), while the test joins that replacement before reaching the finally block that releases the latch. The test then times out instead of exercising stale-result rejection. Signal entry into the first Alice mock and await that signal before launching Bob and the replacement Alice, so the blocked response is deterministically associated with the older request token.
```
</details>

wallet/src/de/schildbach/wallet/service/platform/work/RestoreIdentityWorker.kt
HashEngineering: Valid, fixed in `3b6252991`. This is the sibling of the unreadable-guard case from the last round, and I had only closed half the problem.

Verified the path: the skip branch leaves `currentUsername` null, the walks re-read through `platformRepo.getVoteContenders` at `:485` and `:650`, and that collapses an exception into an empty map (`PlatformRepo.kt:207`) — so a contender read that errors is indistinguishable there from a name with no contenders. Neither walk populates `currentUsername`, and the fall-through parks the identity. Your point about what parking actually does is the one that matters: it is not a neutral "ask again", it clears `restoring`, shows the name as unavailable and routes the user to pick a DIFFERENT one — throwing away a contested name whose 0.2 DASH prefund is already spent, on evidence that cannot support the conclusion.

**Fix.** I took your second option rather than reusing the response. The positive verdict is carried out of the guard block to the fall-through, and once a contender has been confirmed, "not found" can only mean "could not read it back" — so the worker throws the same retryable error the registration paths use instead of parking. Reusing the fetched contender data would also work, but it would mean reconstructing the username state the walks build, in a second place, and I would rather not have two writers of that.

**On the test:** the decision is now a pure `missingNameOutcome(confirmedOwnContender)`, pinned with three cases — defer when confirmed, keep the historic ask-for-a-new-name path when nothing was confirmed, and one asserting the destructive branch is reachable ONLY with nothing confirmed. I still cannot give you the end-to-end worker test (`RestoreIdentityWorker` has no harness in this repo), so this is the same trade as last round, stated plainly rather than implied.

wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt
HashEngineering: Correct — thanks, this would have been an intermittent CI hang rather than a failure.

`aliceCalls++` bound the blocked response to whichever call reached the mock first, and since `getUsername` runs on real `Dispatchers.IO` threads, starting the first lookup established nothing about entry order; joining bob ordered bob against each alice but not the two alices against each other. If the replacement had arrived first it would have taken `firstAlice.await()`, and the test joins that replacement *inside* the try, before the `finally` releases the latch — so it would hang until the Gradle timeout instead of failing. Exactly the failure mode I had just fixed elsewhere in this file by moving the latch releases into `finally`, and I missed that the same reasoning applied to entry order.

Fixed in `3b6252991`: the first alice signals entry through its own latch, and the test awaits that signal (bounded, 10 s, asserted) before launching bob and the replacement. The stale response is now deterministically bound to the older token.

wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt
```diff
@@ -847,6 +850,263 @@ class RequestUserNameViewModelTest {
...
+
+        // The user types on: the field now holds 'bob', whose lookup answers first.
+        viewModel.checkUsernameValid("bob", UsernameType.Primary)
+        viewModel.checkUsername("bob")?.join()
+        assertTrue("bob is registered", viewModel.uiState.value.usernameExists)
+
+        // Now alice's answer lands. It says "free" — for a name that is no longer there.
+        // Join ALICE's own job: bob has already cleared checkingUsername, so waiting on
+        // the state could return from bob's value before alice reached the stale guard.
+        aliceInFlight.countDown()
```
thepastaclaw: **🟡 Suggestion: Release the stale-lookup latch when an assertion fails**

If Alice has entered the mock's blocking await() and Bob's assertion fails, execution skips aliceInFlight.countDown(). runVmTest cancels viewModelScope, but cancellation does not interrupt CountDownLatch.await() inside withContext(Dispatchers.IO); tearDown then waits indefinitely in scopeJob.join(). This turns a regression failure into a hung suite. Put the latch release in a finally block, as the same-label and spinner tests already do, and protect the adjacent latch-based tests against early exits as well.

```suggestion
        try {
            viewModel.checkUsernameValid("bob", UsernameType.Primary)
            viewModel.checkUsername("bob")?.join()
            assertTrue("bob is registered", viewModel.uiState.value.usernameExists)
        } finally {
            aliceInFlight.countDown()
        }
```

thepastaclaw (commented):
## Re-review — Final validation — Phase 1 + Phase 2

Independently inspected the complete diff at 3b625299123a7420f8f3fdbac11bd10d01a66ef2 and revalidated all seven prior findings: six are fixed, and the mutable-dependency concern remains an explicitly deferred production release gate. One non-blocking regression-test cleanup issue remains; no in-scope blocker was confirmed. The complete-range diff check passed; unit tests were not independently rerun in this verification.

🟡 1 suggestion(s)

<details>

Source: reviewer 1: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `security-auditor`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); reviewer 5: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 6: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `critical` by `gpt-6-astra` (effort low) — The diff is intricate and directly changes funds-spending authorization in RestoreIdentityWorker.kt by deciding whether to execute paid DPNS re-registration or defer recovery based on contender ownership, failed reads, and persisted restoration state.
- Phase 1 reviewers: `muse-spark-1.3-contributor` — general (completed, effort xhigh); agent `phase1-reviewer`, `muse-spark-1.3-contributor` — security-auditor (completed, effort xhigh); agent `phase1-reviewer`
- Phase 1 model: `muse-spark-1.3-contributor` — not quota-gated; passed over `gemini-3.8-flash-high` (lane failed), `glm-5.3-flash` (not used above high effort; tier asks max)
- Fresh final gate: an independent Phase-2 review ran after iterative findings were reconciled
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort xhigh); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort xhigh); agent `phase2-reviewer`, `gpt-6-astra` — general (completed, effort xhigh); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort xhigh); agent `phase2-reviewer`

</details>

<details>

```

In `wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt`:
- [SUGGESTION] wallet/test/de/schildbach/wallet/ui/username/request/RequestUserNameViewModelTest.kt:883-890: Release the stale-lookup latch when an assertion fails
  If Alice has entered the mock's blocking await() and Bob's assertion fails, execution skips aliceInFlight.countDown(). runVmTest cancels viewModelScope, but cancellation does not interrupt CountDownLatch.await() inside withContext(Dispatchers.IO); tearDown then waits indefinitely in scopeJob.join(). This turns a regression failure into a hung suite. Put the latch release in a finally block, as the same-label and spinner tests already do, and protect the adjacent latch-based tests against early exits as well.
```
</details>

<details>
<summary>Out-of-scope follow-up suggestions (1)</summary>

These are valid observations, but they are outside this PR's scope and should be handled in separate issues or author/maintainer-requested PRs rather than blocking this review.

- **Pin and verify both SDK dependency families before production release** — The current prod flavor resolves the newly selected DPP 4.0.1-SNAPSHOT and the base's existing unscoped dash-sdk-android SNAPSHOT. No checked-in dependency verification metadata constrains their contents. This is a concrete production artifact-integrity concern, but the author's explicit statement that this line is not shipping to production makes it a deferred release-wide gate rather than a blocker for this bugfix PR.
  - Follow-up: Track a production release gate to select reviewed immutable artifacts or verify the exact approved artifact contents for both SDK dependency families before shipment.

</details>

thepastaclaw (changes requested):
## ⚠️ DEGRADED — Re-review — Preliminary review — Phase 1 blocker gate

> **⚠️ DEGRADED review.** The primary review models were unavailable (`gpt-6-astra` unavailable: All credentials for model gpt-6-astra are cooling down (last error: usage_limit_reached: The usage limit has been reache), so this review ran on stand-in models: `gpt-5.6-luna` → `muse-spark-1.3-contributor`, `gpt-5.6-sol` → `muse-spark-1.3-contributor`, `gpt-5.6-terra` → `muse-spark-1.3-contributor`, `gpt-6-astra` → `muse-spark-1.3-contributor`. Both review phases and the independent verifiers still ran, but on weaker models, with Phase 1 capped at `high` effort. Treat the verdict as provisional; a full-strength re-review will run on the next push once the primary models are back.

Revalidated all 8 prior findings at 6cc5108c: 7 guard, sync-stage, spinner, and test-ordering fixes are verified present and correct. Production still resolves the mutable DPP snapshot with no immutable alternative published yet, so one blocking release-integrity finding carries forward.

Validated blockers were found by the Phase-1 review and confirmed by a fresh verifier. Phase 2 is deferred until a fresh same-head revalidation clears the blocker gate.

🔴 1 blocking

_1 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

Source: reviewer 1: `muse-spark-1.3-contributor` (agent: `phase1-reviewer`, role: `general`); final verifier: `muse-spark-1.3-contributor` (standing in for `gpt-6-astra`) (agent: `astra-gate-verifier`, role: `verifier`)

- **Degraded mode**: `gpt-6-astra` unavailable: All credentials for model gpt-6-astra are cooling down (last error: usage_limit_reached: The usage limit has been reache (detected by probe, since 2026-09-20T20:45:34Z); stand-ins `gpt-5.6-luna` → `muse-spark-1.3-contributor`, `gpt-5.6-sol` → `muse-spark-1.3-contributor`, `gpt-5.6-terra` → `muse-spark-1.3-contributor`, `gpt-6-astra` → `muse-spark-1.3-contributor`; Phase 1 effort capped at `high`
- Triage: `normal` by `muse-spark-1.3-contributor` (standing in for `gpt-6-astra`) (effort low) — Large 22-file username/identity/UI/sync fix with no direct change to consensus rules, coin selection/funds movement, cryptography/signatures/keys, peer deserialization, or storage migrations.
- Phase 1 reviewers: `muse-spark-1.3-contributor` — general (completed, effort high); agent `phase1-reviewer`
- Phase 1 model: `muse-spark-1.3-contributor` — not quota-gated; passed over `gemini-3.8-flash-high` (lane failed), `glm-5.3-flash` (zai below 15% reserve: 5h 100% left, weekly 13% left)
- Fresh verifier: `muse-spark-1.3-contributor` (standing in for `gpt-6-astra`) — verifier; agent `astra-gate-verifier`
- Phase 2 reviewers: **not run (deferred by blocker gate)**

</details>

<details>

```

In `build.gradle`:
- [BLOCKING] build.gradle:23: Production builds resolve a mutable DPP SNAPSHOT dependency
  (existing thread: https://github.com/dashpay/dash-wallet/pull/1564#discussion_r4010588230)
  Verified at head: build.gradle:23 sets dppVersion = "4.0.1-SNAPSHOT" and wallet/build.gradle:35 maps prod to dppVersion, so prodImplementation resolves a mutable snapshot. No gradle/verification-metadata.xml or lockfile constrains it (confirmed no such files in the repo). A prod build from this head would embed an unconstrained mutable artifact whose contents can change between builds. Pin prod to an immutable release once 4.0.1 publishes, or pin and verify the exact snapshot contents before any production shipping.
```
</details>

